PR #6, opened 2017-10-09
Bad link.

rafaelschipiura:
Link doesn't work with the .html at the end.


hughsie:
Fixed in a slightly different way, thanks.

PR #19, opened 2018-01-09
Wip/hughsie/reports

hughsie:


hughsie:
This is what the new UI looks like, the first one being admin-only.

screenshot from 2018-01-09 15-16-38
screenshot from 2018-01-09 15-16-54
screenshot from 2018-01-09 15-16-47

superm1:
This LGTM.  The only thing that I'd say is missing is a way on that Reports page to view just failures.
Success is mostly interesting for analytics types graphs, but if there are too many of those it will just flood the page and make it hard to get at the failures.

hughsie:
Was there any data you expected to see? I think from my point of view the distro version, architecture and device quirks would also be useful.

superm1:
Distro is probably important; good point.
Could you put all device metadata and quirks both?  We might consider to put in the metadata for plugins later the reason they failed last time if we have it.

I'm thinking specifically about the UEFI plugin and how we can capture it's failures.  Since it requires a reboot, it can be as simple as if fwupd was run again it's marked as a failure and look at ESRT and encode last_attempt_status and last_attempt_version into metadata.

PR #21, opened 2018-01-10
Add a privacy statement for the upcoming GDPR legislation

hughsie:


superm1:
IANAL, but looks good to me.

hughsie:
I don't think you need to be a layer to write a privacy statement for the GDPR, but I appreciate the review.

PR #23, opened 2018-01-17
Use GObject Introspection to read and write AppStream

hughsie:
This means we're using the exact same code as fwupd and gnome-software.

massive plus point: huge amounts of code deleted and much more battle-tested code (also much faster)
massive minus point: you need an up to date version appstream-glib installed on the server which rules out things like openshift in the future

Gut reaction welcome.


superm1:
Gut reaction; It's the right thing to do.  Hearing it's much faster is a great thing for scalability too.  Does openshift not let you have your own custom containers?  I feel like requiring an installed package isn't a giant bar these days.  You can run lxc or docker images on anything and just set up the environment that way.

hughsie:
The new openshift does let you basically host a docker container, so maybe my worries are unfounded. I'll do some more checks and then deploy the new code to the new test instance. Details to follow.

PR #24, opened 2018-01-18
Use Puppet to set up the server

hughsie:
There still is a database, so we can't exactly shoot the instance in the head
but it means we can spin up a test instance much easier.

Still WIP, there's lots of stuff still to do properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,135 +3,47 @@ Linux Vendor Firmware Service
 
 This is the website for the Linux Vendor Firmware Service
 
-IMPORTANT: This needs to be hosted over SSL, i.e. with a `https://` prefix.
+Missing firmware at LVFS
+------------------------
 
-## Using distro packages ##
+If your device is missing a firmware update that you think should be on LVFS
+please file an issue against this project and apply the Github label *missing-firmware*.
 
-    yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-    yum install \
-      bsdtar \
-      cabextract \
-      git \
-      httpd \
-      mariadb-server \
-      mod_wsgi \
-      MySQL-python \
-      python2-boto3 \
-      python2-gnupg \
-      python2-mysql \
-      python-flask \
-      python-flask-login \
-      python-flask-wtf
+Setting up the web service
+--------------------------
 
-Other useful packages might be:
-    yum install \
-      apachetop \
-      certbot-apache \
-      iftop \
-      mytop
+The official instance is set up using puppet on RHEL 7, on which you could use:
 
-If using RHEL-7 you can get python-flask-login using:
-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3,136 +3,48 @@ Linux Vendor Firmware Service
...
-## Using the CDN ##
-
-Create a `.aws/credentials` file like:
-
-    [default]
-    region=us-east-1
-    aws_access_key_id=foo
-    aws_secret_access_key=bar
+IMPORTANT: The LVFS needs to be hosted over SSL.
+If you want to use LetsEncrypt you can just do `certbot --nginx`.
```
superm1: You should mention here that the puppet configuration expects LetsEncrypt certs to be put in `/etc/letsencrypt/`

contrib/lvfs.pp
```diff
@@ -0,0 +1,338 @@
...
+    enable   => true,
+    require  => Package["munin"],
+}
+package { 'httpd-tools':
+    ensure => installed,
+}
+exec { "munin-htpasswd":
+    command     => "/usr/bin/htpasswd -cb /etc/munin/munin-htpasswd ${munin_username} ${munin_password}",
+    unless      => "/usr/bin/test -s /etc/munin/munin-htpasswd",
+    require     => [ Package["munin"], Package['httpd-tools'] ],
```
superm1: Does puppet let you specify alternatives?  httpd-tools doesn't exist in deb based.  It would be in `apache2-utils` on deb based installs.

superm1: If not, it's probably best to just mention in LVFS setup that the puppet script expects a rpm based distro with access to $packages

README.md
```diff
@@ -3,136 +3,48 @@ Linux Vendor Firmware Service
...
+--------------------------
+
+This instance is set up using puppet, so you can use something like:
+
+    wget -q https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+    yum install epel-release-latest-7.noarch.rpm
+    yum install puppet
+    hostname admin
+    puppet module install puppetlabs-vcsrepo --version 2.2.0
+    touch keys.pp
```
superm1: You probably want a step that's `cp contrib/keys.pp.in keys.pp` no?

README.md
hughsie: Right, agreed.

contrib/lvfs.pp
hughsie: I think it's fair to say the puppet script will only work with CentOS 7 -- and I should probably say that in the README. I don't particularly need people to set up their own instances, and now we have a testing server we can start to deploy to testing, and then deploy to production when we're 100% sure it's going to work (the proposal being Sunday afternoon we have any scheduled downtime). There's certainly more that would need changing for a Debian distro, even to the location of the nginx module files and that kind of thing.

README.md
hughsie: I wasn't sure how to explain this. Basically you have to copy the keys.pp.in file, then set up the variables to the right values.

PR #25, opened 2018-01-20
Move the CDN sync code to a new microservice

hughsie:
The idea here is to make the 'core' of the LVFS simpler and easier to test.

Gut feedback welcome. The logic was that mirroring files isn't anything really to do with the LVFS, and is only going to be needed in the one biggest instance. The alternative here was to create a plugin system for the LVFS, but that requires a lot more Python knowledge than I have. This sets up a tiny microservice here: https://github.com/hughsie/lvfs-sync -- ideas welcome.

The other bit thing I wanted to move out of the LVFS was the signing, nobody really understands Affidavit, and it only works with GPG. I think moving the signing out either to a plugin or a microservice would make the LVFS easier to audit.


hughsie:
This was way too complicated, and the polling was a PITA. See https://github.com/hughsie/lvfs-website/pull/26 for try 2.

PR #26, opened 2018-01-21
Wip/hughsie/plugin sync

hughsie:


app/views.py
```diff
@@ -259,27 +259,30 @@ def upload():
...
     try:
-        arc = _build_cab(fileitem.filename, data)
-    except cabarchive.CorruptionError as e:
+        if fileitem.filename.endswith('.cab'):
+            istream = Gio.MemoryInputStream.new_from_bytes(GLib.Bytes.new(data))
+            arc = GCab.Cabinet.new()
+            arc.load(istream)
+            arc.extract(None)
+        else:
+            arc = _build_cab(fileitem.filename, data)
```
superm1: At least when I was reading through this, it was confusing - but this is for handling alternate extensions.
Can you rename` _build_cab` to something like `_repackage_archive` and add some comments?

schema.sql
```diff
@@ -141,3 +140,15 @@ INSERT INTO settings (config_key, config_value)
   VALUES ('server_warning', 'This is a test instance and may be broken at any time.');
 INSERT INTO settings (config_key, config_value)
   VALUES ('firmware_baseuri', 'https://fwupd.org/downloads/');
+INSERT INTO settings (config_key, config_value)
+  VALUES ('cdn_sync_uri', 'https://s3.amazonaws.com/lvfstestbucket');
```
superm1: I think you should add something to README.md about updating this when setting up the instance.

superm1: And actually this URI doesn't seem to be used in code, is that intentional?

app/views.py
hughsie: Agreed, thanks.

schema.sql
hughsie: It is unused, I've removed it. I've not added anything about changing the database values in the README as all the plugin settings can be trivially changed in the admin-only "Server Setting" UI.

hughsie:
I think Python3 is a good idea. I'll have to check the gnupg status of python3, but with cabarchive gone it's actually now possible.

hughsie:
I did a quick run of 2to3 and there's quite a few places where we need to be more careful with the encoding. I'll merge this, and work on the Py3 stuff more tomorrow. Thanks for the review.

PR #28, opened 2018-01-24
Port to Python3

hughsie:


superm1:
Lgtm.

hughsie:
So, most of the required python3 packages are not available in centos 7.5 :/ I'll see what I can do with import future.

superm1:
That's too bad.  Are they easy to install from pypi maybe though?

PR #30, opened 2018-01-28
Add a plugin to add a README file to the archive

hughsie:
Ths idea is we explain that we're re-packed the archive from the vendor and
that the LVFS authors are not liable if you brick your computer.

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -327,6 +327,9 @@ def upload():
         flash('The firmware file had no .metadata.xml files', 'danger')
         return redirect(request.url)
 
+    # a good guess, but everyone should have this
+    fwupd_min_version = '0.8.0'
+
     # parse each MetaInfo file
     apps = []
     for cf in cfs:
@@ -410,6 +413,12 @@ def upload():
             flash('Firmware cannot specify vendor-id', 'danger')
             return redirect(request.url)
 
+        # does the firmware require a specific fwupd version?
+        req = component.get_require_by_value(AppStreamGlib.RequireKind.ID,
+                                             'org.freedesktop.fwupd')
+        if req:
+            fwupd_min_version = req.get_version()
+
         # add to array
         apps.append(component)
 
@@ -444,6 +453,15 @@ def upload():
 
         # allow plugins to sign files in the archive too
         ploader.archive_sign(arc, cfs[0])
+
+    # allow plugins to add files
+    settings = db.settings.get_all()
+    metadata = {}
+    metadata['$DATE$'] = datetime.datetime.now().replace(microsecond=0).isoformat()
+    metadata['$FWUPD_MIN_VERSION$'] = fwupd_min_version
+    metadata['$CAB_FILENAME$'] = new_filename
+    metadata['$FIRMWARE_BASEURI$'] = settings['firmware_baseuri']
+    ploader.archive_finalize(arc, metadata)
 
     # export the new archive and get the checksum
     ostream = Gio.MemoryOutputStream.new_resizable()
@@ -506,19 +524,17 @@ def upload():
         for req in component.get_requires():
             fwreq = FirmwareRequirement(AppStreamGlib.Require.kind_to_string(req.get_kind()),
                                         req.get_value(),
-                                        AppStreamGlib.Require.compare_from_string(req.get_compare()),
+                                        AppStreamGlib.Require.compare_to_string(req.get_compare()),
                                         req.get_version())
             md.requirements.append(fwreq)
 
         # from the first screenshot
         if len(component.get_screenshots()) > 0:
             ss = component.get_screenshots()[0]
-            if ss.caption:
-                md.screenshot_caption = ss.caption
-            if len(ss.images) > 0:
-                im = ss.images[0]
-                if im.url:
-                    md.screenshot_url = im.url
+            md.screenshot_caption = ss.get_caption(None)
+            if len(ss.get_images()) > 0:
+                im = ss.get_images()[0]
+                md.screenshot_url = im.get_url()
 
         # from the content checksum
         csum = rel.get_checksum_by_target(AppStreamGlib.ChecksumTarget.CONTENT)
--- a/plugins/info-readme/__init__.py
+++ b/plugins/info-readme/__init__.py
@@ -1,0 +1,70 @@
+#!/usr/bin/python2
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
+# Licensed under the GNU General Public License Version 2
+
+from __future__ import print_function
+
+import os
+import datetime
+
+from gi.repository import Gio
+from gi.repository import GLib
+from gi.repository import GCab
+
+from app.pluginloader import PluginBase, PluginError, PluginSettingText, PluginSettingBool
+from app import db, ploader
+from app.util import _archive_get_files_from_glob
+
+class Plugin(PluginBase):
+    def __init__(self):
+        PluginBase.__init__(self)
+
+    def name(self):
+        return 'Readme'
+
+    def summary(self):
+        return 'Add a README file to the archive.'
+
+    def settings(self):
+        s = []
+        s.append(PluginSettingBool('info_readme_enable', 'Enabled', False))
+        s.append(PluginSettingText('info_readme_filename', 'Filename',
+                                   'README.txt'))
+        s.append(PluginSettingText('info_readme_template', 'Template',
+                                   'plugins/info-readme/template.txt'))
+        return s
+
+    def archive_finalize(self, arc, metadata):
+
+        # get settings
+        settings = db.settings.get_filtered('info_readme_')
+        if settings['enable'] != 'enabled':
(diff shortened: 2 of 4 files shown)


plugins/info-readme/template.txt
```diff
@@ -0,0 +1,18 @@
+README FIRST
+============
+
+This firmware archive was signed by the LVFS on $DATE$.
+You can install this archive on supported hardware using the command:
+
+    fwupdmgr install this-file.cab
```
superm1: Other than legal stuff, I think the README can be more informative on how to use the package.
Eg:
* Include a link back to fwupd.org.
* If you're going to be changing `$DATE`, you may as well also fix the filename for installation at the same time instead of `this-file.cab`
* I suppose you may also have metadata that shows the minimum version of fwupd needed.  This is useful information to see in the README.

plugins/info-readme/template.txt
```diff
@@ -0,0 +1,18 @@
+README FIRST
+============
+
+This firmware archive was signed by the LVFS on $DATE$.
+You can install this archive on supported hardware using the command:
```
superm1: Since CAB is a Windows thing too you should make it clearer that this README is oriented to Linux information.

eg: 

> You can install this archive on Linux distributions with supported hardware using the command

plugins/info-readme/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+
+class Plugin(PluginBase):
+    def __init__(self):
+        PluginBase.__init__(self)
+
+    def name(self):
+        return 'Readme'
+
+    def summary(self):
+        return 'Add a REAME file to the archive.'
```
superm1: Typo

plugins/info-readme/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def settings(self):
+        s = []
+        s.append(PluginSettingBool('info_readme_enable', 'Enabled', False))
+        s.append(PluginSettingText('info_readme_filename', 'Filename',
```
superm1: Rather than markdown, can you make this extension `txt`?  I think for the purposes of opening the archive in places that might not have fwupd (such as Windows or on a Linux machine without fwupd) it will make it more apparent what can be done with it.

plugins/info-readme/template.txt
hughsie: Fixed, thanks.

plugins/info-readme/__init__.py
hughsie: Fixed, thx.

plugins/info-readme/__init__.py
hughsie: Sure can, thanks.

plugins/info-readme/template.txt
hughsie: What about something like this?

    README FIRST
    ============

    You can install this archive on Linux using fwupd >= $FWUPD_MIN_VERSION$ on
    supported hardware. Use this command to install:

        fwupdmgr install $CAB_FILENAME$

    You can check the version of fwupd using:

        fwupdmgr --version

    LEGAL NOTICE
    ============

    This firmware archive was signed by the LVFS on $DATE$ and was made available
    from $FIRMWARE_BASEURI$.

    The firmware file is provided "as is", without warranty of any kind, express or
    implied, including but not limited to the warranties of merchantability,
    fitness for a particular purpose and noninfringement.

    In no event shall the authors or copyright holders of the LVFS be liable for
    any claim, damages or other liability, whether in an action of contract, tort
    or otherwise, arising from, out of or in connection with the firmware archive.

superm1: LGTM

app/views.py
```diff
@@ -327,6 +327,9 @@ def upload():
         flash('The firmware file had no .metadata.xml files', 'danger')
         return redirect(request.url)
 
+    # a good guess, but everyone should have this
+    fwupd_min_version = '0.8.0'
```
superm1: I shed a tear for xenial which is still on 0.7.x :(

hughsie: I could use that if you'd prefer, but man, 0.7.x is old.

superm1: I know.  I think leave it as is.  Really should try to get everyone off 0.7.x.
Maybe SRU'ing 0.8.x is doable.  it's before the move to meson, still seems to compile on xenial.

superm1: I've backported a few more commits to 0_8_X.  Next time you spin a new 0_8_X release I'll try to get it SRU'ed.

hughsie: I've also just pushed some more. If 0_8_X looks good to you now I can spin you a new tarball tomorrow if that helps.

superm1: I think the only potential problem would be if gnome-software had to move at the same time as fwupd.  I don't remember the exact timing of that.  They're on a bit of an ancient version of that too (3.20.5) and that will be more of a mountain to move.

I'll file the appropriate paperwork once you release a 0.8.x release again.
Some EOL statement on 0.7.x I can link to would probably be useful too.

hughsie: > would be if gnome-software had to move at the same time 

I think it's compatible, but if you could check before I do the tarball I'd very much appreciate it.

> Some EOL statement on 0.7.x I can link to would probably be useful too.

Anything less that 0.8.x is certainly end of life. I wouldn't even do security updates on anything less than 0.8.x, and even 0_8_X is scraping the barrel.

superm1: I just did a quick smoke test on xenial with gnome-software from the archive and fwupd 0.8.x branch.

```
gnome-software --verbose 2>&1 | grep fwupd
(gnome-software:23025): Gs-DEBUG: opened plugin /usr/lib/gs-plugins-9/libgs_plugin_fwupd.so: fwupd
(gnome-software:23025): As-DEBUG: run GsPlugin::fwupd(gs_plugin_initialize)
(gnome-software:23025): As-DEBUG: run GsPlugin::fwupd(gs_plugin_setup)
(gnome-software:23025): As-DEBUG: run AsStore:app-info{/usr/share/app-info/xmls/org.freedesktop.fwupd.xml}
(gnome-software:23025): As-DEBUG: using icon path /usr/share/app-info/icons/fwupd
(gnome-software:23025): As-DEBUG: adding existing file: /usr/share/app-info/xmls/org.freedesktop.fwupd.xml
(gnome-software:23025): As-DEBUG: run AsStore:app-info{/var/cache/app-info/xmls/fwupd.xml}
(gnome-software:23025): As-DEBUG: using icon path /var/cache/app-info/icons/fwupd
(gnome-software:23025): As-DEBUG: adding existing file: /var/cache/app-info/xmls/fwupd.xml
(gnome-software:23025): GsPlugin-DEBUG: origin fwupd provides 12 apps (1%)
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to UEFI-dummy-dev0
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.via.VL811.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.via.VL811+.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.via.VL812.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.via.VL812_B2.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.dell.uefi124c207d.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.dell.uefi33773727.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.dell.uefi43ca3264.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.dell.uefie0f614ed.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.hughski.ColorHug.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.hughski.ColorHug2.firmware
(gnome-software:23025): GsPlugin-DEBUG: Adding keyword 'fwupd' to com.hughski.ColorHugALS.firmware
(gnome-software:23025): Gs-DEBUG: [enabled]     0.0     ->      fwupd
(gnome-software:23025): As-DEBUG: run 0x37322d0~GsPlugin::fwupd(gs_plugin_add_updates_historical)
(gnome-software:23025): As-DEBUG: run 0x37322d0~GsPlugin::fwupd(gs_plugin_add_updates)
```

It didn't explode there, so I'd say yeah it's got the right interface.

Moving to 0.9.x might be possible too with https://github.com/hughsie/fwupd/commit/ba5599fa9b5a10929fdcdaae2f8fa9baafd7fa44 but anything newer is blocked by the need to jump to newer appstream glib from https://github.com/hughsie/fwupd/commit/584921f16c3e37c00d874ba37c6c110bb62a556a

hughsie: Okay, release done. I think 0.9x might be a stretch for xenial given the constraints.

Issue #32, opened 2018-01-30
Overall Statistics view page for OEM use


superm1:
I've been discussing the new functionality that will be coming for success/fail telemetry in the future and feedback I've gotten is that we'll probably want an OEM specific statistics page we can see all of the updates aggregated for all updates so that trends could be observed for failures without having to drill down into each platform/update.

My thought is to look something like this:

**Overall downloads:**
| Total | | Success | Fail |
| ###    |  ###      | ### |

**Overall Breakdown:**
| Item | Total downloads | Total Success | Total Fail |
|ESRT XYZ | ###    |  ###      | ### |
|ESRT ABC | ###    |  ###      | ### |
|TBT NVM GUID 0x1234 | ###    |  ###      | ### |
|TBT NVM GUID 0x3456 | ###    |  ###      | ### |

**Downloads this month**
| Total | | Success | Fail |
| ###    |  ###      | ### |

**Breakdown this month**
| Item | Downloads | Success | Fail |
|ESRT XYZ | ###    |  ###      | ### |
|ESRT ABC | ###    |  ###      | ### |
|TBT NVM GUID 0x1234 | ###    |  ###      | ### |
|TBT NVM GUID 0x3456 | ###    |  ###      | ### |

* Update breakdown would aggregate the numbers across versions (to also capture downgrades).
* In the breakdown make the items link back to the device page to allow looking at the failures

hughsie:
> ESRT XYZ

Is this the device "name"?

> Update breakdown would aggregate the numbers across versions

Not sure I understand this; can you expand a bit please.

superm1:
> Is this the device "name"?

Yeah, device name makes more sense than device ID.

>>Update breakdown would aggregate the numbers across versions

>Not sure I understand this; can you expand a bit please.

So "Acme Box 123" has version 1.0.0 FW on LVFS.  version 1.1.0 is also released.  Want to add the # of downloads for both FW versions in the line item rather than just the newest (show aggregate downloads, aggregate success, aggregate fail)

hughsie:
I do think this is a useful thing, although I'd much rather get some more reports in first. At the moment we have... 4... I think waiting for people and distros to update to >= 1.0.4 might take some time. On that note, we will be "saving" the history even if it's not being reported straight away, so if gnome-software can start do this more automatically in the future they should flood in.

hughsie:
> we'll probably want an OEM specific statistics page

[...]

PR #35, opened 2018-02-01
Add some vendor telemetry data for each firmware file

hughsie:
Use this to update the database:

   ALTER TABLE firmware ADD download_cnt INT DEFAULT 0;
   CREATE INDEX filename_idx ON clients (filename(40));

Fixes https://github.com/hughsie/lvfs-website/issues/32

### DIFF
--- a/app/templates/telemetry.html
+++ b/app/templates/telemetry.html
@@ -1,0 +1,92 @@
+{% extends "default.html" %}
+{% block title %}Telemetry{% endblock %}
+
+{% block content %}
+<div class="container mt-3">
+
+<nav>
+  <ul class="nav nav-tabs">
+    <div><h3></h3></div>
+    <li class="nav-item">
+      <a class="nav-link {{'active' if age == 0}}" href="/lvfs/telemetry/0/{{sort_key}}/{{sort_direction}}">All</a>
+    </li>
+    <li class="nav-item">
+      <a class="nav-link {{'active' if age == 30}}" href="/lvfs/telemetry/30/{{sort_key}}/{{sort_direction}}">Last 30 days</a>
+    </li>
+    <li class="nav-item">
+      <a class="nav-link {{'active' if age == 180}}" href="/lvfs/telemetry/180/{{sort_key}}/{{sort_direction}}">Last 6 months</a>
+    </li>
+  </ul>
+</nav>
+
+<div class="container mt-3">
+
+{% if show_duplicate_warning %}
+<div class="alert alert-info" role="alert">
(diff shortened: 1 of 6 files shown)


app/templates/telemetry.html
```diff
@@ -0,0 +1,89 @@
...
+<div class="container mt-3">
+
+<h2>Telemetry for group {{group_id}}</h2>
+<div class="container mt-3">
+<table class="table">
+  <tr class="row">
+    <th class="col-sm-4">Hardware</th>
+    <th class="col-sm-2">
+      Version
+      <a href="/lvfs/telemetry/{{age}}/nameversion/up">▼</a>
```
superm1: If possible, can you make it clearer which way the page is currently sorted?  As it stands, the page renders the same no matter what sort you pick.

app/templates/telemetry.html
```diff
@@ -0,0 +1,89 @@
...
+{% endfor %}
+  <tr class="row">
+    <td class="col-sm-4"><b>&nbsp;</b></td>
+    <td class="col-sm-2"><b>&nbsp;</b></td>
+    <td class="col-sm-2"><b>{{total_downloads}}</b></td>
+    <td class="col-sm-2"><b>{{total_success}}</b></td>
+    <td class="col-sm-2"><b>{{total_failed}}</b></td>
+  </tr>
+</table>
+{% if show_duplicate_warning %}
```
superm1: Can you move this to the top of the page instead of the bottom?

app/views_firmware.py
```diff
@@ -436,6 +453,113 @@ def firmware_component_show(firmware_id, cid, page='overview'):
...
+            continue
+
+        # reports
+        if age == 0:
+            cnt_download = fw.download_cnt
+        else:
+            cnt_download = db.clients.get_firmware_count_filename(fw.filename, age)
+        cnt_success = 0
+        cnt_failed = 0
+        rpts = db.reports.get_all_for_firmware_id(fw.firmware_id, age=age, limit=999)
```
superm1: Why is this limited to 999?

app/templates/telemetry.html
hughsie: Fixed, now one of the arrows is always orange.

app/templates/telemetry.html
hughsie: Fixed.

app/views_firmware.py
hughsie: It was a hack. Fixed properly now.

Issue #36, opened 2018-02-02
Uploads are failing


superm1:
I've gotten word that pretty basic capsules are failing to upload.

CAB File:
```
$ gcab --list firmware.Latitude_5X90_1.1.8.wu.cab 
DriverPackage\firmware.bin
DriverPackage\firmware.inf
DriverPackage\firmware.metainfo.xml

```
This is the error:

> No firmware.bin found in the archive

Here's a sample:
https://dell.box.com/s/cgq4xqjs8le30rsrqghnv9yedr3759s6

hughsie:
Two things that stick out:

 * It's got path (can fix)
 * It's got two folders, which is a little unusual (although shouldn't be a problem)

I'll test and fix when I'm done with the school run.

hughsie:
Backslashes!

superm1:
I'm guessing this was after moving to appstream glib? The cab file should be same format that we've normally done cab files.

PR #37, opened 2018-02-02
Support cab archives that have paths as backslashes

hughsie:
This is a regression from when we switched to using GCab.

Refactor out the code into a new UploadedFile class that has unit tests, to
prevent this kind of thing happening again.

Fixes: https://github.com/hughsie/lvfs-website/issues/36


app/util.py
```diff
@@ -4,17 +4,23 @@
...
+import os
 import calendar
 import datetime
 from glob import fnmatch
 
 from flask import session, request, flash, render_template
 
+def _get_basename_safe(fn):
+    """ gets the file basename, also with win32-style backslashes """
+    return os.path.basename(fn.replace('\\', '/'))
```
superm1: Would using `ntpath `work better for this?  Although I suppose archives with the slashes the other way would break in the same way.

hughsie: Right, python seems to really want to stick to the current OS type. I didn't find a better way to do this, and it seemed harmless enough.

hughsie:
As an aside, do you know how to set up CI to run the tests for this trivial python case? I'm planning to add a lot more to cover the flask side of the app.

superm1:
I think https://docs.travis-ci.com/user/languages/python/ should cover exactly what you need.  No needs to run in docker or anything.

Regarding when this fix goes into production, is it immediate or when you restart LVFS this weekend? (2018-02-04 at 20:00 GMT according to fwupd.org)

hughsie:
I pushed it to the server just now as it was blocking you. At the weekend we should get a hardware upgrade all being well.

hughsie:
> No needs to run in docker or anything.

How's that going to work with the GObjectIntrospection requirement?

superm1:
Thanks.

I would think that you would just install the appropriate dependencies for gobject introspection during the "install" command.  If it does turn out to be too challenging on Ubuntu trusty (Travis CI env),  then a Docker container is obviously an option.

PR #39, opened 2018-02-05
Wip/hughsie/no auth with session

hughsie:
This is the 3rd version of this patch; the first added an `is_analyst` property to the User model, but had all kinds of messy fallbacks. The 2nd defined 'qa' and 'analyst' groups that users could be added to, but this turned into architecture astronaut territory, with "primary" groups for the user, and all that kind of mess.

This version does a lot of cleanups, then adds the property to the model, cleaning up the code a lot in the process. Comments welcome, especially as all this is security sensitive.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -7,6 +7,7 @@
 class UserCapability(object):
     Admin = 'admin'
     QA = 'qa'
+    Analyst = 'analyst'
     User = 'user'
 
 class User(object):
@@ -17,6 +18,7 @@ class User(object):
         self.display_name = None
         self.email = None
         self.is_enabled = False
+        self.is_analyst = False
         self.is_qa = False
         self.group_id = None
         self.is_locked = False
@@ -43,6 +45,16 @@ class User(object):
         # admin only
         if required_auth_level == UserCapability.Admin:
             if self.group_id == 'admin':
+                return True
+            return False
+
+        # analysts only
+        if required_auth_level == UserCapability.Analyst:
+            if self.group_id == 'admin':
+                return True
+            if self.is_qa:
+                return True
+            if self.is_analyst:
                 return True
             return False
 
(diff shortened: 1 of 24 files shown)


app/models.py
```diff
@@ -14,10 +18,61 @@ def __init__(self):
...
+        if required_auth_level == UserCapability.Admin:
+            if self.group_id == 'admin':
+                return True
+            return False
+
+        # analysts only
+        if required_auth_level == UserCapability.Analyst:
+            if self.group_id == 'admin':
+                return True
+            if self.is_qa:
```
superm1: I actually expected the inverse here - that QA should include analysts but analysts shouldn't include QA.

app/util.py
```diff
@@ -32,17 +32,12 @@ def _get_client_address():
 
 def _event_log(msg, is_important=False):
     """ Adds an item to the event log """
-    username = None
-    group_id = None
+    username = 'anonymous'
+    group_id = 'admin'
```
superm1: I'd think you want to default this to anonymous group_id still too.

app/models.py
hughsie: That's what I did, right? If the permission required is `Analyst` and you have `QA` then it returns True regardless.

app/util.py
hughsie: It certainly needs a comment; the idea is that only the admin can see things with the admin group_id. It only worked before due to dumb luck, rather than design.

hughsie:
Thanks for the review. If you send me (email) with any of the non-qa groups you want to make into analysts I can make that happen. Do you also need these changes to the LVFS stable instance straight away, or can it wait for the next Sunday maintenance window?

app/models.py
superm1: Yeah I guess I read it wrong. Thanks for confirming.

Issue #40, opened 2018-02-06
The WU files(*.inf and *.cat) will be deleted after LVFS submission


GavinChengGao:
Before LVFS submission, our original *.cab package will contain both WU(Windows Update) files(*.inf and *.cat) and LVFS files(*.metadata.xml and *.ROM). But after LVFS submission, LVFS will delete the WU files(*.inf and *.cat) and only leave the LVFS files with a new *.asc signature file generated. We saw this change today and before that it always worked well to contain both WU and LVFS files after LVFS submission. We and our customer all prefer to have both WU and LVFS files in one *.cab package even after LVFS submission. Hope that LVFS can change its policy to allow this.

PR #41, opened 2018-02-06
Add a plugin that allows copying .cat and .inf files from the source …

hughsie:
…archive

This allows a vendor to use the 'output' of Windows Update as the 'input' to
the LVFS and have one archive that can be used on either OS.

Fixes https://github.com/hughsie/lvfs-website/issues/40

### DIFF
--- a/plugins/wu-copy/__init__.py
+++ b/plugins/wu-copy/__init__.py
@@ -1,0 +1,34 @@
+#!/usr/bin/python2
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
+# Licensed under the GNU General Public License Version 2
+
+from app.pluginloader import PluginBase, PluginSettingBool
+from app import db, ploader
+from app.util import _get_basename_safe, _archive_add
+
+class Plugin(PluginBase):
+    def __init__(self):
+        PluginBase.__init__(self)
+
+    def name(self):
+        return 'Windows Update'
+
+    def summary(self):
+        return 'Copy files generated using Windows Update.'
+
+    def settings(self):
+        s = []
+        s.append(PluginSettingBool('wu_copy_inf', 'Include .inf files', True))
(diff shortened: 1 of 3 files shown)


superm1:
Just to play devil's advocate, why not just copy all files? It would be unfortunate if Microsoft introduced some new extension they used and we didn't catch it for a while etc.

hughsie:
> why not just copy all files

Several reasons:

 * We want to tightly control the files provided by the LVFS, in case a vendor included a nonfree `flash.exe` -- the legal advice we've got on this makes it clear that shipping this kind of not-explicitly-redistributable executable binary would make things somewhat more icky. I'm happy enough to argue that the .inf file was created by the vendor for distribution without a EULA, and that there's no inherent license on a .cat file as it's the binary output of a reproducible signing procedure.

 * We want to force the vendor to put the release notes in the metainfo file (or using the web interface) rather than build a `Release Nores.doc` file that's not designed to be user or machine readable.

 * We want to keep the size of the files as small as possible; more than one vendor shipped different variants of the firmware in the .cab archive, referencing only one in the metainfo file. By stripping out anything not in the metainfo that's a pretty easily understood case where the vendor actually needs more than one metainfo file in the .cab file, with a different GUID and appstream `<id>`.

superm1:
OK, make sense to me 👍

plugins/wu-copy/__init__.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/python2
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
+# Licensed under the GNU General Public License Version 2
+
+from __future__ import print_function
```
superm1: You don't actually use the print function function in this plugin

hughsie: Fixed, thanks. I wonder why pylint doesn't catch unused imports in files named `__init__.py`...

PR #42, opened 2018-02-06
Add CI

hughsie:
This works, and as it is using a centos7 base I'm comfortable it's testing things on the right library and python versions.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -1,0 +1,10 @@
+FROM centos:7
+
+RUN yum -y update
+RUN yum -y install epel-release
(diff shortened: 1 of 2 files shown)


contrib/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM centos:7
+
+RUN yum -y update
+RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
superm1: For CentOS 7 epel-release is already in the repo.  You can just change this to
```yum install -y epel-release```

Issue #49, opened 2018-02-07
Uploads with multiple XML files in archive under subdirectory failing


superm1:
I'm still having problems with MST CAB upload.  The last one fixed yesterday was built by Linux, this one is generated from Windows.

https://dell.box.com/s/gxflb2mo1wqdkdwtkkkf2vz10pjc20ax

```
$ gcab --list MSTFirmwareUpdateLinux_1.0.0.cab
DriverPackage\MST-62D-272.metainfo.xml
DriverPackage\MST-62E-272.metainfo.xml
DriverPackage\MST-62A-272.metainfo.xml
DriverPackage\MST-62B-272.metainfo.xml
DriverPackage\MST-62C-272.metainfo.xml
DriverPackage\MST-6DB-273.metainfo.xml
DriverPackage\MST-6DC-273.metainfo.xml
DriverPackage\MST-6DD-273.metainfo.xml
DriverPackage\MST-6DE-273.metainfo.xml
DriverPackage\MST-6DF-273.metainfo.xml
DriverPackage\MST-6E0-273.metainfo.xml
DriverPackage\MST-71D-277.metainfo.xml
DriverPackage\MST-71E-277.metainfo.xml
DriverPackage\MST-71C-277.metainfo.xml
DriverPackage\leaf_firmware_collan_Ver3_10_000.eeprom
DriverPackage\leaf_firmware_bc-pc_Ver3_10_003.eeprom
DriverPackage\leaf_firmware_mlk-houston_Ver3_10_001.eeprom
```

PR #50, opened 2018-02-07
Do not try to re-sign firmware referenced from multiple metainfo files

hughsie:
Fixes: https://github.com/hughsie/lvfs-website/issues/49

### DIFF
--- a/app/pluginloader.py
+++ b/app/pluginloader.py
@@ -46,6 +46,9 @@ class PluginBase(object):
     def settings(self):
         return []
 
+    def __repr__(self):
+        return "Plugin object %s" % self.id
+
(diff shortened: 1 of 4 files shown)


superm1:
While fixing this, I think it would also be useful to adjust the 50x error page to show more about what was causing the failure (for the next issue that is to come up for example).

hughsie:
You mean leaking the stack trace? That's usually frowned upon, and it's in the wsgi logs anyway.

hughsie:
Okay, new code deployed to fwupd.org now too.

superm1:
Oh as long as you're logging it somewhere useful, that's what's most important.

PR #55, opened 2018-02-11
Wip/hughsie/orm

hughsie:
Reviewing this in detail isn't really what I expect; it's a pretty huge change to the back-end of the LVFS webservice. What I do need is some review of the tests we're doing and the general approach; other ideas very welcome.


lvfs_test.py
```diff
@@ -0,0 +1,591 @@
...
+        fd = open(filename, 'rb')
+        return self.app.post('/lvfs/upload', data={
+            'target': target,
+            'file': (fd, filename)
+        }, follow_redirects=True)
+
+    def upload(self, target='private'):
+        rv = self._upload('contrib/hughski-colorhug2-2.0.3.cab', target)
+        assert b'com.hughski.ColorHug2.firmware' in rv.data, rv.data
+        assert b'e133637179fa7c37d7a36657c7e302edce3d0fce' in rv.data, rv.data
```
superm1: When testing upload functionality, I think it would be good to make sure that there is a test across the different variations that have been introduced or broken recently or can break.  There are apparently a ton of different ways they can be built.  Here's the ones that come to mind for me that it should be good to split up.

1. Good CAB file with everything in one block of the archive (XML/BIN)
2. CAB file that also has valid INF, in one block of archive (XML/BIN/INF)
3. Bad CAB file with invalid XML (XML, BIN)
4. CAB file with parts split across two blocks. (XML/INF in one, BIN in the other)
3. CAB file with 2x metadata, shared bin (2x XML, 1 BIN)
4. CAB file with 2x metadata, 2xbin (2x XML, 2 BIN).
5.  ZIP file

lvfs_test.py
```diff
@@ -0,0 +1,591 @@
...
+        assert b'2082b5e0-7a64-478a-b1b2-e3404fab6dad' in rv.data, rv.data
+
+        # check devices page shows private firmware as admin -- and hidden when anon
+        rv = self.app.get('/lvfs/device')
+        assert b'2082b5e0-7a64-478a-b1b2-e3404fab6dad' in rv.data, rv.data
+        rv = self.app.get('/lvfs/device/2082b5e0-7a64-478a-b1b2-e3404fab6dad')
+        assert b'MCDC04 errata' in rv.data, rv.data
+        self.logout()
+
+        # check private firmware isn't visable when not logged in
```
superm1: visible

lvfs_test.py
```diff
@@ -0,0 +1,591 @@
...
+        assert b'>stable<' in rv.data, rv.data
+        assert b'>testing<' not in rv.data, rv.data
+
+        # check it's now in the devicelist as anon
+        self.logout()
+        rv = self.app.get('/lvfs/devicelist')
+        assert b'ColorHug' in rv.data, rv.data
+        self.login()
+
+        # test deleting the firmware
```
superm1: I might have missed it, but I think you should explicitly check to make sure that only the right user groups can delete firmware.

lvfs_test.py
```diff
@@ -0,0 +1,591 @@
...
+
+        # promote the firmware to testing then stable
+        rv = self.app.get('/lvfs/firmware/e133637179fa7c37d7a36657c7e302edce3d0fce/promote/testing',
+                          follow_redirects=True)
+        assert b'>testing<' in rv.data, rv.data
+        assert b'>stable<' not in rv.data, rv.data
+        rv = self.app.get('/lvfs/firmware/e133637179fa7c37d7a36657c7e302edce3d0fce/promote/stable',
+                          follow_redirects=True)
+        assert b'>stable<' in rv.data, rv.data
+        assert b'>testing<' not in rv.data, rv.data
```
superm1: You should also include some testing that does demotion; specifically that the rules around who can promote/demote work.

superm1:
Per your comment only really reviewed the unit testing.  I think the approach is pretty good and you've got really good coverage.  It will make it much easier when "simple" new features are introduced to add the tests too.

Most of the tests are written from the perspective of an admin user.  Since there are now different user classes, it should be important to test each item that can (or can't) be done as that user class for the expected results.

lvfs_test.py
hughsie: I'm happy still using `uploadedfile_test.py` for this kind of parsing test -- but do you know to create archives with control over what goes in each folder? I guess I could use GObjectIntrospection from GCab to do that, but I'd rather something ready built.

superm1: I'm guessing it's an optimization that `makecab.exe` does when it encounters big files vs small ones.  If you create it from windows that's by default how it was getting spit out for me.  I'd think you can do a one time run from a Windows VM containing the metadata you want and some junk in the binary and then keep the binary CAB in tree under tests/ or something.

lvfs_test.py
hughsie: Fixed.

lvfs_test.py
hughsie: Valid, but I'll add this to the wip/hughsie/orm-users-own-firmware-only patch as it has all the user tests mostly in place for this extra test.

lvfs_test.py
hughsie: Just to clarify, is a non-qa users allowed to push firmware from `private` to `embargo`? At the moment they can't, but that doesn't make sense as we allow non-qa users to upload to either private or embargo.

superm1: Hmm, yeah I would expect that they should be able to move stuff back and forth from private and embargo.  I don't think Dell would use it, but someone else might.

hughsie: Fixed

lvfs_test.py
hughsie: I've added this to the other branch

PR #56, opened 2018-02-12
Wip/hughsie/orm users own firmware only

hughsie:


app/db.py
```diff
@@ -110,6 +110,33 @@ def modify_db(self):
...
+
+        # version 13 adds the uploader username to the firmware table
+        if int(setting.value) == 12:
+            print('Adding username to Firmware table')
+            self.engine.execute('ALTER TABLE firmware ADD username VARCHAR(40) DEFAULT NULL;')
+            setting.value = 14
+            self.session.commit()
+            return
+
+        # version 14 migrates the username using the event log
```
superm1: Why not do these in the same schema version?

app/db.py
```diff
@@ -110,6 +110,33 @@ def modify_db(self):
                     self.session.add(Guid(md.component_id, guid))
             setting.value = 12
             self.session.commit()
+            return
+
+        # version 13 adds the uploader username to the firmware table
+        if int(setting.value) == 12:
+            print('Adding username to Firmware table')
+            self.engine.execute('ALTER TABLE firmware ADD username VARCHAR(40) DEFAULT NULL;')
+            setting.value = 14
```
superm1: setting.value = 13

app/templates/firmware.html
```diff
@@ -7,7 +7,11 @@ <h1>Existing Firmware</h1>
 
 {% if fw_by_name|length == 0 %}
 <p>
+{% if g.user.check_capability('qa') %}
   No firmware has been uploaded to the &lsquo;{{g.user.group_id}}&rsquo; QA group yet.
+{% else %}
+  No firmware has been uploaded yet.
```
superm1: It would be clearer to say `"No firmware has been uploaded by the $NAME account yet"`

app/templates/useradmin.html
```diff
@@ -26,25 +26,25 @@ <h1>Details of user ‘{{u.username}}’</h1>
...
       <input class="checkbox" type="checkbox" name="is_locked" value="1" {{'checked' if u.is_locked}} />
-      Prevent password change:
+      Prevent password change
     </label>
   </div>
   <div class="form-group">
     <label for="is_analyst">
       <input class="checkbox" type="checkbox" name="is_analyst" value="1" {{'checked' if u.is_analyst}} />
-      Allow viewing analytics reports:
+      Allow read-only access to all firmware in group
```
superm1: While changing this how about "Allow read-only access to all firmware and associated reports in group"

app/db.py
hughsie: sqlalchemy restriction; it freaks out when the table changes underneath itself.

app/db.py
hughsie: Fixed, thanks.

app/templates/useradmin.html
hughsie: Fixed.

hughsie:
Pushed directly as I rebased, thanks for the review.

PR #57, opened 2018-02-13
Wip/hughsie/issues

hughsie:


app/templates/issue-details.html
```diff
@@ -0,0 +1,58 @@
...
+
+<nav>
+  <ul class="nav nav-tabs">
+    <li class="nav-item">
+      <a class="nav-link active" href="/lvfs/issue/{{issue.issue_id}}/details">Details</a>
+    </li>
+    <li class="nav-item">
+      <a class="nav-link" href="/lvfs/issue/{{issue.issue_id}}/conditions">Conditions</a>
+    </li>
+<!--
```
superm1: Probably can remove this now

app/templates/issue-conditions.html
```diff
@@ -0,0 +1,98 @@
...
+      <code>Regex</code>
+{% else %}
+      Comparison invalid!
+{% endif %}
+    </td>
+    <td class="col-sm-4">
+      <code>{{format_truncate(condition.value,39)}}</code>
+    </td>
+    <td class="col-sm-2">
+{% if g.user.check_for_issue(issue) %}
```
superm1: I'm not sure I follow the logic in having a delete button available to anyone but admin.  At least non-admin I think should be calling check_for_issue with readonly.

app/templates/issue-details.html
```diff
@@ -0,0 +1,58 @@
...
+    <label for="display_name">Description:</label>
+    <input type="text" class="form-control" name="description" value="{{issue.description}}" required {{'disabled' if not g.user.check_for_issue(issue)}}>
+  </div>
+  <div class="form-group">
+    <label for="enabled">
+      <input class="checkbox" type="checkbox" name="enabled" value="1" {{'checked' if issue.enabled}} {{'disabled' if not g.user.check_for_issue(issue)}}/>
+      Enabled
+    </label>
+  </div>
+{% if g.user.check_for_issue(issue) %}
```
superm1: I'm not sure I follow the logic in having a delete button available to anyone but admin.  At least non-admin I think should be calling check_for_issue with readonly.

app/views_issue.py
```diff
@@ -0,0 +1,206 @@
...
+    issue = db.session.query(Issue).\
+                filter(Issue.issue_id == issue_id).first()
+    if not issue:
+        flash('No issue found', 'info')
+        return redirect(url_for('.issue_conditions', issue_id=issue_id))
+    if not g.user.check_for_issue(issue):
+        return _error_permission_denied('Unable to add condition to report')
+
+    # already exists
+    if db.session.query(Condition).\
```
superm1: Is it not possible that the same key might be used in two directions?  Say gt "value" and lt "value"?  Maybe it's better to point at glob in that scenario.

app/templates/issue-details.html
hughsie: I was thinking of doing something like:

screenshot from 2018-02-13 16-38-25

app/templates/issue-details.html
hughsie: Other way round, we only show the delete button if you actually have privs to delete the issue.

app/templates/issue-details.html
superm1: Oh that can be useful, I see OK.

app/views_issue.py
hughsie: Yes, that's why I included the `glob` and `regex` comparison cases. Trying to group `AND` and `OR` in a web UI would be, well, unfun.

app/templates/issue-details.html
superm1: Oh it's the case that's not explicitly checked, I see.
The issue will be marked not readonly in that case.  Sorry for the noise.

app/templates/issue-details.html
hughsie: I've added that to this PR too.

hughsie:
> I'm amazed at how quickly you turn out quality stuff like this

I think it's something to do with not having meetings and committees. :)

PR #59, opened 2018-02-14
Wip/hughsie/priority

hughsie:


hughsie:
screenshot from 2018-02-14 12-24-15

superm1:
I think you also want to automatically want to put QA level issues higher priority than LVFS wide issues.  Eg if there is a more specific problem that needs to be matched.

hughsie:
You could argue the opposite, surely? That admin issues trump all others? With the shared priority model we have in this PR you can "negotiate" with the admin if a specific ordering is required.

superm1:
Yeah I guess you could argue both ways, let's see how it works and can raise issues if necessary as usage goes up.

PR #62, opened 2018-02-15
Wip/hughsie/block user agents

hughsie:


app/util.py
```diff
@@ -16,6 +16,39 @@
 from gi.repository import GCab
 from gi.repository import GLib
 
+def _vercmp(ver1, ver2):
```
superm1: You've already got version comparing from appstream glib no?  Why not use that?

app/views.py
```diff
@@ -11,17 +11,40 @@
...
+from .util import _error_internal, _error_permission_denied, _vercmp
+
+def _user_agent_safe_for_requirement(user_agent):
+
+    # very early versions of fwupd used 'fwupdmgr' as the user agent
+    if user_agent == 'fwupdmgr':
+        return False
+
+    # gnome-software/3.26.5 (Linux x86_64 4.14.0) fwupd/1.0.4
+    sections = user_agent.split(' ')
```
superm1: I think you can combine this and the bottom for loop for an optimization.

```
for chunk in sections:
    toks = chunk.split('/')
    if len(toks) == 2:
        if toks[0] == 'fwupd':
            return _vercmp('0.8.0', toks[1]) >= 0
        else if toks[0] == 'gnome-software':
            return _vercmp('3.26.0', toks[1]) >= 0
```

uploadedfile_test.py
```diff
@@ -75,6 +75,34 @@ def archive_sign(self, arc, firmware_cff):
             return
         _archive_add(arc, detached_fn, 'signed')
 
+class VersionTestCase(unittest.TestCase):
```
superm1: If you can end up using the appstream glib version compare you can probably drop this test case.

app/views.py
hughsie: Nahh, we can't unroll it as we want to match `fwupd/1.0.4` before `gnome-software/3.26.5` -- i.e. it's the last token. I suppose we could reverse the sections although that's a little magic.

app/util.py
hughsie: Ohh, and this is why you shouldn't code at 10pm. I'll use that indeed.

uploadedfile_test.py
hughsie: Fixed.

hughsie:
I'll merge this as I've got other things depending on this branch. Thanks all for comments so far. If Ubuntu could merge the 16.04 that'd make me feel warm and fuzzy.

PR #64, opened 2018-02-19
Rewrite a lot of the user-accessible website content

hughsie:
This is easier to review by visiting https://7bcca6c4-e943-426b-8039-4c024186ee68.pub.cloud.scaleway.com/


app/templates/developers.html
```diff
@@ -5,47 +5,78 @@
...
+  (gnome-software), <em>mechanism</em> (fwupd) and <em>data-provider</em> (LVFS)
+  and each can be disabled or replaced if required.
+</p>
 
 <h2>Getting the code</h2>
 <p>
-  The latest fwupd code is always available at <a href="https://github.com/hughsie/fwupd">GitHub</a>
-  and this is also the place to file <a href="https://github.com/hughsie/fwupd/issues">bugs or feature requests</a>.
-  You can trivially get the code by doing:
+  The LVFS is a simple Flask web service using a MySQL database and the code is
```
superm1: I think it would be better to put this and fwupd separably under `<h3>` tags.

app/templates/developers.html
```diff
@@ -5,47 +5,78 @@
...
-                   sqlite-devel gpgme-devel fwupdate-devel libsoup-devel
+$ git clone https://github.com/hughsie/fwupd.git
 </pre>
-
-<h2>Architecture</h2>
 <p>
-  The 'client' which is typically <code>gnome-software</code> or <code>fwupdmgr</code>
-  interacts with the daemon using D-DBus using PolicyKit.
-  The daemon has no network access and only acts as the mechanism for clients.
+  The fwupd project ships a number of plugins:
```
superm1: Rather than enumerate the plugins, I think it would be good to describe the plugin architecture and then refer to these as example plugins that are popular.  Also a link back to fwupd's plugins directory on Github might be good.

app/templates/developers.html
```diff
@@ -5,47 +5,78 @@
...
+<div class="card mt-3">
+  <h3 class="card-header list-group-item-info">Is updating firmware secure?</h3>
+  <div class="card-body">
+    <p class="card-text">
+      The LVFS and fwupd codebases have had several independant security audits.
+      The LVFS has a huge number of tests <a href="https://travis-ci.org/hughsie/lvfs-website">run for each commit</a>,
+      and fwupd has a <a href="https://travis-ci.org/hughsie/fwupd">comprehensive test suite</a>,
+      and is regularly scanned using both clang and <a href="https://scan.coverity.com/projects/10744">Coverity</a>.
+    </p>
+    <p class="card-text">
```
superm1: At least reading it to me, these two 'cards' should be swapped.  Eg reading about what crypto is used followed by coverage for the code.

app/templates/developers.html
```diff
@@ -5,47 +5,78 @@
...
+      On both the LVFS and using fwupd, GPG crypto is being performed using GnuPG and PKCS#7 crypto is using GnuTLS.
+      There have been no known security issues using this architecture.
+      The fwupd daemon has no network access and only acts as the mechanism for
+      clients using D-DBus and PolicyKit.
+    </p>
```
superm1: It is probably worth mentioning that the authenticity of the updates being flashed relies upon signature verification implemented by the device manufacturer (eg Dell checks the signature on all UEFI payloads before flashing and Logitech's plugin checks the signature on those payloads etc).

app/templates/index.html
```diff
@@ -21,14 +21,27 @@ <h1 class="display-3">Linux Vendor Firmware Service</h1>
...
 </div>
 
 <div class="container">
-  <!-- Example row of columns -->
+
+  <div class="card mb-3">
+    <h3 class="card-header list-group-item-warning">
+      Looking for firmware to fix Meltdown and Spectre?
+      <img src="/img/meltdown.min.svg" alt="meltdown logo" height="32"/>
+      <img src="/img/spectre.min.svg" alt="spectre logo" height="32"/>
```
superm1: IIRC there are no firmware fixes for meltdown, only for spectre. Am I wrong?

app/templates/metainfo.html
```diff
@@ -0,0 +1,137 @@
...
+  ID, much like Microsoft update:
+</p>
+<pre>
+  &lt;!-- only newer versions of fwupd understand 'hardware' requirements --&gt;
+  &lt;requires&gt;
+    &lt;id compare="ge" version="1.0.1"&gt;org.freedesktop.fwupd&lt;/id&gt;
+    &lt;hardware&gt;6de5d951-d755-576b-bd09-c5cf66b27234&lt;/hardware&gt;
+  &lt;/requires&gt;
+</pre>
+<p>CHIDs can also be added or removed in the LVFS admin panel.</p>
```
superm1: It's not called the admin panel though right?
I think it would be good to explain where to access this.  It's gotten buried a lot deeper in since it was first added.

Also it's worth mentioning that changes can only be made before the firmware is published to `stable` channel.

app/templates/users.html
```diff
@@ -7,12 +7,29 @@
...
 </p>
+
+<div class="card mb-3">
+  <h3 class="card-header list-group-item-info">Is my hardware supported?</h3>
+  <div class="card-body">
+    <p class="card-text">
+      The LVFS only provides firmware once the original OEM vendor has joined
+      the project and uploaded files themselves.
+      The OEM vendor is in full control over what models are supported and is the
+      only person that can add support for new hardware.
```
superm1: s/person/entity/

app/templates/users.html
```diff
@@ -21,97 +38,71 @@ <h1>Introduction</h1>
   updates, the command-line tool or the system D-Bus interface directly.
```
superm1: I don't think you want people using D-Bus directly any more right?

app/templates/users.html
```diff
@@ -21,97 +38,71 @@ <h1>Introduction</h1>
...
-
-<h2>Using the command line</h2>
-<p>
-  fwupd ships a command line <code>fwupdmgr</code> program.
-  This allows administrators to get the list of upgradable devices,
-  schedule offline updates or installing firmware on the live system.
-  You can manually download released firmware updates from the LVFS
-  <a href="/lvfs/devicelist">device list</a>.
-</p>
 <pre class="prettyprint">
```
superm1: I think you want the fwupdmgr command line tool under it's own `<h2>` to match what you did for gnome-software

app/templates/vendors.html
```diff
@@ -6,199 +6,77 @@
...
+<div class="card mb-3">
+  <h3 class="card-header list-group-item-info">Why does the LVFS project sign the archive?</h3>
+  <div class="card-body">
+    <p class="card-text">
+      The upload process repacks the uploaded archive into a new cabinet file
+      and signs the firmware image using a detached GPG or PKCS#7 signature
+      so client tools can be sure the firmware actually originated from the LVFS.
+      Any existing Windows Update signatures are also copied into the new
+      archive although are not used on Linux.
+    </p>
```
superm1: You should also mention that the archive is renamed with a random string at the beginning to avoid clashes with other uploads.

hughsie: Fixed.

app/templates/developers.html
hughsie: Something like `Various plugins are included to update devices, which includes native support for UEFI, DFU, Unifying, ThunderBolt and ColorHug.` ?

app/templates/developers.html
hughsie: Fixed.

app/templates/developers.html
hughsie: How about getting rid of `There have been no known security issues using this architecture.`
 and adding `Some devices also have additional hardware signature verification schemes implemented by the device manufacturer.`?

app/templates/index.html
hughsie: You're completely right, fixed.

app/templates/users.html
hughsie: Fixed.

app/templates/users.html
hughsie: It is something I get asked every now and then; people are wanting to use the DBus interface from Cockpit for instance.

app/templates/developers.html
superm1: Yeah that's better.

app/templates/developers.html
superm1: Yep that's good.

app/templates/users.html
superm1: I would think it makes more sense to tell them to wrap fwupdmgr.  There are so many checks that happen in it and what libfwupd do.

hughsie: They really wanted to use D-Bus. I might be able to get them to use libfwupd using GObjectIntrospection, but they really won't like parsing another tool output.

superm1: I think GObjectIntrospection seems like the smarter way to encourage people to adopt if they're going to speak directly to the daemon.

app/templates/users.html
superm1: Actually maybe it's best to kick the command line stuff over to a wiki page or README.md for fwupd?

hughsie: Wikis get out of date pretty quickly. Is there any way to generate an html page from the `help2man` script or something?

superm1: Maybe a github IO page?  it seems like you might be able to commit directly to the repository with it (Pick project site)
https://pages.github.com/

hughsie: Yes, I'll look into this.

PR #66, opened 2018-02-22
Don't store the report data as a JSON blob

hughsie:
I did originally think that the JSON would just be useful for read-only review,
until we started "fixing up" reports as conditions were added.

Parsing a large blob of JSON blobs when we have thousands of reports just isn't
going to scale. Split out a new table that we can query directly.

To migrate, do:

    ALTER TABLE reports CHANGE id report_id INT NOT NULL AUTO_INCREMENT PRIMARY KEY;

Then do:

    FLASK_APP=app/__init__.py flask-2 modifydb

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -517,40 +517,75 @@ def _get_flat_dict_from_json(txt):
             for key2 in items2:
                 data[key2] = items2[key2]
             continue
-        data[key] = items[key]
+        data[key] = unicode(items[key]).encode('ascii', 'ignore')
     return data
+
+class ReportAttribute(db.Base):
+    __tablename__ = 'report_attributes'
+    report_attribute_id = Column(Integer, primary_key=True, nullable=False, unique=True)
+    report_id = Column(Integer, ForeignKey('reports.report_id'), nullable=False)
+    key = Column(Text)
+    value = Column(Text)
+
+    # link back to parent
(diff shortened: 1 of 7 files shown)


app/models.py
```diff
@@ -517,40 +517,75 @@ def _get_flat_dict_from_json(txt):
...
-    id = Column(Integer, primary_key=True, nullable=False, unique=True)
+    report_id = Column(Integer, primary_key=True, nullable=False, unique=True)
     timestamp = Column(DateTime, nullable=False, default=datetime.datetime.utcnow)
     state = Column(Integer, default=0)
-    json = Column(Text)
+    unsed_json = Column('json', Text)
     machine_id = Column(String(64), nullable=False)
     firmware_id = Column(Integer, ForeignKey('firmware.firmware_id'), nullable=False)
-    checksum = Column(String(64), nullable=False)
+    checksum = Column(String(64), nullable=False) #fixme remove?
```
superm1: I would just drop the dead variables.  If they're not being used why waste the memory to initialize them (especially at scale).

hughsie: That's the patchset part 2; you have to do the `unused_` dance so you can use the existing data in the migration script. I've not been using something like alembic as the database schema was far from ideal (e.g. some tables lacking primary keys) but now we're getting close to be able to do proper migrations up and down.

PR #70, opened 2018-02-23
Create a table to log when firmware moves between different targets

hughsie:
I'm not super happy with the "Promotions" name -- any better ideas welcome. It's really a table that hold what user moved the firmware to different targets, and when they did it.

### DIFF
--- a/app/db.py
+++ b/app/db.py
@@ -35,12 +35,67 @@ class Database(object):
     def modify_db(self):
 
         # get current schema version
-        from .models import Setting
+        from .models import Setting, Event, FirmwareEvent, Firmware, User
         setting = self.session.query(Setting).filter(Setting.key == 'db_schema_version').first()
         if not setting:
             print('Setting initial schema version')
             setting = Setting('db_schema_version', str(0))
             self.session.add(setting)
+
+        if int(setting.value) == 20:
+            print('Creating firmware_events table')
+            self.Base.metadata.tables['firmware_events'].create(bind=self.engine, checkfirst=True)
+            setting.value = 21
+            self.session.commit()
+            return
+
+        if int(setting.value) == 21:
+            print('Getting all Firmwares')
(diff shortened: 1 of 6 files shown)


hughsie:
screenshot from 2018-02-23 20-59-12

hughsie:
Ideas while sitting on a bus: Note, FirmwareNote, FirmwareEntry, FirmwareEvent, FirmwareRecord, FirmwareStatement, FirmwareNote, FirmwareAction or FirmwareStatus.

superm1:
For a table name `channel_movements`?

app/db.py
```diff
@@ -35,13 +35,68 @@ def init_app(self, app):
...
+                    spl = e.message.split(' ')
+                    if len(spl) != 5:
+                        continue
+                    filename = spl[2]
+                    if filename not in fws_filename:
+                        print('filename %s does not exist, skipping' % filename)
+                        continue
+                    if e.username not in users:
+                        if e.username == 'dell':
+                            user_id = users['dellqa'].user_id
```
superm1: Is dell a special case I guess?

app/models.py
```diff
@@ -360,6 +361,31 @@ def find_req(self, kind=None, value=None):
     def __repr__(self):
         return "Component object %s" % self.appstream_id
 
+class Promotion(db.Base):
```
superm1: If you're not crazy about `Promotion`, how about `Channel`?

app/db.py
hughsie: Yup, when we started we had no user/qa split. Just Dell that's affected as it was the first real user.

superm1: Probably worth leaving a comment in the code around that.

hughsie: This is the hacky db migration code; it'll be gone in a few commits after being deployed. :)

superm1: OK :)

PR #72, opened 2018-02-25
Move the login entries into a modal dialog

hughsie:
Advice welcome for the text above the username/password entries. I wanted to make it  clear that normal users don't have to log in.

### DIFF
--- a/app/templates/default.html
+++ b/app/templates/default.html
@@ -93,11 +93,7 @@
             </div>
           </li>
 {% else %}
-        <form method="POST" class="form-inline" action="/lvfs/login">
-          <input class="form-control mr-sm-2" type="text" name="username" placeholder="Username" aria-label="Username" required>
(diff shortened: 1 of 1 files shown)


app/templates/default.html
```diff
@@ -114,6 +110,42 @@
...
+      <div class="modal-header">
+        <h5 class="modal-title" id="loginModalLabel">Log in to the LVFS?</h5>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <form method="post" action="/lvfs/login">
+        <div class="modal-body">
+          <p>
+            Only OEMs and ODMs are permitted to access the LVFS.
```
superm1: How about:
"No login is needed to download firmware.  Accounts are only used for OEM/ODM use when uploading new firmware.
If you have an account please log in below"

hughsie:
What about this? 
screenshot from 2018-02-26 17-36-09

superm1:
Looks great.

Issue #71, opened 2018-02-23
Feature request: "Search" box for platforms


superm1:
I'm sharing a request that's come up internally to let people find platforms/firmware by a search box at the top of the home page.

Visiting /devicelist and searching the page is a bit too difficult was the feedback.

hughsie:
Some initial thoughts:

 * "people" as in random users looking for firmware, or just OEMs with logins?
 * What would be searched? Component <name>'s seem obvious, but what about release notes? GUIDs?
 * Would searching for "foo bar" be an AND search or an OR search?
 * Would we expect "xps" to match a keyword of "xps13"?
 * Would we expect "13" to match "xps13"?
 * Would the results be 'components' or firmware names?
 * What would the ordering of results be?
 * Would we index things in testing+stable, or just do everything and then filter results based on user level?
 * Where would the search entry go? In the page header or in the devicelist page? If the former, will uses expect a google-like search of all the pages on the site? If the latter, we could move the Login/Password entries into a modal behind a [Login] button.

In the coming months we've got quite a few OEMs coming onboard, so the number of devices and firmwares is only going to increase. If nothing else, I'd find this kind of search helpful.

superm1:
I've also noticed /devicelist is taking a longer time to load lately.

For your questions:

* People is "random" people.
* For the need asked for a box with a watermark "Search platforms" would be adequate. I don't think you need to search other fields for now.
* Yes I'd expect XPS to match xps13 
* I wouldn't expect 13 to match xps13.
* I think firmware names would be more obvious.
* I think ordering should be best match
* Yeah I think this should be based on whether logged in or not
* On the header is where it was desired. I think you can water mark the search box with something descriptive to make it clear what it's providing.

hughsie:
Any initial comments? I'd like to polish things up a bit before we get real world testing.

superm1:
It looks like it's just matching title, which seems fine to me.
I don't think it's necessarily useful to strike through the unfound results.
If it's not a performance hit it would be useful to show "Title" and latest version in the table of results.

hughsie:
Okay, comments welcome: https://7bcca6c4-e943-426b-8039-4c024186ee68.pub.cloud.scaleway.com/lvfs/search?value=xps+13

superm1:
Feedback so far:

*Searching for the 3040 is giving different results each time.
*The firmware version returned isn't always the newest.

hughsie:
Both "fixed" but need tests.

[...]

PR #74, opened 2018-02-26
Allow searching for firmware using defined keywords

hughsie:
To migrate, do:

    FLASK_APP=app/__init__.py flask-2 modifydb

Fixes https://github.com/hughsie/lvfs-website/issues/71


app/templates/devicelist.html
```diff
@@ -33,6 +33,14 @@ <h1>Devices using LVFS for firmware updates</h1>
...
         please see the <a href="/vendorlist">vendor engagement list</a>.
       </p>
+      <form method="GET" class="form-inline mb-3" action="/lvfs/search">
+        <div class="input-group">
+          <input type="text" class="form-control" aria-label="search" name="value" required>
+          <div class="input-group-append">
+            <button class="input-group-text" type="submit">Search</button>
+          </div>
+        </div>
+      </form>
```
superm1: Could you also keep this on the main home page?

app/views_component.py
```diff
@@ -146,3 +169,65 @@ def firmware_requirement_add():
...
+    # log
+    flash('Removed keyword %s' % kw.value, 'info')
+    return redirect(url_for('.firmware_component_show',
+                            component_id=md.component_id,
+                            page='keywords'))
+
+@app.route('/lvfs/component/keyword/add', methods=['POST'])
+@login_required
+def firmware_keyword_add():
+    """ Modifies the update urgency and release notes for the update """
```
superm1: comment should be updated for what this is

app/templates/devicelist.html
hughsie: I did originally have it there -- I moved it for two main reasons:

 * It doesn't work very well on a small screen (i.e. it forces the top bar into 2-3 "rows")
 * It's not really good enough yet

On the latter point you might notice I'm logging the search string and the number of results; after a few weeks deployed we can see how real users are using this and fix things accordingly. I think putting it so "front and center" it needs to be much better than it is now.

superm1: OK.

app/models.py
```diff
@@ -294,6 +294,61 @@ def __init__(self, component_id=None, value=None):
...
+                 'for',
+                 'limited',
+                 'system',
+                 'the',
+                 'update']:
+        return False
+    return True
+
+def _sanitize_keyword(value):
+    for rpl in ['(', ')', '[', ']']:
```
superm1: You probably want to look for escape characters i'd think too right?

hughsie:
Okay, merged for wider feedback. Check out https://fwupd.org/lvfs/search for the first cut.

hughsie:
Some initial results:
screenshot from 2018-03-01 11-24-03

One thing that's obvious from the logs is that some people are searching for software, e.g. "Ubuntu" -- we should do something more useful there to help them.

Issue #77, opened 2018-03-01
LVFS should enforce firmware files have urgency set on upload


hughsie:
This allows us to prioritize the notification in gnome-software. An alternative would be to require the urgency is set in the web-ui before moving to testing or stable.

superm1:
I would prefer to see a modal dialog come up when moving to testing/stable, something like this:

**Missing urgency**
This firmware doesn't have an urgency level set.
Please choose the urgency level for this firmware.
1. Required
2. Install when convenient

The remaining problem would be uploads that are uploaded directly to testing/stable.  I don't know how common that is, but the obvious solution is to in the POST have an optional urgency field and if the binary doesn't have urgency set throw an error and complain about the need for that field to be set during upload.

hughsie:
> would be uploads that are uploaded directly to testing/stable

I've been wondering about disabling that for a while now. It's probably not a very good idea generally.

superm1:
Can you check stats if much gets uploaded directly there before making that kind of change?

hughsie:
> Can you check stats if much gets uploaded directly there before making that kind of change?

I can, but tomorrow. I guess there's not much point removing it as an option if a vendor (think smaller vendor, only the one qa user) can just upload to embargo, and then move straight away to testing, then straight away to stable. I guess we could enforce the amount of time it spends in testing (say 1 day?) using the new FirmwareEvent thing we added the other day. I think allowing upload to testing is probably okay to keep, but maybe not stable.

hughsie:
> Enforcing a migration period might just slow things down needlessly.

Yes, perhaps you're right. I do think a warning is a good idea, but a warning that can be ignored.

hughsie:
Something like this?

![screenshot from 2018-03-02 11-07-38](https://user-images.githubusercontent.com/151380/36896331-27cda8dc-1e0a-11e8-9294-4935fe36e722.png)

hughsie:
and

![screenshot from 2018-03-02 13-47-53](https://user-images.githubusercontent.com/151380/36901823-557ed740-1e20-11e8-983b-374f665e9290.png)

PR #79, opened 2018-03-02
Require urgency and release notes before moving to stable

hughsie:
Fixes https://github.com/hughsie/lvfs-website/issues/77

### DIFF
--- a/app/templates/firmware-details.html
+++ b/app/templates/firmware-details.html
@@ -3,8 +3,60 @@
 
 {% block content %}
 <div class="container mt-3">
-<h1>Firmware Details</h1>
 
+<!-- veto information -->
+{% if fw.target == 'testing' and 'no-release-urgency' in vetos %}
+<div class="card mb-3">
+  <h3 class="card-header list-group-item-warning">
+    More information soon required!
+  </h3>
+  <div class="card-body">
+    <p class="card-text">
+      All components should have an appropriate update urgency before a
+      firmware is moved to stable.
+      You can do this for each component on this firmware by clicking <b>Details</b>
+      &raquo; <b>Update Details</b> and setting an urgency from the dropdown.
+    </p>
+    <p class="card-text">
+      For future firmware uploads, this can be set automatically using
+      <code>&lt;release urgency="high"&gt;</code> in the
+      <a href="/metainfo"><code>.metainfo.xml</code></a> file.
+    </p>
+    <p class="card-text">
+      From 1<sup>st</sup> September 2018 it will not be possible to move firmware
+      files to stable that do not have an urgency value.
+    </p>
+  </div>
+</div>
+{% elif fw.target == 'testing' and 'no-release-description' in vetos %}
+<div class="card mb-3">
+  <h3 class="card-header list-group-item-warning">
+    More information soon required!
+  </h3>
+  <div class="card-body">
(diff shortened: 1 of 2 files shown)


app/templates/firmware-details.html
```diff
@@ -3,8 +3,50 @@
...
+  <div class="card-body">
+    <p class="card-text">
+      All components need an update urgency before a firmware can be moved to stable.
+      You can do this for each component on this firmware by clicking <b>Details</b>
+      &raquo; <b>Update Details</b> and setting an urgency from the dropdown.
+    </p>
+    <p class="card-text">
+      For future firmware uploads, this can be set automatically using
+      <code>&lt;release urgency="high"&gt;</code> in the
+      <a href="/metainfo"><code>.metainfo.xml</code></a> file.
```
superm1: Why not show the urgency drop down in the card rather than make them run around to do it?

app/templates/firmware-details.html
```diff
@@ -3,8 +3,50 @@
...
+      You can do this for each component on this firmware by clicking <b>Details</b>
+      &raquo; <b>Update Details</b> and writing some content that is shown to users.
+      Writing good quality release notes are really important as some users may be
+      worried about an update that does not explain what it fixes.
+    </p>
+    <p class="card-text">
+      For future firmware uploads, this can be set automatically using this
+      in the <a href="/metainfo"><code>.metainfo.xml</code></a> file:<br/>
+      <code>&lt;release&gt;&lt;description&gt;&lt;p&gt;Release note
+      text&lt;/p&gt;&lt;/description&gt;&lt;/release&gt;</code>
```
superm1: At least for Dell this is going to break workflow.  Release notes aren't required for capsules today.  I can talk to people about including them, but can't force it.
Can you make this a warning but not block the button?

hughsie: So, to clarify you're submitting updates to stable with no update description at all? How is the user supposed to know what the updates fix? My fix here only requires the release notes on the push-to-stable, rather than on the upload, and so can be fixed-up by the QA user using the admin panel.

app/templates/firmware-details.html
hughsie: Are we assuming all the components in the firmware all have the same priority?

superm1: I thought that was a pretty reasonable assumption. I can't think of as situation that we would ever for example say that in like a dock update the "EC" was optional but the MST was compulsory.  We would just advertise you need to take dock update package XYZ.

app/templates/firmware-details.html
superm1: Yeah folks who promote to stable are not consistently filling out update descriptions. You can check most of the "Dell" files and see they're missing.

I'd like to solve that but a hammer isn't going to go over well.  How about make it a warning with on "X/Y/ZZZZ" this is going to be mandatory before promoting to stable?  I can start some discussions but need time for such a process change to be adopted.

hughsie: > How about make it a warning with on "X/Y/ZZZZ" this is going to be mandatory before promoting to stable?

This seems a reasonable compromise. How about 1st September 2018?

superm1: I think that's a reasonable date.  I'll start to socialize it.

Issue #78, opened 2018-03-01
LVFS should fail to upload files with incomplete release notes


hughsie:
At least one vendor is doing `Please refer the changes.new for the changes.` which of course isn't included, and really isn't visible in gnome-software. The alternative, like issue 77 is that we enforce they are set before moving to testing and stable.

I'm not sure how we'd detect the release notes are "redirected" -- ideas welcome. Maybe look for any "words" in the text that are also filenames in the uploaded cab file?

PR #80, opened 2018-03-02
Do not allow release notes to reference missing files

hughsie:
This prevents vendors doing things like 'Please refer to README.txt'.

Fixes https://github.com/hughsie/lvfs-website/issues/78

### DIFF
--- a/app/uploadedfile.py
+++ b/app/uploadedfile.py
@@ -212,6 +212,16 @@ class UploadedFile(object):
         if not release_default:
             raise MetadataInvalid('The metadata file did not provide any releases.')
 
+        # ensure the update desciption does not refer to a file in the archive
+        release_description = release_default.get_description()
+        if release_description:
+            for word in release_description.split(' '):
+                if word.find('.') == -1: # any word without a dot is not a fn
+                    continue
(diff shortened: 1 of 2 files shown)


app/uploadedfile.py
```diff
@@ -212,6 +212,16 @@ def _load_metainfo(self, cf):
         if not release_default:
             raise MetadataInvalid('The metadata file did not provide any releases.')
 
+        # ensure the update desciption does not refer to a file in the archive
+        release_description = release_default.get_description()
+        if release_description:
+            for word in release_description.split(' '):
```
superm1: This approach is likely to cause false positives on URL's with periods isn't it?  Or references that were like this: "Upgrading from 28.0 may require a power cycle of the controller to prevent problems with xyz"

hughsie: We check if the 'dotted' word is a file in the archive, so I think it should be pretty safe.

superm1: That's what I thought but I got thrown off by our error.  Shouldn't it be 

> "the release notes shouldn't reference other files"

Because you're checking explicitly for files mentioned in release notes that are in the archive.

PR #81, opened 2018-03-02
trivial: Try to convince people to use the testing target

hughsie:
Show a warning when trying to promote a firmware into stable that has spent
less than a day in testing.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -486,6 +486,12 @@ class Firmware(db.Base):
     # link using foreign keys
     vendor = relationship('Vendor', foreign_keys=[vendor_id])
 
+    @property
+    def target_duration(self):
+        if not self.events:
+            return 0
+        return datetime.datetime.utcnow() - self.events[-1].timestamp
+
     def __init__(self):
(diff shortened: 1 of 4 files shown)


superm1:
I'd argue you should also let time in the embargo target also count for this.

hughsie:
It's tricky as embargo is only available to the vendor, so the number of testers are going to be counted on one hand. Maybe we can tie in the number of downloads too?

superm1:
At least with our use on uefi the testing is mostly done offline before it even hits LVFS. I'd think that mirrors at least some other vendors use too.

superm1:
Once reports are coming in more consistently (maybe after next Ubuntu and Fedora releases out) could be to only put in testing and automatically promote to stable after some X number of reports with at least Y% stable. Or vice versa to pull back to testing after so many failed reports with an email / issue to say go investigate why there are so many failures.

hughsie:
Okay, I've scaled back this down to just informational -- you're right we can do something better when the reports are flooding in.

Issue #82, opened 2018-03-05
Drop usernames and use email addresses everywhere


hughsie:
At the moment we have three user identifiers used in the database, the `username`, the `email` and the `user_id` integer primary key. When we add OAuth we'll be adding a fourth, an `oauth_id` string as well. This is getting quite confusing and unwieldy in the code, for something that is supposed to be transparent and easy to understand.

My proposal would be to stop using the username for login, and use the email as the user-visible string to represent the user representing the vendor. This causes problems where the email is associated with many different usernames, e.g. `vendor_qa` and `vendor_$odm`. In this case we could use a "fake" email address, e.g. existing_email+foxconn@existing_domain.com and use that for login.

I guess we'd also support using the old username as the way to log in (with an appropriate warning message?) for some time to allow ODMs and OEMs to update any process instructions.

@superm1 Comments?

hughsie:
>When we add OAuth we'll be adding a fourth, an oauth_id string as well

Actually, we can do without the oauth_id if we use userPrincipalName (the email address) although the same problem applies when more than one user "shares" an email address.

superm1:
Yeah something like this will definitely take some time for adjustment to stopping logging in using an ID, so this would need some transitional time period that old logins via username still worked like you're doing for description/urgency.

> Actually, we can do without the oauth_id if we use userPrincipalName (the email address) although the same problem applies when more than one user "shares" an email address.

WIth OAuth how can that happen?  I guess by user you mean "LVFS user account"?  Like currently i'm same contact email for some accounts that are ODM upload only.

I do think switching to "email" logins makes sense though when OAuth is turned on. It is more logical then.  How about for those existing ODM accounts that are upload only just change them over to fake email addresses?
Eg `compal@dell.com`.  Then for some time period allow logging in with: 

- the old account name (Show a warning)
- that fake email (Don't show anything)
- oauth (Don't show anything)

hughsie:
> just change them over to fake email addresses?

This is fine if the LVFS is authenticating old-school style. In the case of using OAuth the userPrincipalName has to match the "fake" email address. So if you've got a user 'Alice' who's trying to log in using alice@dell.com that's not going to match the fake compal@dell.com account and that's going to fail.

[...]

PR #83, opened 2018-03-06
Allow logging in with the account email address

hughsie:
Fixes some of https://github.com/hughsie/lvfs-website/issues/82

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -214,9 +214,16 @@ def new_account():
 def login():
     """ A login screen to allow access to the LVFS main page """
     # auth check
+    used_deprecated_username = False
     user = db.session.query(User).\
-            filter(User.username == request.form['username']).\
+            filter(User.email == request.form['username']).\
             filter(User.password == _password_hash(request.form['password'])).first()
+    if not user:
+        # fallback, but not forever
+        user = db.session.query(User).\
+                filter(User.username == request.form['username']).\
+                filter(User.password == _password_hash(request.form['password'])).first()
+        used_deprecated_username = True
     if not user:
         flash('Failed to log in: Incorrect username or password for %s' % request.form['username'], 'danger')
         return redirect(url_for('.index'))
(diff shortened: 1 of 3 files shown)


app/views.py
```diff
@@ -214,9 +214,16 @@ def new_account():
 def login():
     """ A login screen to allow access to the LVFS main page """
     # auth check
+    used_deprecated_username = False
     user = db.session.query(User).\
-            filter(User.username == request.form['username']).\
+            filter(User.email == request.form['username']).\
             filter(User.password == _password_hash(request.form['password'])).first()
+    if not user:
```
superm1: So what's going to happen if multiple accounts match that email then?  You might need to define that behavior and keep it in test case. Or possibly do a migration in the DB for anything with common email addresses to append a email+user@domain.com for now.

superm1:
The other part is that currently users are able to change their email address in the panel, which effectively changes their login.  Should that stay?  I'm guessing yes - it should only be restricted from changing if you logged in with OAuth.

app/views.py
hughsie: I was going to just do the `email+user@domain.com` thing manually, and then perhaps make the column unique.

hughsie:
>The other part is that currently users are able to change their email address in the panel, which effectively changes their login. Should that stay? I'm guessing yes

I guessed no -- we'd only let the user change the displayname -- the email address would be the 100% thing that tied the user to the account. If a user changed employment from Dell to Logitech you don't want to "update" the profile on the LVFS, you want a whole new user in the new vendor group.

> - it should only be restricted from changing if you logged in with OAuth.

I think the profile should be "locked" i.e. you can't change your display name either, as this will be coming from OAuth too.

hughsie:
This is too intertwined with the OAuth stuff, I'll close this, merge the branches and open a new PR with it all working. There's no point allowing logins with the email if we don't do OAuth and there's a whole lot of other stuff that makes sense to clear up at the same time.

PR #85, opened 2018-03-12
Allow vendors to hide the direct download link in the search results

hughsie:
For arguable reasons, some vendors want to hide the direct download links for
firmware on the LVFS pages. This means the firmware files are only available to
download by programs that parse the XML metadata, for instance gnome-software
and fwupdmgr.

To migrate, do:

    ALTER TABLE firmware ADD inhibit_download TINYINT DEFAULT 0;

To use this in firmware files, include a section in the metainfo.xml as follows:

    <custom>
      <value key=LVFS::InhibitDownload/>
    </custom>

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -505,6 +505,7 @@ class Firmware(db.Base):
     target = Column(String(255), nullable=False)
     checksum_signed = Column(String(40), nullable=False)
     user_id = Column(Integer, ForeignKey('users.user_id'), nullable=False)
+    inhibit_download = Column(Boolean, default=False)
 
     # include all Component objects
     mds = relationship("Component", back_populates="fw", lazy='joined')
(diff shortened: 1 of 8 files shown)


GavinChengGao:
By adding the new section in XML file, this gives vendor more choice to decide the source that user can download the LVFS package based on their policy. I'm glad to see this enhancement happen.

superm1:
I think this might be confusing to users. Rather than just hide the download link you should put a message along the lines of OEM disabling downloads from the web browser and users needing to use a client to do the install.

Also the documentation for metadata should be updated for this.

hughsie:
@superm1 all valid comments; I think all fixed in this latest commit.

hughsie:
@GavinChengGao the wording is okay for you? I can attach screenshots if you need.

PR #86, opened 2018-03-15
Wip/hughsie/flask migrate

hughsie:
This makes it easier to deploy without SSH access...


superm1:
Nothing stands out to me with this, but admittedly I don't know most of it :)

hughsie:
I'd like to run this on the testing server before it merges with master, but I'l wait for the OAuth stuff to be confirmed to avoid complicating things.

superm1:
The OAuth stuff is really close.  I think the only remaining problems are cosmetic now.

hughsie:
Okay, running a modified version of this cleaned up the database considerably. Merged, thanks.

PR #91, opened 2018-03-20
Wip/hughsie/target to remote ids

hughsie:
A slightly scary change, but this means we generate, sign and upload metadata in a cron job, rather than in the uwsgi process. This also lets us change the permissions of the GPG private key to be unreadable by uwsgi (and thus nginx), and also allows us to use signing methods like Sigul that take ~30 seconds to complete. If this looks broadly acceptable I'll push it to the testing server to test a bit more in anger.


hughsie:
I've pushed this to the testing instance, that is https://7bcca6c4-e943-426b-8039-4c024186ee68.pub.cloud.scaleway.com/ -- and I'm amazed it all works just like it should. Feedback very welcome about some of the new UI cleanups.

cron.py
```diff
@@ -0,0 +1,170 @@
...
+    for r in db.session.query(Remote).all():
+        if r.name == 'private':
+            continue
+        if r.is_dirty:
+            remotes.append(r)
+            remote_names.append(r.name)
+
+    # nothing to do
+    if not len(remote_names):
+        print('No dirty remotes')
```
superm1: Since you'll be running this in cron typically, shouldn't you try to avoid printing extra superfolous like this without a debug flag?  Or are you just redirecting everything to /dev/null or a rotating log etc?

cron.py
```diff
@@ -0,0 +1,170 @@
...
+if __name__ == '__main__':
+
+    if len(sys.argv) != 2 or sys.argv[1] not in ['metadata', 'firmware']:
+        print('Usage: %s [metadata|firmware]' % sys.argv[0])
+        sys.exit(1)
+
+    # regenerate and sign metadata
+    if sys.argv[1] == 'metadata':
+        try:
+            rc = _regenerate_and_sign_metadata()
```
superm1: I'm a little confused since you jump back and forth with return codes and exceptions, but since you catch the exception and the only possible return code is 0 why bother with a return code at all?

cron.py
```diff
@@ -0,0 +1,170 @@
...
+        except NotImplementedError as e:
+            print(str(e))
+            rc = 1
+    elif sys.argv[1] == 'firmware':
+        try:
+            rc = _regenerate_and_sign_firmware()
```
superm1: same as above comment

app/views_metadata.py
```diff
@@ -68,10 +70,10 @@ def metadata_rebuild():
...
+    for r in db.session.query(Remote).filter(Remote.is_public).all():
+        r.is_dirty = True
     for vendor in db.session.query(Vendor).\
                     filter(Vendor.is_account_holder != 'no').all():
-        _metadata_update_group(vendor.group_id)
-    _metadata_update_targets(['stable', 'testing'])
-    _metadata_update_pulp()
-    flash('Metadata rebuilt successfully', 'info')
+        vendor.remote.is_dirty = True
+    flash('Metadata will be rebuilt soon', 'info')
```
superm1: If you're going to be flashing a message, I think it would make sense to flash exactly when is "soon".

hughsie: As in a time, or something like "in 2 minutes, 3 seconds"? If the former, we need to know the users timezone.

cron.py
hughsie: We return 1 if there was an exception, no?

app/views_metadata.py
superm1: I think saying in "2 minutes, 3 seconds" is fine.

cron.py
superm1: I mean specifically from `_regenerate_and_sign_metadata()` no need to return any code whatsoever.  All it will ever return is 0.  If you hit an exception, then it wouldn't have returned anything in the first place.

app/templates/firmware-details.html
```diff
@@ -1,176 +1,102 @@
...
   </tr>
   <tr class="row">
-    <th class="col-sm-4 align-middle">Current Target</th>
-    <td class="col-sm-6 align-middle">
-      {{fw.target}}
+    <th class="col-sm-3 align-middle">Current Target</th>
+    <td class="col-sm-7 align-middle">
+{% if fw.remote.name != 'private' and fw.remote.is_dirty %}
+      <div class="alert alert-warning" role="alert">
+        The <code>{{fw.remote.name}}</code> metadata is waiting to be signed.
```
superm1: It would also be good to mention here when this will happen.

superm1:
Some feedback about the U/I:
* The breadcrumb at the bottom just shows "Firmware #XYZ".  I think it would be better to show the title with parentheses around #XYZ.  Other than internally to LVFS no one will know what firmware #'s mean.
* I think it would make more sense for the history tab to show updates in descending order like a lot of other LVFS pages do.
* Under recent downloads what is address supposed to mean?  That doesn't look like any address I know.
* It's not clear why the `embargo` target is renamed in the U/I to `embargo-oem`.  Is there a particular reason for this?
* `Components » Details » Update Details and setting an urgency from the dropdown.` How about hyperlinking this too?

cron.py
hughsie: Excellent point, fixed.

cron.py
hughsie: Fixed.

app/views_metadata.py
hughsie: Fixed, and I've also done the same change on the metadata page.

hughsie:
> I think it would be better to show the title with parentheses around #XYZ

Agreed, I'll have a think on how to do this properly.

> I think it would make more sense for the history tab to show updates in descending order like a lot of other LVFS pages do.

Fixed.

> Under recent downloads what is address supposed to mean? That doesn't look like any address I know.

It's a hash of the IP address; the only reason it's there is so you can tell if it's one "client" (e.g. the CI server) or if it's different. Maybe it's not hugely useful.

> It's not clear why the embargo target is renamed in the U/I to embargo-oem. Is there a particular reason for this?

I wanted to make it clear that the embargo remote was only available for that one vendor, rather than "shared" like the testing and stable remotes are. Would `oem-embargo` have been clearer?

> How about hyperlinking this too?

Sorry, a hyperlink from where?

superm1:
I think the hash is only really useful in the event you show all records. Since it's just the last 30 or so it's not too useful to me. 

Hyperlink from the error message about needing to set description over to the page that does it in addition to telling them how to get there.

hughsie:
> Hyperlink from the error message about needing to set description over to the page that does it in addition to telling them how to get there.

The problem with this is that you get one "problem" per firmware, and a firmware can have multiple components. In the case where there are 5 different metainfo files in one .cab file you'd have to have five different links. I suppose we could make problems per-component rather than per-firmware but it complicates things a little.

superm1:
Ah okay then probably not worth the trouble

hughsie:
> Agreed, I'll have a think on how to do this properly.

Fixed.

superm1:
> I think the hash is only really useful in the event you show all records. Since it's just the last 30 or so it's not too useful to me.

Regarding this point, what do you think about having a button or link that gives "Raw" data (Of course keep the IP hashed)?  It might be useful to provide to a data scientist for further processing and statistical analysis.  The 30 records doesn't really show much, and even hiding it behind a page2/page3/etc button would be tedious to fetch all the results.

hughsie:
> what do you think about having a button or link that gives "Raw" data 

I've deliberately avoided this for OEM and ODMs so far; I'm acutely aware of the privacy implications people are worried about (although I'm not that bothered myself). If there's any specific stat you're interested in I'm sure we can work something out, but I think the unfiltered anonymized data is perhaps a step too far.

superm1:
> I've deliberately avoided this for OEM and ODMs so far; I'm acutely aware of the privacy implications people are worried about (although I'm not that bothered myself). If there's any specific stat you're interested in I'm sure we can work something out, but I think the unfiltered anonymized data is perhaps a step too far.

I was hoping hashing IP addresses would be enough for the privacy implications.  The things *I* think that are most interesting are:

- pie chart showing the OS break down where known by user agent data (unknown for generic or something)
- pie chart showing fwupd version breakdown
- histogram showing number of downloads made by a single client

Those same pie charts and histrograms I think are useful both on an "OEM" account wide basis and per-firmware basis.  I'm no statistician though, i'm sure there are other things that are useful and would be analyzed across the bigger data set.

hughsie:
All three seem pretty sane; can you file a new issue and we'll track those three stats items there.

superm1:
Sure thing, opened up another issue on it.

Can you let me know when you've refreshed the staging instance and i'll take another look for this PR?

hughsie:
Staging instance updated.

PR #101, opened 2018-04-04
Only show the legal agreement once rather than on every upload

hughsie:
Also, add the ability to version the agreements and only prompt users of the
LVFS to re-accept the terms when the version number changes.

We need this for the migration to the Linux Foundation, where users will have
to agree that a different legal entity is going to be distributing thier
firmware. The LF legal team also want to polish up the current agreement.

### DIFF
--- a/app/templates/agreement-admin.html
+++ b/app/templates/agreement-admin.html
@@ -1,0 +1,20 @@
+{% extends "default.html" %}
+{% block title %}Legal Agreement{% endblock %}
+
+{% block content %}
+
+<h2>Legal Agreement</h2>
+<form method="post" action="/lvfs/agreement/{{agreement.agreement_id}}/modify">
+  <div class="form-group">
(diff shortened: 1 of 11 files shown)


app/templates/agreement-admin.html
```diff
@@ -0,0 +1,20 @@
...
+{% block content %}
+
+<h2>Legal Agreement</h2>
+<form method="post" action="/lvfs/agreement/{{agreement.agreement_id}}/modify">
+  <div class="form-group">
+    <label for="display_name">Version (incrementing will prompt users):</label>
+    <input type="number" class="form-control" name="version" value="{{agreement.version}}" required>
+  </div>
+  <div class="form-group">
+    <label for="password">Markup in HTML format:</label>
```
superm1: Label for password seems wrong doesn't it?

hughsie: good catch!

PR #102, opened 2018-04-05
Add two static files that can be used in fwupd

hughsie:
I'd love some ideas on the prose here; it'll take some edits before it's suitable. When you're happy here I'll run it past PM my side.

### DIFF
--- a/app/static/org.fwupd.lvfs-stable.metainfo.xml
+++ b/app/static/org.fwupd.lvfs-stable.metainfo.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 Richard Hughes <richard@hughsie.com> -->
+
+<component type="source">
+  <id>org.fwupd.lvfs-stable</id>
+  <name>Linux Vendor Firmware Service (stable firmware)</name>
+  <metadata_license>CC0</metadata_license>
+  <agreement type="warning" version="1.0">
+    <agreement_section>
+      <description>
+        <!-- TRANSLATORS: do not translate the variables marked using $ -->
+        <p>
+          The LVFS is a free service that operates as an independent legal
+          entity and has no connection with $OS_RELEASE:NAME$.
+          Your distributor may not have verified any of the firmware updates for
+          compatibility with your system or connected devices.
+          All firmware is provided only by the original equipment manufacturer.
+        </p>
+        <p>
+          Enabling this functionality is done at your own risk, which means you
+          have to contact your original equipment manufacturer regarding any
+          problems caused by these updates.
+          Only problems with the update process itself should be filed at
+          $OS_RELEASE:BUG_REPORT_URL$.
+        </p>
+      </description>
+    </agreement_section>
+  </agreement>
+</component>
--- a/app/static/org.fwupd.lvfs-testing.metainfo.xml
+++ b/app/static/org.fwupd.lvfs-testing.metainfo.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 Richard Hughes <richard@hughsie.com> -->
+
+<component type="source">
+  <id>org.fwupd.lvfs-stable</id>
+  <name>Linux Vendor Firmware Service (testing firmware)</name>
+  <metadata_license>CC0</metadata_license>
+  <agreement type="warning" version="1.0">
+    <agreement_section>
+      <description>
+        <!-- TRANSLATORS: do not translate the variables marked using $ -->
+        <p>
+          The LVFS is a free service that operates as an independent legal
+          entity and has no connection with $OS_RELEASE:NAME$.
+          Your distributor may not have verified any of the firmware updates for
+          compatibility with your system or connected devices.
+          All firmware is provided only by the original equipment manufacturer.
+        </p>
+        <p>
+          This remote contains firmware which is not embargoed, but is still being
+          tested by the hardware vendor.
+          You should ensure you have a way to manually downgrade the firmware if
+          the firmware update fails.
+        </p>
+        <p>
+          Enabling this functionality is done at your own risk, which means you
+          have to contact your original equipment manufacturer regarding any
+          problems caused by these updates.
+          Only problems with the update process itself should be filed at
+          $OS_RELEASE:BUG_REPORT_URL$.
+        </p>
+      </description>
+    </agreement_section>
+  </agreement>
+</component>


app/static/org.fwupd.lvfs-stable.metainfo.xml
```diff
@@ -0,0 +1,32 @@
...
+        <p>
+          The LVFS is a free service that operates as an independent legal
+          entity and has no connection with your distribution.
+          Your distributor may not have verified any of the firmware updates for
+          compatibility with your system.
+          All the content is instead provided by the original equipment manufacturer.
+        </p>
+        <p>
+          Enabling this functionality is done at your own risk, so please do not
+          file issues with your distributor as they will not be able to help.
```
superm1: However if there are issues with the flashing we want those reported.

app/static/org.fwupd.lvfs-stable.metainfo.xml
```diff
@@ -0,0 +1,32 @@
...
+        </p>
+        <!-- for the testing remote only -->
+        <p>
+          This remote contains firmware which is not embargoed, but is still being
+          tested by the OEM.
+          You should ensure you have a way to manually downgrade the firmware if
+          the firmware update fails.
```
superm1: Fine by me

app/static/org.fwupd.lvfs-stable.metainfo.xml
```diff
@@ -0,0 +1,25 @@
...
+<component type="source">
+  <id>org.fwupd.lvfs-stable</id>
+  <name>Linux Vendor Firmware Service (stable firmware)</name>
+  <metadata_license>CC0</metadata_license>
+  <agreement type="warning">
+    <agreement_section>
+      <description>
+        <p>
+          The LVFS is a free service that operates as an independent legal
+          entity and has no connection with your distribution.
```
superm1: So something to consider is that if this is going to be distributed _by_ fwupd, distribution can be switched out with `lsb_release` output and will look more integrated.

app/static/org.fwupd.lvfs-stable.metainfo.xml
```diff
@@ -0,0 +1,25 @@
...
+          Your distributor may not have verified any of the firmware updates for
+          compatibility with your system.
```
superm1: Since fwupd does devices too, "system or devices".

app/static/org.fwupd.lvfs-stable.metainfo.xml
superm1: Although even if it's distributed by LVFS, fwupd and gnome-software has to get updated to be showing it, so perhaps it can easily be changed when it's received the first time.

app/static/org.fwupd.lvfs-stable.metainfo.xml
```diff
@@ -0,0 +1,25 @@
...
+          All the content is instead provided by the original equipment manufacturer.
```
superm1: Similarly can determine the OEM of the machine from SMBIOS information, so maybe "by $SYSTEM_OEM and other device manufacturers"?

app/static/org.fwupd.lvfs-stable.metainfo.xml
hughsie: I hadn't thought about this. It's a good idea, and certainly makes things more explicit rather than talking about "your distribution".

app/static/org.fwupd.lvfs-stable.metainfo.xml
hughsie: Fixed.

app/static/org.fwupd.lvfs-stable.metainfo.xml
hughsie: Hmm, I think this is stretching this too far; it's not unlikely fwupd is going to be offering the Logitech security updates for an external mouse not purchased at the same time as the laptop/desktop and it would be weird saying the updates are from "Dell" in this case.

hughsie:
@ximion -- would this warning make the LVFS acceptable to Debian too?

ximion:
@hughsie Yes, if the user has to explicitly confirm that they want the additional data download, it should be perfectly fine with Debian. 
(It is not for the FSF though, and thereby PureOS - I'm not yet sure how to address that at all yet)

ximion:
@superm1 On Debian, having the system automatically download proprietary components and install them from repositories not controlled by Debian without asking for user confirmation first is highly likely very problematic. Usually, software that depends on non-free components lives in the "contrib" part of the archive for that reason. The case of fwupd is really special here though, because it itself is free software and it only downloads non-free components if it has repositories containing them (unlike other stuff in contrib that needs non-free code to build, or to use it at all). It also is useful without the non-free stuff.
I didn't want to initiate a debate about fwupd on debian-devel without talking to you first though, and tbh I wasn't looking forward to that debate too much anyway ^^ - it would likely be useful to avoid having it at all.
In any case, if the user says yes to adding the LVFS repository in some way, Debian is without a doubt fine with fwupd and LVFS being there by default.
So, that should be fine :-)

With PureOS it's different though - PureOS is a FSF-endorsed Linux distribution, which means that it has to adhere to the FSF guidelines on (non-)free software. Which means that not only should there be no reference to non-free software anywhere, but also even downloading information about it is forbidden. So, downloading metadata that references non-free firmware *at all* is not allowed, and we aren't even allowed to ask the user about it (we recently were ask to remove kernel log messages on missing firmware...).
So, to satisfy the FSF, we would need to split the LVFS into a "only free software" and a "non-free" repository, and only ever use the first one.
Apparently the FSF position (from discussion, no clear official statement exists that I know of) on "shouldn't we allow our users to update their insecure proprietary firmware to a more secure version?" is that if you have bought hardware needing non-free code you have already lost and not allowing the user to upgrade the firmware - even if it is to fix security issues - is actually a favor for them because it prevents the proprietary vendors to lock down devices even further via firmware updates after you bought the device.
So yeah, this is a mess.

hughsie:
> is that if you have bought hardware needing non-free code you have already lost

I think that flies in the face of  both pragmatism and 100 years of commercial intellectual property law, but it's an *interesting* stance. I'm not too worried about not being included in PureOS, I'm certainly not going to split up the LVFS into free and nonfree server-side unless someone is volunteering to do all the engineering and testing.

app/static/org.fwupd.lvfs-stable.metainfo.xml
```diff
@@ -0,0 +1,26 @@
...
+          The LVFS is a free service that operates as an independent legal
+          entity and has no connection with $OS_RELEASE:NAME$.
+          Your distributor may not have verified any of the firmware updates for
+          compatibility with your system or devices.
+          All the content is instead provided by the original equipment manufacturer.
+        </p>
+        <p>
+          Enabling this functionality is done at your own risk, so please do not
+          file issues at $OS_RELEASE:BUG_REPORT_URL$ as they will be ignored.
+        </p>
```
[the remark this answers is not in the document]
hughsie: I've tweaked your wording a bit and used that, thanks.

ximion:
@hughsie I can completely accept that. I want everything to be software, of course, but applying bad engineering practice for ideological reasons and denying users security or documentation to make the choice to use a certain feature (even if it's proprietary) is something I am not fine with personally.
Debian is no FSF-endorsed distribution because our Wiki contains information on how to obtain non-free software and we make it easy for our users to get that stuff if they need it and provide documentation for it...

Also, from the recent Linux Libre release:
http://lists.gnu.org/archive/html/info-gnu/2018-04/msg00002.html
> Another significant change in this release is that it was pointed out
that there were error messages in Linux suggesting users to update x86
CPU microcode.  Since such microcode is non-Free Software, such messages
don't belong in GNU Linux-libre.  We now have patterns to detect and
clean up this sort of message.  A number of them were introduced
recently, relying on microcode changes to mitigate Spectre and Meltdown
problems, but there might be others that go farther back.  I haven't yet
made my mind on whether to go back, check and possibly respin such
earlier releases.

While the impact of this is likely minimal due to BIOS updates containing the change and other mitigations already being in place, it is IMHO unacceptable to hide this issue from the user and make software less transparent and potentially less secure just because it mentions non-free stuff.

Anyway, enough OT rant for today ^^ - Chances are that I might need to deal with the FSF/PureOS issue later in some way.
For now, having some kind of notification or even prompt for the user in the form of "hey, we download firmware for you in the background! do you want that?" would presumably resolve 90% of all concerns distros or people would have with the service.

superm1:
@ximion Thanks for that background on all this.  Glad you think Debian should be OK here.  I agree with @hughsie I don't think it's worth wasting too many brain cycles on sorting out philosophical differences with PureOS.

I suppose logistically we'll need to sort out a few things regarding this turning on remotes after these agreements are in place in LVFS and the associated code in fwupd and gnome-software.  Here's my thoughts:

**Upstream fwupd/gnome-software:**
1. By default upstream fwupd for the 1.0.7 release will probably need to set the LVFS stable remote to off.
2. Running the `refresh` fwupdmgr command in this state should probably prompt to enable the stable remote, or at least guide the user to the `modify-remote` command.
4. G-S: Rather than bury the option to enable LVFS stable somewhere you need to know to look, I think it would be worthwhile to have gnome-software check some heuristic and prompt once the first time it's opened, or during gnome-initial-setup or something to turn on the remote the first time.  Maybe look for any devices supported by 'uefi, thunderbolt, or unifying' and if so prompt to enable?  As more devices are added this heuristic can change in the future to reflect what's on LVFS.  Interested in thoughts here.
3. The packaging for deb and rpm should detect this fwupd upgrade scenario (vs clean install) from something less than 1.0.7 and automatically turn the remote back on during upgrade if it was on before.

If that sounds good to you guys i'll open a bug to track those in fwupd.

**For Debian (Although these probably also apply to what we'll do in some other OSes too for consistency):**
1. After a new enough gnome-software for the agreement support and new enough fwupd are in place, I think you can add an explicit dependency for fwupd to gnome-software to pull it in by default.
2. We probably need to add `fwupdate` as recommends from something, or configure it to install from gnome-software the first time someone performs a UEFI update or has one available or something.  Not sure what yet makes sense.  Interested in your thoughts here too..

hughsie:
> By default upstream fwupd for the 1.0.7 release will probably need to set the LVFS stable remote to off.

Why? Upstream it can stay on, and Debian can patch it downstream.

superm1:
> Why? Upstream it can stay on, and Debian can patch it downstream.

Consistency of experience I would think across distros, but maybe it doesn't matter.

hughsie:
> Consistency of experience I would think across distros, but maybe it doesn't matter.

It HUGELY matters; studies have shown the colossal importance of choosing the right default. I guarantee 99.99% of people downloading Fedora 28 in a few weeks time won't know what *the LVFS* is, and I'd guess 90% of those people would never even touch the console or any kind of GUI settings. **Those** are the users we service with the choice of upstream default (keeping them safe) and the more advanced/limited distros can either chose a different default or let users choose their own adventure.

Also, having something on by default tells the world what you think: *this is good enough for primetime*. Nothing is off-by-default in the real world that doesn't degrade performance in some way.

superm1:
OK sounds fine to me that we don't flip the default upstream then.  I do think upstream we should still try to get people onto that path of enabling FW updates though anywhere that we can if it's off by a distro decision by default but someone tries to use a client.  I've added this PR (to pair with your agreement one) which should help in that area: https://github.com/hughsie/fwupd/pull/459

hughsie:
Okay, I'll merge this when the AppStream `<agreement>` XML schema changes are agreed. Thanks.

app/static/org.fwupd.lvfs-stable.metainfo.xml
```diff
@@ -0,0 +1,29 @@
...
+<!-- Copyright 2018 Richard Hughes <richard@hughsie.com> -->
+
+<component type="source">
+  <id>org.fwupd.lvfs-stable</id>
+  <name>Linux Vendor Firmware Service (stable firmware)</name>
+  <metadata_license>CC0</metadata_license>
+  <agreement type="warning" version="1.0">
+    <agreement_section>
+      <description>
+        <!-- TRANSLATORS: do not translate the variables marked using $ -->
```
superm1: I just realized you have a translation comment.  Should these be coming from LVFS in the form of static files or from fwupd?

hughsie: Hosted *from* the LVFS, but we also have to work out how to translate that file. Either we wire up lvfs-website to transifex, or we just copy the file to fwupd and get it translated there "for free".

superm1: Ah OK.

hughsie: Note: I'm not sure which is the *best* way; ideas welcome.

superm1: Well the thing about legal agreement type output that needs to be blessed by lawyers is that usually they also require professional translations (at least at my company that's the case).  I assume you're already running it by those folks, and should confirm if they require professional translations or not.

hughsie: We need the translations to be verified (and in some cases fixed) but in 99% of cases the upstream translation is fine.

hughsie:
I'm having second thoughts about putting these as separate files; would we also have files for the embargo remotes? Do we have *yet another*metainfo file for the global-remote privacy policy?

What about a better proposal where we have one appstream file, with the remotes as sub components of the master document? That also means we share one download with lvfs and lvfs-testing, and means we can call the fwupd API bits AppStreamURI rather than the inelegant MetainfoURI which is way too close to MetadataURI for my taste.

hughsie:
And third thoughts; do we actually need the remote enable warning to be downloaded over HTTP? Could we just ship the metainfo files in fwupd and install them there? If we needed a different message we could just patch it there.

superm1:
I know you already started to implement <3rd thoughts> into fwupd, but I think that of the three approaches I think the biggest limiting factor will be how often you expect this agreement to change.  It's more difficult to push updated agreements into the field across all the distros through new fwupd versions, but you can do that pretty dynamically with LVFS.

If you keep it in located at LVFS, i've got a 4th idea.  What if you just merge it into the existing metadata that's already built?  So if you choose to "enable" a source then your action is to download the metadata for the first time and display that agreement.  If they agreed you already have the metadata in place and can start using it.

It's a few extra kilobytes in everyone's daily download, but as a bonus this lets you later possibly also revision the agreement easily from the web side and prevent updates from being run until a new agreement is agreed to. (If that's a potential)

hughsie:
> but you can do that pretty dynamically with LVFS.

Right.

> What if you just merge it into the existing metadata that's already built?

So, i actually considered that. The problem there is that you have to refresh the metadata *before* you enable the remote in that case, which doesn't play super well with the GUIs that expect to be able to download metadata on idle.

I suppose we could show the warning in fwupd, and then show a different warning when the user has refreshed but that kinda sucks. I'll loop back to legal and see what they say; I have a feeling that the current warning is sufficient and unlikely to need lots of different revisions.

hughsie:
Lets close this, and we can reopen with a better PR (i.e. putting it in the real metadata) if required.

PR #104, opened 2018-04-09
Allow undeleting firmware

hughsie:
This means instead of deleting the firmware file we switch to renaming it with
a .bak suffix. This is not included in the metadata or in the search results.

This allows us to preserve the database history for audit reasons, and also to
implement an 'undelete' action.


app/views_firmware.py
```diff
@@ -119,30 +151,24 @@ def firmware_delete(firmware_id):
...
 
+    # find private remote
+    remote = db.session.query(Remote).filter(Remote.name == 'deleted').first()
+    if not remote:
+        return _error_internal('No deleted remote')
+
+    # rename file so it's no longer downloadable
+    path = os.path.join(app.config['DOWNLOAD_DIR'], fw.filename)
+    if os.path.exists(path):
+        os.rename(path, path + '.bak')
```
superm1: How does renaming to .bak guarantee it's not downloadable?  I would think it should need to be moved to somewhere that's not exposed to the web endpoint too.

superm1:
So one concern I have with this approach is what happens if a file is uploaded, a mistake is discovered that didn't require updating the firmware version but just say a no-change rebuild or metadata update so it's re-uploaded. I would think that this could no longer work would it?   You would need to be able to account for this specific case.

app/views_firmware.py
```diff
@@ -103,6 +105,36 @@ def firmware_modify(firmware_id):
...
+        return _error_internal("No firmware file with ID %s exists" % firmware_id)
+    if not g.user.check_for_firmware(fw):
+        return _error_permission_denied('Insufficient permissions to undelete firmware')
+
+    # find private remote
+    remote = db.session.query(Remote).filter(Remote.name == 'private').first()
+    if not remote:
+        return _error_internal('No private remote')
+
+    # rename file so it's no longer downloadable
```
superm1: Comment should be updated

lvfs_test.py
```diff
@@ -211,7 +211,7 @@ def test_upload_valid(self):
 
         # download missing file
         rv = self.app.get('/downloads/7514fc4b0e1a306337de78c58f10e9e68f791de2-hughski-colorhug2-2.0.3.cab')
-        assert rv.status_code == 404, rv.status_code
+        assert rv.status_code == 410, rv.status_code
```
superm1: It should be good to add some test cases around the flows for delete->undelete as well as what you expect to happen with redoing the upload if the file is currently in the deleted state.

app/views_firmware.py
hughsie: Well, it means it's not *readily* downloadable, but sentiment agreed. We can move this to a `trash` folder or something that nginx won't serve.

lvfs_test.py
hughsie: > what you expect to happen with redoing the upload if the file is currently in the deleted state

I assumed we'd just show the user the existing deleted file and give them a chance to undelete it. Is that wrong?

app/views_firmware.py
hughsie: Fixed.

hughsie:
> what happens if a file is uploaded, a mistake is discovered that didn't require updating the firmware version but just say a no-change rebuild or metadata update so it's re-uploaded

It'll have a different SHA1 hash, so a different firmware_id. The re-upload will work, but we'll have to be a bit more careful on the "dropping GUID" checks for instance.

lvfs_test.py
superm1: As long as new uploads of the "same" file show up as a different firmware_id that's fine.

superm1:
> It'll have a different SHA1 hash, so a different firmware_id. The re-upload will work, but we'll have to be a bit more careful on the "dropping GUID" checks for instance.

I think that's specifically what should be tested in the test suite then since it can be fragile later.

hughsie:
I've been testing this for a while on the staging instance. Fingers crossed.

Issue #87, opened 2018-03-15
Add sigul support


jordane:
We need to add sigul support. Per discussion with @hughsie, this should be done before lvfs.linuxfoundation.org or similar is launched.

hughsie:
Signing the metadata using sigul should be pretty straightforward; we just give it a `metadata.xml.gz` and it returns us a `metadata.xml.gz.asc` -- and then when we write metadata we'd just put it in a special place for a cron job to pick up, send to sigul and then move the detached signature into `downloads`.

The bit I'm not sure how to do is the cabinet archive. The signature of that is actually the detached signature of the firmware blob (e.g. `firmware.ROM`) added to the archive as `firmware.ROM.asc`. Because of the slightly unusual way Microsoft Windows Update builds the archive we really need to use GCab (which we do in the LVFS using GObjectIntrospection) rather than anything in pure python. I see two solutions here:

* We teach sigul about GCab, GObjectIntrospection and how to sign a firmware archive
* We do some kind of wrapper so that we ask sigul to sign `firmware.ROM` and then the LVFS/a cron job builds the destination .cab file and copies it into `downloads/`

I'm erring on solution 2 for simplicity.

@jordane so I can prototype something; a few questions:

* When you run the sigul command line on the client (e.g. to sign a metadata file) does it block until the destination file has been written?
* If there's an error while signing, does sigul return with a non-zero exit code and dump something on stderr?
* Do we know how long sigul is going to take? I know we guessed 5 minutes, but an actual number would be super from a UX point of view.
* In sigul, when you specify the `KEY` is it the `UID` (lfvs+firmware@linuxfoundation.org) it wants or the `KEY_ID` e.g. 183A73F1?
* Do you need to be a specific user on the client to use sigul? e.g. do we do it from the uwsgi user as a LVFS plugin in the flask process, or do we do it as a cron job as another user for example?

jordane:
> I'm erring on solution 2 for simplicity.

I agree.

> When you run the sigul command line on the client (e.g. to sign a metadata file) does it block until the destination file has been written?

Yes, but see the last answer.

> If there's an error while signing, does sigul return with a non-zero exit code and dump something on stderr?

Yes, but see the last answer.

> Do we know how long sigul is going to take? I know we guessed 5 minutes, but an actual number would be super from a UX point of view.

It's basically instantaneous as soon as the data is transferred to the sigul server. Easily less than 30s unless the firmware files are large. 

[...]

PR #105, opened 2018-04-09
Add support for signing firmware using Sigul

hughsie:
Fixes https://github.com/hughsie/lvfs-website/issues/87

### DIFF
--- a/plugins/sign-sigul/__init__.py
+++ b/plugins/sign-sigul/__init__.py
@@ -1,0 +1,115 @@
+#!/usr/bin/python2
+# -*- coding: utf-8 -*-
+#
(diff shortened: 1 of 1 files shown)


superm1:
No test suite, i'm guessing since it would add a new dependency on lvfs-website?

hughsie:
Well, we'd have to mock-up the sigul command line, so I didn't see much point. Other ideas welcome.

superm1:
Can you deploy a sigul instance with a test key?  Maybe not worth the effort though.  Otherwise looks fine by me.

hughsie:
> Can you deploy a sigul instance with a test key

Dude, it took me a day to deploy Sigul on my local system and have it half work. :)

Issue #100, opened 2018-04-02
Allow setting default upload target per-account


superm1:
At least for my company we don't use the "private" target at all.  Would you consider to let me configure company wide to set the default to "embargo" rather than "private"?

Would mean a few less clicks at least.  Obviously really low priority.

hughsie:
Set the default, or remove "private" from the drop down when logged in as Dell? If you don't want people to use it it seems better to just remove the upload location drop-down completely.

superm1:
Good point, I think removing it completely is better.

hughsie:
Is this per-vendor (i.e. for all the Dell ODMs) or per user (i.e. for a specific ODM)?

superm1:
per-vendor

PR #107, opened 2018-04-11
Allow uploading by default to embargo for some vendors

hughsie:
Fixes: https://github.com/hughsie/lvfs-website/issues/100

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -234,6 +234,7 @@ class Vendor(db.Model):
     description = Column(Unicode, default=None)
     visible = Column(Boolean, default=False)
     visible_for_search = Column(Boolean, default=False)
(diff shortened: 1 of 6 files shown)


superm1:
Looks fine to me.  can you push it to staging and i'll confirm it works there?

hughsie:
Done. I've set up dell to be embargo-by-default there too.

hughsie:
I wasn't sure about that; the new key is `is_embargo_default` rather than `disallow_using_private_target` -- do you want to prevent the use of private completely? I suppose there's now a delete/undelete action if you want to remove the firmware from the dell embargo remote for a short time.

superm1:
Yeah that was exactly my thought too, private really doesn't provide anything more than "delete" does.

PR #108, opened 2018-04-23
Wip/hughsie/firmware limits

hughsie:


hughsie:
screenshot from 2018-04-23 10-53-25

superm1:
I think it may be useful to pre-populate possible user agent prefixes that LVFS has seen in this interface.

hughsie:
> pre-populate possible user agent prefixes

Got any idea on how the UI would look? Or do you mean perhaps show a list of the most common user agents for that firmware in a table below?

superm1:
I think showing common ones in the table below would be fine. I just think if you aren't familiar with the client code you probably have no idea what user agents exist.

The other ideai had was to show radio buttons or check boxes with the common ones.


Related thought; what does the company looking for this expect to happen on the client side? Are they wanting to use this like a slow roll out? The client may need to recognize this 429 error code specifically and show a message about being rate limited and maybe internally store something in the  daemon about not trying again for 24h.

If they're looking for staged rollout another approach can be to automatically demote to "testing" after so many downloads and repromote to stable the next day. That would prevent a lot of clients getting the 429 in the first place. People who opted into testing would always be able to get it then too.

hughsie:
> I think showing common ones in the table below would be fine.

Okay, will add something.

> Are they wanting to use this like a slow roll out? 

Exactly that.

> The client may need to recognize this 429 error code specifically and show a message about being rate limited and maybe internally store something in the daemon about not trying again for 24h.

Yes, if you thought this was a good idea this way my next patch; basically teach fwupd about 429 and show something translated on the console. For GUI clients we'd just not show anything which is fine.

> another approach can be to automatically demote to "testing" after so many downloads and repromote to stable the next day.

I actually tried that first. It means running another cron job, and means you can't filter based on the user agent string.

superm1:
> I actually tried that first. It means running another cron job, and means you can't filter based on the user agent string.

I guess I'm not absolutely sure the value of matching the user agent string.  To not `429` people who manually download from a browser or wget?

I actually really like the idea of an automatic way to demote to testing.  As more people opt in to reporting their results you can start to have a threshold for an automatic demotion and notify the uploader there is a problem (> X downloads and Y% are failure reports that aren't triaged).  This approach has a handful of benefits:

- Implicitly slows the rollout specifically when problems happen
- It prevents potentially troublesome firmware from being installed
- Forces the reports to be investigated

hughsie:
> I'm not absolutely sure the value of matching the user agent string.

It means the vendor QA team can still download the signed cab from the LVFS (user agent = firefox) and means you can also restrict the firmware using automatic GUIs (e.g. people who don't know what SPI flash is) and still let people on the command line do it. It also means from a UX point of view if you really want the firmware right now you can do it interactively with the CLI and stopping it from GUI clients doesn't impact their UX at all. It's really a rate-limiting thing rather than a policy thing.

> I actually really like the idea of an automatic way to demote to testing.

Yes, but I think that is a different PR. Moving from stable to testing on event of a bad firmware is a one-time thing (re-upload new firmware, actually QA this time...) but resigning two popular remotes every day means that clients are re-downloading new metadata every day, rather than the once per week-ish they are at the moment. It also means we don't churn too much if we're using "dumb" mirror.

superm1:
> It's really a rate-limiting thing rather than a policy thing.

OK that's the key - if rate limiting is the goal, then yeah this approach makes sense.

> Yes, but I think that is a different PR. Moving from stable to testing on event of a bad firmware is a one-time thing (re-upload new firmware, actually QA this time...) but resigning two popular remotes every day means that clients are re-downloading new metadata every day, rather than the once per week-ish they are at the moment. It also means we don't churn too much if we're using "dumb" mirror.

OK

app/templates/firmware-limits.html
```diff
@@ -0,0 +1,62 @@
+{% extends "default.html" %}
+{% block title %}Firmware Limits{% endblock %}
```
superm1: Probably better to make this "Firmware download limits"

app/templates/firmware-limits.html
```diff
@@ -0,0 +1,62 @@
...
+{% else %}
+      <code>*</code>
+{% endif %}
+    </td>
+    <td class="col-sm-4">
+{% if fl.response %}
+      <code>{{fl.response}}</code>
+{% else %}
+      <code>429 Too Many Requests</code>
+{% endif %}
```
superm1: Not sure this needs to be customized per limiter.  I would think you should make the code and error consistent.

app/templates/firmware-limits.html
hughsie: I made it `Download Limits`

app/templates/firmware-limits.html
hughsie: The code is always 429 -- I guess it's misleading to include it there. I'll remove it.

app/views.py
```diff
@@ -86,6 +86,21 @@ def serveStaticResource(resource):
...
+                cnt = _execute_count_star(db.session.query(Client).\
+                            filter(Client.firmware_id == fw.firmware_id).\
+                            filter(Client.timestamp >= yesterday))
+                if cnt >= fl.value:
+                    response = fl.response
+                    if not response:
+                        response = 'Too Many Requests'
+                    return Response(response=response,
+                                    status=429,
+                                    mimetype='text/plain')
```
superm1: So a 429 response allows a [retry after](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429).  Since this is going to be reset at the day mark, it should be useful to give this to the client to allow it to know when to try again.

superm1:
Other than my last comment, LGTM.

app/views.py
hughsie: Good idea -- I've hardcoded the seconds in one day. I did consider working out how many seconds it was until the next day and using that, but if the front-end started using Retry-After they'd all hammer the server/CDN at exactly the same second. I think hard-coding is fine as an advisory thing.

PR #109, opened 2018-04-25
Add a username policy for vendor managers

hughsie:
This allows the site-admin to restrict vendor managers to specific domain globs,
for instance allowing dave@oem.com to only create accounts ending with @oem.com
and *not* test@hotmail.com


superm1:
Actually I'm a little concerned, this means that you couldn't make an ODM account unless they had the @domain.com matched for the vendor ID doesn't it?  Maybe instead it would be better to blacklist things like gmail.com, yahoo.com, hotmail.com?

hughsie:
> you couldn't make an ODM account unless they had the @domain.com matched for the vendor ID doesn't it

Nahh, the admin (me) can still do anything. This just allows the admin to specify the domain globs that vendor managers (who can create other users themselves, without getting me involved) are restricted to. It would for example restrict *you* from creating `someone@lenovo.com` as a local account, but you would be allowed to create `someone@dell.com`.

superm1:
Hmm, can you let there be multiple domains that vendor managers can make rather than just matching to the one?  For example if I want to cover 2 of the bigger ODMs, my company and contractors for my company I might have:
compal.com, dell.com, dellteam.com, wistron.com

hughsie:
> can you let there be multiple domains that vendor managers can make

Hmm, good idea. Will fix.

Issue #110, opened 2018-04-30
LVFS Computer hardware IDs doesn't behave as expected


superm1:
The current U/I for CHIDs allows you to apply multiple hardware IDs, but those HWIDs are treated as *AND* rather than *OR*.  The similar U/I in WU treats them as *AND* so this is a confusing experience to folks uploading.

At least one FW has been affected that two CHIDs were set intending to go to two machines but the end result was the FW wasn't offered to either machine since it's impossible to meet both CHID requirements.

So I'd like to see the UX changed to match the WU approach where multiple HWIDs default to "OR".  This means creating a regex for the requirements around HWIDs.

superm1:
> According to https://fwupd.org/lvfs/firmware/551 it's been downloaded 2271 times, so I'm not sure how that's happened.

Theory: running on older fwupd (or fwupd compiled against older appstream-glib) that wasn't enforcing HWID checks.

> which means also fixing fwupd to split that before matching. Sane?

I wish fwupd 1.0.7 hadn't "just dropped" in that situation so we could get a version with this type of fix out together.  Although it sounds like there might not really be a no-client changes solution from your above description.

hughsie:
> that wasn't enforcing HWID checks

Maybe anything compiled against the old version of appstream-glib?

> I wish fwupd 1.0.7 hadn't "just dropped"

There's always "just one more bug" -- and I can do a new release next week if required. Releases are cheap.

superm1:
> Maybe anything compiled against the old version of appstream-glib?

I think that looks like a believable root cause.  At least the "recent" downloads I see are gnome-software with no fwupd version reported which could mean Ubuntu 16.04's fwupd / gnome-software.

> There's always "just one more bug" -- and I can do a new release next week if required. Releases are cheap.

Well but client changes means that stuff like RHEL is broken and can't be fixed for a while :/

hughsie:
> can't be fixed for a while

Nothing official but 7.6 is coming up and I feel good things.

PR #111, opened 2018-05-01
Only generate one HWID requirement if multiple are used

hughsie:
Using Windows Update the CHIDs are treated as 'OR' when matching and we should
probably follow the same logic to avoid confusing people.

Requires the fwupd daemon fix obviously too before it will work on real
hardware with the new-style requirements metadata from the LVFS.

Fixes https://github.com/hughsie/lvfs-website/issues/110

### DIFF
--- a/app/metadata.py
+++ b/app/metadata.py
@@ -107,6 +107,8 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri=''):
 
             # add manual firmware or fwupd version requires
             for rq in md.requirements:
+                if rq.kind == 'hardware':
+                    continue
                 req = AppStreamGlib.Require.new()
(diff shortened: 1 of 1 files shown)


hughsie:
> I think you should special case the "1", HWID scenario

Isn't that the effect here? i.e. it's just used to *join* the strings, and if there's just one string then the joining char isn't going to be used.

superm1:
> Isn't that the effect here? i.e. it's just used to join the strings, and if there's just one string then the joining char isn't going to be used.

I guess that isn't how I _expected_ python to handle it but you're right.  Carry along :)

```
$ cat test.py
#!/usr/bin/python3

rq_hws = ['abc', '123']
print ('|'.join(rq_hws))
rq_hws = ['abc']
print ('|'.join(rq_hws))

$ ./test.py
abc|123
abc

```

PR #120, opened 2018-06-13
Wip/hughsie/version format

hughsie:


superm1:
> This can be used to provide hints to tools such as fwupdmgr.

I guess i'm curious what will fwupdmgr be able to do with this? I would think the downloads are treated the same from fwupdmgr's perspective.  

So all it could really do is show a message that fwupdgmr is the only way the OEM wants you to download this update.

hughsie:
> I guess i'm curious what will fwupdmgr be able to do with this?

This part of the patch was more of a "this is a good idea to export" rather than driven by any requirement. At the moment we have fwupd and gnome-software, but we might have other clients in the future that might need access to this metadata key.

app/views_upload.py
```diff
@@ -63,6 +63,8 @@ def _create_fw_from_uploaded_file(ufile):
     # this allows an OEM to hide the direct download link on the LVFS
     if 'LVFS::InhibitDownload' in ufile.metadata:
         fw.inhibit_download = True
+    if 'LVFS::VersionFormat' in ufile.metadata:
+        fw.version_format = ufile.metadata['LVFS::VersionFormat']
```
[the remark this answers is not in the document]
hughsie: Good idea.

hughsie:
> Shouldn't this also be used to determine how to render the version in LVFS U/I too?

Yes, and this highlights another problem with version_format (and inhibit_download) -- they're supposed to be per-*component* rather than per-*firmware*. I'll have a rethink and fix up both.

FireWolf74:
hughsie,
On that note is there a way that each firmware update have a better understand in what your updating????
I just looked in my gnome-software under hardware this was all I see.
screenshot from 2018-06-13 12-59-28
I don't see HDD, video graphic for AMD, or even updates to the trackpad or many other components  that I have on my system.   
 Is that not an issue?

hughsie:
@FireWolf74 Yu're describing a different problem. Can you file another bug, but for gnome-software (so in Git*Lab*, rather than Git*Hub*) and we can discuss there.

hughsie:
> Shouldn't this also be used to determine how to render the version in LVFS U/I too?

I've fixed this. I was hesitant to change the fw.display_version as this is the thing that comes from the .inf file to describe the firmware, rather than the component. Checking if the firmware version of the component is an integer actually works pretty well. I think that only two LVFS firmwares need to be fixed up so they are superficially correct (the two TPM ones) as the metadata will be using the actual version, not the display version. On that note, I assume the TPM update is supposed to be a quad, not a triplet? I'm happy to just change the database rather than have to re-upload firmwares with new metadata if this patchset now looks okay.

app/views_upload.py
```diff
@@ -125,6 +121,13 @@ def _create_fw_from_uploaded_file(ufile):
         md.checksum_contents = csum.get_value()
         md.filename_contents = csum.get_filename()
 
+        # this allows an OEM to hide the direct download link on the LVFS
+        metadata = component.get_metadata()
+        if 'LVFS::InhibitDownload' in metadata:
+            md.inhibit_download = True
+        if 'LVFS::VersionFormat' in metadata:
```
superm1: nit, should probably put a comment about why version format is customizable here.

app/models.py
```diff
@@ -481,6 +483,20 @@ def __init__(self):
...
+            if self.version_format == 'quad':
+                return '%i.%i.%i.%i' % ((v & 0xff000000) >> 24,
+                                        (v & 0x00ff0000) >> 16,
+                                        (v & 0x0000ff00) >> 8,
+                                        v & 0x000000ff)
+            return '%i.%i.%i' % ((v & 0xff000000) >> 24,
+                                 (v & 0x00ff0000) >> 16,
+                                 v & 0x0000ffff)
+        return self.version
+
```
superm1: Can you push this to staging?  I'd like to see what it does to all the existing firmware.  I'm a little worried that it's going to mess up a lot of stuff since the "default" is triplet but https://github.com/hughsie/fwupd/blob/674ed347d27f4fb1fb7e6a45ff0e0d95102bfdfc/plugins/dell/dell.quirk
exists

superm1:
> On that note, I assume the TPM update is supposed to be a quad, not a triplet? I'm happy to just change the database rather than have to re-upload firmwares with new metadata if this patchset now looks okay.

Yeah everything Dell does so far has been quad.  I almost wonder if since there is a quirk in place for Dell and UEFI plugins (linked above) it's better to also set the default across the LVFS account?

app/models.py
hughsie: Most of the Dell updates come with a .inf file, which is used over this for the display version you see on the LVFS. In the metadata they'll all still be the raw hex or integer versions.

superm1: Ah I see.  So anything without an INF will need to have this field set to work properly.  Can you still push it to staging?  There's a few non-public items other than TPM I know without INF I'd like to see what happens.

hughsie:
> Yeah everything Dell does so far has been quad. 

Ohh? Looking a random .inf file (firmware 544) I saw:

    DriverVer   = 03/27/2018,1.6.0

>  it's better to also set the default across the LVFS account?

I did ponder this, but the bigger vendors have inconsistent version formatting styles; e.g. Logitech has the leading zero thing too.

superm1:
> Ohh? Looking a random .inf file (firmware 544) I saw:
> DriverVer   = 03/27/2018,1.6.0

Well so this gets to something that has been annoying to look at in fwupdmgr and gnome-software but otherwise has not been important to fix.

The display version is `1.6.0` that you got from the INF , the ESRT version from both INF and XML is 0x10600.  So if you take it as quad then it is `0.1.6.0`, which is _closer_ to right. The **real firmware version** is `1.6.0` not `0.1.6.0`.

superm1:
I guess what I really want is "quad" but don't render the leading zero if there is nothing there.

hughsie:
> I guess what I really want is "quad" but don't render the leading zero if there is nothing there.

You can't do that. The *don't render the leading zero* bit changes how the vercmp works. Imagine you release a `0.1.6.0` which gets "fixed" to `1.6.0`. You then release version `1.0.0.0` which doesn't get fixed, and you expect the update to happen from `1.6.0` to `1.0.0.0`:

    $ rpmdev-vercmp 1.6.0 1.0.0.0
    1.6.0 > 1.0.0.0

superm1:
> You can't do that. The don't render the leading zero bit changes how the vercmp works. Imagine you release a 0.1.6.0 which gets "fixed" to 1.6.0. You then release version 1.0.0.0 which doesn't get fixed, and you expect the update to happen from 1.6.0 to 1.0.0.0:

Right that's why I'm saying just "render".  The version comparison should always use what's in the metadata exactly (which in your example should be the comparison between 0x00010600 and 0x01000000)

hughsie:
I'm not sure I'm super comfortable having *different* versions for the display and for the comparison; in the case of `0x01000000` -> `1.0.0.0` it's exactly the same thing (and thus reversible, and also testable).

superm1:
> I'm not sure I'm super comfortable having different versions for the display and for the comparison; in the case of 0x01000000 -> 1.0.0.0 it's exactly the same thing (and thus reversible, and also testable).

Well in a sense it's the same thing that's actually done by LVFS.  There is a "Display Version" that is rendered in LVFS using what was in the INF, but none of the tools use it, it's just to look pretty.

What about actually _populating_ that into metadata and letting all the tools (eg `fwupdtool get-devices` and `gnome-software`) render an extra field `DisplayVersion`?

hughsie:
> What about actually populating that into metadata

Per-component or per-firmware?

superm1:
> Per-component or per-firmware?

It should probably be per-component

app/models.py
hughsie: Pushed to staging with a data snapshot of the 10th.

hughsie:
`LVFS::DisplayVersion` seems like the obvious thing to add, but it doesn't work as it needs to be per-release rather than per-component for the merge into the composite AppStream document. We'd also still need `LVFS::VersionFormat` as this changes how we format the version for comparison. fwupd could have a `fwupd_release_get_version_display()` although then we also have to patch all the clients, and then I'm sure someone is going to ask why each release has *two version numbers*.

tl;dr: I'm not convinced avoiding the `0.` prefix is reason enough for a world of pain.

superm1:
I took a look at this specifically with staging and I think that it's going to be problematic still for lots of existing firmware.  As an example, with firmware /lvfs/firmware/356/ the components page shows v0.1.2304 which should either be 00.01.09.00 (with the leading zeros) or 01.09.00 if going off what's in in the INF.
The same problem exists on all the UEFI updates this way.

hughsie:
I've re-pushed the branch to staging; does that look better?

superm1:
Yes, that looks much better for all the firmware I looked at (TPM and UEFI).  This PR is fine by me now.

Issue #122, opened 2018-06-18
Composite firmware uploads are rejected unless all components change


superm1:
This is with `/lvfs/firmware/650`. I found that if only one of the components changed the upload is rejected saying that there already is a firmware that has the same version for one of the other components.

At least for this device it's quite unlikely that all the components will move at the same time, so I think that composite firmware CAB files need to cover this scenario.

superm1:
If you need the CAB file as it was built to reproduce this error (or event just the metainfo files from it) happy to send it.

hughsie:
It would be useful to have the .cab pls.

superm1:
Sent it over email.

hughsie:
^this fixes the issue for me, but it does depend on https://github.com/hughsie/lvfs-website/pull/120 as we have to show the versions for the components to make any kind of sense.

PR #124, opened 2018-06-19
Allow uploading duplicate components in composite firmwares

hughsie:
Fixes https://github.com/hughsie/lvfs-website/issues/122

I've pushed this to staging, can you test please?

### DIFF
--- a/app/views_upload.py
+++ b/app/views_upload.py
@@ -134,6 +134,16 @@ def _create_fw_from_uploaded_file(ufile):
 
     return fw
 
+def _filter_fw_by_guid_version(fws, provides_value, release_version):
+    for fw in fws:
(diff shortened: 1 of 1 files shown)


superm1:
Yes confirmed it works on staging for me.  Looks good.

superm1:
Some more checks and I noticed the "delete" button isn't working.
`The LVFS is temporarily unavailable and should be back soon.`

hughsie:
> I noticed the "delete" button isn't working.

Incorrect permissions on a directory just on the staging instance. Should be fixed now.

hughsie:
Do you need this on the main instance ASAP or can it wait until Sunday?

superm1:
It can wait.  This is just very early testing.

PR #127, opened 2018-06-29
Allow any administrator to move users to different vendors

hughsie:


superm1:
Is this use case for employees changing companies it companies being renamed?

If the former I would argue firmware shouldn't follow them.

hughsie:
In this case it's a company with three divisions wanting to split up on the LVFS. I agree in some cases you wouldn't want the reparenting, eg if an odm is no longer being used by an OEM. I'll add a checkbox or something.

hughsie:
Okay, I've split this out to a dialog as it's going to be rarely used and changing it accidentally would be a Bad Thing (tm). I've added a checkbox for reparenting too.

hughsie:
Closing in preference to https://github.com/hughsie/lvfs-website/pull/129

PR #128, opened 2018-06-29
Deprecate old usernames harder

hughsie:
Provide the user with the information they need, but don't allow them to log in
with the old username.

In September we'll be be removing the old usernames completely as the GDPR
and product security make it important to map accounts to specific people.

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -271,18 +271,19 @@ def _create_user_for_oauth_username(username):
 def login():
     """ A login screen to allow access to the LVFS main page """
     # auth check
-    used_deprecated_username = False
     user = db.session.query(User).\
             filter(User.username == request.form['username']).\
             filter(User.password == _password_hash(request.form['password'])).first()
     if not user:
-        # fallback, but not forever
-        user = db.session.query(User).\
-                filter(User.username_old == request.form['username']).\
-                filter(User.password == _password_hash(request.form['password'])).first()
-        used_deprecated_username = True
-    if not user:
-        # user is NOT added to the database
+        # check to see if the user used the fallback name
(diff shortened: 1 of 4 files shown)


hughsie:
Also, a few vendors will need to change the email addresses of `lvfs_odm@oem.com` to point at actual people with real email addresses. I can make people managers (so they can add other **local** accounts, not just using Azure) if required to ease this transition.

superm1:
I think this PR is fine as it is, but making the email addresses real addresses will take some time as we make more accounts.

app/views.py
```diff
@@ -259,16 +259,17 @@ def _create_user_for_oauth_username(username):
...
-                filter(User.password == _password_hash(request.form['password'])).first()
-        used_deprecated_username = True
+        # check to see if the user used the fallback name
+        user_old = db.session.query(User).\
+                        filter(User.username_old == request.form['username']).\
+                        filter(User.password == _password_hash(request.form['password'])).first()
+        if user_old:
+            flash(u'You have to use %s as the username for this user' % user_old.username, 'danger')
+            return redirect(url_for('.index'))
     if not user:
```
superm1: Actually just noticed this clause can combine with the above `if not user` now

hughsie: Fixed.

hughsie:
I'm happy leaving things as they are until the 1st September if that makes your life easier; maybe we make the `wip/hughsie/deprecate-old-usernames-harder` branch the one we deploy on that day with the various fixes.

superm1:
That sounds good to me.  All the changes that will happen can be staged here,

PR #129, opened 2018-07-02
Wip/hughsie/reverse split

hughsie:
Fair warning; this is already running on the LVFS due to a small amount of panic from a major vendor. I'm happy to fix things up before pushing to master obviously.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -666,7 +666,7 @@ class Firmware(db.Model):
     filename = Column(Text, nullable=False)
     download_cnt = Column(Integer, default=0)
     checksum_upload = Column(String(40), nullable=False, index=True)
-    version_display = Column(Text, nullable=True, default=None)
+    _version_display = Column('version_display', Text, nullable=True, default=None)
     remote_id = Column(Integer, ForeignKey('remotes.remote_id'), nullable=False)
     checksum_signed = Column(String(40), nullable=False)
     user_id = Column(Integer, ForeignKey('users.user_id'), nullable=False)
@@ -708,6 +708,20 @@ class Firmware(db.Model):
         return datetime.datetime.now() + datetime.timedelta(seconds=secs)
 
     @property
+    def version_display(self):
+        if self._version_display:
+            return self._version_display
+        md_versions = []
+        for md in self.mds:
+            if md.version not in md_versions:
+                md_versions.append(md.version)
+        return ','.join(md_versions)
+
+    @version_display.setter
+    def version_display(self, value):
+        self._version_display = value
+
+    @property
     def problems(self):
         # does the firmware have any warnings
         problems = []
@@ -728,7 +742,7 @@ class Firmware(db.Model):
         self.timestamp = None
         self.filename = None        # filename of the original .cab file
         self.checksum_upload = None # SHA1 of the original .cab file
-        self.version_display = None # from the firmware.inf file
+        self._version_display = None # from the firmware.inf file
         self.download_cnt = 0       # generated from the client database
         self.checksum_signed = None # SHA1 of the signed .cab
         self.user_id = None         # user_id of the uploader
--- a/app/models.py
+++ b/app/models.py
@@ -121,10 +121,14 @@ class User(db.Model):
         if self.vendor_id == fw.vendor_id and self.user_id == fw.user_id:
             return True
 
+        # User was the person that uploaded this
+        if self.user_id == fw.user_id:
+            return True
+
         # something else
         return False
 
-    def check_for_vendor(self, vendor):
+    def check_for_vendor(self, vendor, allow_affiliates=False):
 
         # disabled accounts can never see firmware
         if not self.auth_type:
@@ -137,6 +141,13 @@ class User(db.Model):
         # manager user can modify any firmware matching vendor_id
         if self.is_vendor_manager and self.vendor_id == vendor.vendor_id:
             return True
+
+        # allow access for some actions for vendor affiliates
+        if allow_affiliates:
+            if self.vendor_id == vendor.vendor_id:
+                return True
+            if vendor.get_affiliation_by_odm_id(self.vendor_id):
+                return True
 
         # something else
         return False
@@ -224,6 +235,28 @@ class Restriction(db.Model):
     def __repr__(self):
         return "Restriction object %s" % self.restriction_id
 
+class Affiliation(db.Model):
+
+    # database
+    __tablename__ = 'affiliations'
+    __table_args__ = {'mysql_character_set': 'utf8mb4'}
+
+    affiliation_id = Column(Integer, primary_key=True, unique=True, nullable=False)
+    timestamp = Column(DateTime, nullable=False, default=datetime.datetime.utcnow)
+    vendor_id = Column(Integer, ForeignKey('vendors.vendor_id'), nullable=False)
+    vendor_id_odm = Column(Integer, ForeignKey('vendors.vendor_id'), nullable=False)
+
+    # link using foreign keys
+    vendor = relationship("Vendor", foreign_keys=[vendor_id], back_populates="affiliations")
+    vendor_odm = relationship("Vendor", foreign_keys=[vendor_id_odm])
+
+    def __init__(self, vendor_id, vendor_id_odm):
+        self.vendor_id = vendor_id
+        self.vendor_id_odm = vendor_id_odm
+
+    def __repr__(self):
+        return "Affiliation object %s" % self.affiliation_id
+
 class Vendor(db.Model):
 
     # sqlalchemy metadata
@@ -253,6 +286,9 @@ class Vendor(db.Model):
     # magically get the users in this vendor group
     users = relationship("User", back_populates="vendor")
     restrictions = relationship("Restriction", back_populates="vendor")
+    affiliations = relationship("Affiliation",
+                                foreign_keys=[Affiliation.vendor_id],
+                                back_populates="vendor")
 
     # link using foreign keys
     remote = relationship('Remote', foreign_keys=[remote_id])
@@ -287,6 +323,12 @@ class Vendor(db.Model):
         if self.is_uploading == 'na':
             val += 0x1
         return val
+
+    def get_affiliation_by_odm_id(self, vendor_id_odm):
+        for rel in self.affiliations:
+            if rel.vendor_id_odm == vendor_id_odm:
+                return rel
+        return None
 
     def __repr__(self):
         return "Vendor object %s" % self.group_id
--- a/app/models.py
+++ b/app/models.py
@@ -92,6 +92,18 @@ class User(db.Model):
             if self.is_qa or self.is_analyst:
                 return True
             return False
+        elif action == '@add-attribute-manager':
+            if not self.vendor.check_acl('@manage-users'):
+                return False
+            return self.is_vendor_manager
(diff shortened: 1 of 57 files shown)


app/models.py
```diff
@@ -611,7 +653,7 @@ class Firmware(db.Model):
     __table_args__ = {'mysql_character_set': 'utf8mb4'}
 
     firmware_id = Column(Integer, primary_key=True, unique=True, nullable=False, index=True)
-    vendor_id = Column(Integer, ForeignKey('vendors.vendor_id'), nullable=False)
+    vendor_id = Column(Integer, ForeignKey('vendors.vendor_id'), nullable=False) # the OEM, not ODM
```
superm1: It would be better to rename this to `oem_id` and down below not part of this PR where it refers to vendor to be `oem`.

app/templates/firmware-affiliation.html
```diff
@@ -0,0 +1,38 @@
+{% extends "default.html" %}
+{% block title %}Download Limits{% endblock %}
```
superm1: This title needs to be updated right?

app/templates/upload.html
```diff
@@ -26,15 +26,30 @@ <h2>Upload Firmware</h2>
...
 <form action="/lvfs/upload" method="post" enctype="multipart/form-data" class="form">
+{% if affiliations|length %}
+  <div class="form-group row">
+    <label for="target" class="col-sm-2 col-form-label">Upload for vendor</label>
+    <div class="col-sm-10">
+      <select class="form-control" name="vendor_id" required>
+        <option value="{{g.user.vendor.vendor_id}}">{{g.user.vendor.display_name}}</option>
+{% for r in affiliations %}
+        <option value="{{r.vendor.vendor_id}}">{{r.vendor.display_name}}</option>
+{% endfor %}
```
superm1: So in my experience with ODMs:
If there are more than one option available here it's probably best to make this default to a blank option that needs to be set on every upload.  This will help to prevent FW ending up at wrong OEM account.

app/views_firmware.py
```diff
@@ -309,6 +309,64 @@ def firmware_limit_add():
...
+        #if not vendor.get_affiliation_by_odm_id(v.vendor_id):
+        #    continue
+        vendors.append(v)
+
+    return render_template('firmware-affiliation.html', fw=fw, vendors=vendors)
+
+@app.route('/lvfs/firmware/<int:firmware_id>/affiliation/change', methods=['POST'])
+@login_required
+def firmware_affiliation_change(firmware_id):
+    """ Modifies the update urgency and release notes for the update """
```
superm1: Wrong comment here

app/views_vendor.py
```diff
@@ -349,3 +349,82 @@ def vendor_user_add(vendor_id):
...
+        return redirect(url_for('.vendor_affiliations', vendor_id=vendor_id), 302)
+    if not 'vendor_id_odm' in request.form:
+        return _error_internal('No value')
+
+    # check if it already exists
+    vendor_id_odm = int(request.form['vendor_id_odm'])
+    for rel in vendor.affiliations:
+        if rel.vendor_id_odm == vendor_id_odm:
+            flash('Failed to add affiliate: Already a affiliation with that ODM', 'warning')
+            return redirect(url_for('.vendor_affiliations', vendor_id=vendor_id), 302)
```
superm1: I guess this is going to be a manual task for merging accidental instances of `Odm`, `Odm Ltd`, `Odm Inc` etc?

superm1:
Something I think missing is updated documentation for "new" ODMs coming on board around how affiliates work.  Stuff that will land here: https://fwupd.org/vendors

app/templates/firmware-affiliation.html
hughsie: Fixed, thanks.

app/templates/upload.html
hughsie: So more like:?

screenshot from 2018-07-02 20-01-12

app/views_firmware.py
hughsie: Fixed, thanks.

app/views_vendor.py
hughsie: Yes, we can cross that bridge when we come to it :)

hughsie:
> So something that stands out to me is that if an ODM can have an account that is QA

Does an ODM account need QA users? I guess it's not beyond imagination that a vendor can be an OEM and a ODM (e.g. I can think of a couple of examples already).

I don't know if the code matches my understanding, but surely it would just make sense that a ODM QA user should only be able to promote firmware owned by the ODM? That was my logical interpretation, although I'm not sure those protections exist in the code.

Just so I understand it correctly, this is the test I was thinking about:

* add oem1 vendor
* add oem2 vendor
* add odm1 vendor
* make odm1 affiliated with oem1
* make odm1 affiliated with oem2
* add alice@odm1.com to vendor odm1 as a QA user
* add bob@odm1.com to vendor odm1 as a normal user
* bob uploads foo.cab on behalf of vendor oem1 (vendor_id = oem1, user_id=bob)

Is alice is able to see the firmware that was uploaded by bob? I'd say no.
Is alice able to promote the firmware uploaded by bob? Again I think no.

FireWolf74:
I have one question.
Can you do a GUI as well or can we get a list of the updates for each system?

app/templates/upload.html
superm1: Yeah that looks good to me.

superm1:
@FireWolf74 OT, please bring that discussion outside of this PR.

> Does an ODM account need QA users? I guess it's not beyond imagination that a vendor can be an OEM and a ODM (e.g. I can think of a couple of examples already).

Depends on how autonomously they operate from the OEM.  I'd say that is something that "should" be configurable because ODM can own the final QA but that's probably an OEM decision if they want to hand over that control.

>Is alice is able to see the firmware that was uploaded by bob? I'd say no.
> Is alice able to promote the firmware uploaded by bob? Again I think no.

Well I guess that's the confusing part.  Who is responsible to promote it to testing/stable then in this instance?  I guess only someone who is QA for `oem1`?  So that means that bob@odm1.com uploads the file and then someone at oem1 has to promote it (just like LVFS current workflow).

From the way you described this, it makes it sound like the only way an ODM QA user could promote files is if they are both an OEM and ODM.

hughsie:
>From the way you described this, it makes it sound like the only way an ODM QA user could promote files is if they are both an OEM and ODM.

The LVFS doesn't actually distinguish between an ODM or an OEM, to it they are both just instances of Vendor. I think the only way an ODM QA user should be able to move to testing and stable is for the firmware to be "owned" by the ODM, so **never** when the firmware is uploaded "on behalf of" the OEM.

hughsie:
> because ODM can own the final QA

In that case the ODM should upload firmware with the vendor_id owner of the ODM, and then when the firmware is QAd it can have the vendor re-assigned even in the stable remote. It might not even need to be "owned" by the OEM at all if the ODM is responsible for the QA and testing.

hughsie:
> Something I think missing is updated documentation for "new" ODMs coming on board around how affiliates work

How about the last commit?

app/templates/vendors.html
```diff
@@ -77,4 +77,32 @@ <h3 class="card-header list-group-item-info">Why does the LVFS project sign the
...
+  There are some OEMs where the ODM will be the entity responsible for
+  uploading the firmware to the LVFS.
+  The per-device QA can either be done by the OEM or the ODM as required.
+</p>
+<p>
+  The LVFS administrator can mark other vendors as <em>affiliates</em> of
+  other vendors.
+  This gives the ODM permission to upload firmware that is <em>owned</em> by the
+  OEM on the LVFS, and would appear in the OEM embargoed metadata.
+  The OEM QA team is also able to edit the update description, move the firmware
```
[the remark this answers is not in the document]
hughsie: Works for me, thanks.

app/templates/user-admin.html
```diff
@@ -5,6 +5,17 @@
...
+<div class="alert alert-warning mt-1" role="alert">
+  <strong>BE VERY CAREFUL EDITING THE ADMIN</strong>.
+</div>
+{% elif u.user_id == g.user.user_id %}
+<div class="alert alert-warning mt-1" role="alert">
+  <strong>Be careful editing your own profile:</strong>
+  If you remove attributes they may need to be re-added manually by the LVFS administrator.
+</div>
+{% endif %}
+
```
superm1: Rather than the heavily worded message, why not hide the modify button and mention it's read only?

app/models.py
```diff
@@ -49,6 +49,7 @@ class User(db.Model):
     is_qa = Column(Boolean, default=False)
     is_analyst = Column(Boolean, default=False)
     is_vendor_manager = Column(Boolean, default=False)
+    is_approved_public = Column(Boolean, default=False)
```
superm1: I think i'd prefer to see `approved_promoter` to be clearer what this new permission level means.  I think it makes sense though.

hughsie: I went with public as we have a `Remote.is_public` to mean stable and testing. What I'm trying to say is "we know the user can legally distribute what they've uploaded and we trust them enough to push to stable". Maybe `is_approved_stable` instead? or just `is_approved`?

app/templates/user-admin.html
hughsie: I think there are legitimate reasons why the admin or vendor admin remove attributes or change other things like the legacy username.

app/models.py
superm1: Well I guess since they can do both testing and stable, I get what you're trying to convey.  Approved public makes enough sense.

app/templates/user-admin.html
superm1: But so you can guard against the really bad stuff by hiding it then or turning a box grey if it's gonna be dangerous for example.  Like don't let them remove the admin attribute from themselves.

app/templates/firmware-affiliation.html
```diff
@@ -13,6 +13,13 @@ <h2 class="mt-3">Vendor Affiliation</h2>
   owner the ablity to change the target and even delete the firmware.
 </p>
 
+{% if not g.user.check_acl('@admin') %}
+<div class="alert alert-warning mt-1" role="alert">
+  If you move this firmware to an different vendor it will not be possible
+  to edit the update details or move the firmware to testing or stable.
```
superm1: I think you should make it clear that the new owner can still do this.  Proposal:

"It will not be possible **for you** to edit the update details or move the firmware to testing or stable.  The new vendor will be able to perform those actions"

app/models.py
```diff
@@ -727,6 +685,81 @@ def __init__(self):
         self.user_id = None         # user_id of the uploader
         self.mds = []
 
+    def check_acl(self, action, user=None):
```
superm1: I feel like you need a  helper function here for `is_owner(self, user)`.  Adding that you probably don't need most of the comments "is QA group and firmware owner" as it becomes more readable.

superm1:
This is of course a huge PR so I haven't scrutinized super closely, but just a few small comments from me and it seems fine.

app/templates/user-admin.html
hughsie: Right; checking this you can't actually "unadmin" yourself, so I've removed the all caps warning.

app/models.py
hughsie: Good idea, fixed.

PR #130, opened 2018-07-06
Move the high-level introduction to GitHub

hughsie:
I'm intending to write a user-guide for OEMs and ODMs too. I'll probably do that during my offline holiday in a few weeks time.


docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
+# Linux Vendor Firmware Service
+
+Copyright 2016-2018, Richard Hughes, Red Hat
+
+## Abstract
+
+Updating firmware on devices is very hard for users on Linux systems.
```
superm1: My thought on better wording:
`s/very hard/traditionally difficult/`

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+Not knowing exact hardware details, where to look for updates and how to run the
+Windows-specific flashing tools makes it almost impossible to update firmware on
+devices.
+As a result, “_broken_” hardware is being returned to the vendor and customer
```
superm1: `s/and/or/`

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+systems are left in an insecure state even when updates have been released that
+fix the specific issues. Linux as the OS is now mainstream and vendors need to
+support these customers.
+
+The LVFS is a secure web service that can be used by OEM's to upload firmware
+archives and can also be used by users to securely download metadata about
+available updates and optionally, the updates themselves.
+The LVFS is growing fast, delivering metadata to over 30 million clients every month.
```
superm1: With this text living in LVFS, I suspect you might be able to have this dynamically update?

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+
+**Hundreds of thousands of customer devices are being updated every month thanks to the LVFS.**
```
superm1: This could also be dynamic perhaps?

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+- RHEL 7
+- Fedora 22 or newer
+- Ubuntu 16.04 (Xenial) or newer
+- Debian 9 (Stretch) or newer
+
+At session start-up (and on every start thereafter) the metadata XML and detached
+signatures are checked for a specified age, and if required newer files are
+ automatically downloaded from the LVFS and pushed into fwupd over D-Bus.
+When the update list is required we query the fwupd daemon over D-Bus for any
+pending updates.
```
superm1: This might need some rewording since some distributions will not download metadata from LVFS automatically.

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+Other files are permitted in the archive which allows the same deliverable to be
+used for the Windows Update system.
+
+Internally fwupd creates a device with a unique ID, and then a number of GUIDs
+are assigned to the device by the provider.
+It is these GUIDs specified in the update metadata file that are used to match
+a firmware file to a device.
+Although it is usually the responsibility of the system vendor to generate a new
+GUID if the hardware requires a different firmware file, we can match an update
+that only applies to specific versions of hardware using CHID GUIDs.
```
superm1: May as well include a URL to https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/using-chids

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+The fwupd project ships a number of plugins:
+
+- `UEFI`, which reads the ESRT[4] table and adds devices found on the system and
+  then schedules an offline update to run a custom EFI binary.
+- `DFU`, which reads and writes data from USB devices supporting the DFU interface.
+- `ColorHug`, which supports a custom HID protocol.
+- `CSR`, which supports the Cambridge Silicon Radio protocol.
+- `Logitech`, which supports the Unifying protocol.
+- `WacomHID`, which supports the Wacom update protocol.
+- `Thunderbolt`, which supports Thunderbolt controllers and devices.
```
superm1: Missing Dell, Synaptics MST, and the various helper plugins here

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+
+Adding more plugins is of course possible, but where possible vendors should use
+the existing code and for instance add an ESRT data table when building the
+system firmware.
+
+When the user agrees to a UEFI firmware update the firmware is unpacked into the
```
superm1: I think this should be a subsection around updates that apply immediately in OS versus UEFI updates

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+EFI System Partition, several UEFI keys are set and the system reboots.
+On reboot the `fwup.efi` binary is run before the boot-loader is started and
+the `UpdateCapsule` action is performed.
```
superm1: the BIOS's 'UpdateCapsule' UEFI runtime source is called.

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+For USB devices the device is normally updated live without requiring a reboot.
```
superm1: Several other devices can be flashed live, (Thunderbolt, Synaptics).

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+## Performance
+
+The LVFS runs on a dedicated virtual private server and uses a CDN for the
+public metadata.
+Every month there are over 50 million files being downloaded from around
+10 million unique clients.
+About 5% of the firmware downloads are performed using the `fwupdmgr` command
+line client manually and 95% are done from the GNOME Software GUI program.
+
+There are currently over 160 devices supported on the LVFS with over 400
```
superm1: This could be dynamically built.

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+The LVFS has grown to be an essential part of the Linux ecosystem and there
+are currently 26 vendors using the service, 11 of which are multi-billion
+dollar companies.
+4,000 **new** clients use the service _every day_, with over 3,000,000
+returning users.
+The LVFS is a mature service providing important functionality for Linux users.
+
+## Future Work
+
+Various vendors are working on custom providers for fwupd as they either cannot
```
superm1: s/providers/plugins/

docs/Introduction.md
```diff
@@ -0,0 +1,257 @@
...
+
+[3]: [Vendor page for fwupd](http://www.fwupd.org/vendors)
+
+[4]: [ESRT Table Definition](https://msdn.microsoft.com/en-us/windows/hardware/drivers/bringup/esrt-table-definition)
+
+[5]: [GNOME Software Homepage](https://wiki.gnome.org/Apps/Software)
+
+[6]: [AppStream v0.9 Documentation](https://www.freedesktop.org/software/appstream/docs/)
+
+[7]: [Wyse Cloud Client Manager](https://www.cloudclientmanager.com/)
```
superm1: Since this will live in a live HTML page, I think it's better to move the citations directly into places they're referenced.

superm1:
> offline holiday 

Take some time off too~ :)

docs/Introduction.md
hughsie: Fixed.

docs/Introduction.md
hughsie: Fixed.

docs/Introduction.md
hughsie: Added,

docs/Introduction.md
hughsie: Added the "action" ones, I don't think the helper ones need listing yet.

docs/Introduction.md
hughsie: Maybe fixed?

docs/Introduction.md
hughsie: Fixed.

docs/Introduction.md
hughsie: Fixed.

docs/Introduction.md
hughsie: Fixed.

docs/Introduction.md
hughsie: Agreed, thanks. Fixed.

docs/Introduction.md
hughsie: I'm not sure how; any ideas?

hughsie:
If you see anything obviously wrong, please just go ahead and make commits; I'll squash it down when we're happy with it.

docs/Introduction.md
superm1: Oh I figured you were able to easily get the download count since you were building it for telemetry anyway.  If not, then it's probably better to just remove this statement.

docs/Introduction.md
```diff
@@ -0,0 +1,251 @@
...
+
+[GNOME Software](https://wiki.gnome.org/Apps/Software) is an application store
+designed to make installing, removing and updating both easy and beautiful.
+It is available for Linux and used by millions of people on the following
+distributions:
+
+- RHEL 7
+- Fedora 22 or newer
+- Ubuntu 16.04 (Xenial) or newer
+- Debian 9 (Stretch) or newer
```
superm1: I guess you've seen tons of other distros at this point.  Maybe can include the earliest arch, suse, gentoo etc that have been hitting LVFS too?

superm1: Also CentOS I guess?

hughsie: Fixed.

PR #134, opened 2018-07-09
Wip/hughsie/trivial

hughsie:


superm1:
All the other fixes than d945663 look fine though.

superm1:
Here's an idea of what I had in mind:

untitled_page

If logged in an account that is marked deprecated get redirected to a page like that.

If they want to use the account anyway, they have to put a reason in that text box to hit the button to use the group account, and that reason will be emailed to whoever the is the OEM user manager.

Then when 9/1/2018 rolls around, rather than disable that account, just strip it from all it's permissions and change that page to tell people this account is no longer available, you need to login with OAuth.
If you have any problems logging with Oauth, contact `$USER_MANAGER`.

hughsie:
Hmm, I understand your mockup, but it's a fair chunk of work that's going to be obsolete in less than 60 days... This is what the simple per-user warning would look like:
screenshot from 2018-07-09 20-18-37

superm1:
OK, how about this instead.  Rather than just flash the message on top of that page, flash it on top of ALL pages.  That will make it more apparent at least..

hughsie:
> flash it on top of ALL pages

Then it's not really a `login_warning` -- perhaps `auth_warning`, `page_warning` or `account_warning`?

superm1:
Yeah something like that makes a lot of sense to me.

PR #135, opened 2018-07-09
Move the secure response to the nginx config file

hughsie:
It certainly seems a lot more natural place, and hijacking the response like
this triggers a runtime error in newer python versions.


superm1:
I guess it makes sense to also include a sample nvinx configuration file in tree with this change to make it easy to implement this in a local instance.

hughsie:
> also include a sample nvinx configuration file

Is https://github.com/hughsie/lvfs-puppet/blob/wip/hughsie/add_header/lvfs.pp#L229 good enough? Forthe .`/app.py` you don't need any headers set, it's only really applicable per-deployment.

superm1:
Yeah I think that's good, it wasn't obvious to me that they were stored there until you linked it.

Issue #141, opened 2018-07-10
LVFS login can't reset forgotten password


superm1:
In reviewing with folks to arrange LVFS accounts for local accounts for individual people at ODM side a point raised was that LVFS can't currently handle the situation for forgotten password.

Folks might not log into LVFS frequently and with good security policies shouldn't be resuing their passwords in LVFS  with other things.  For this scenario it's possible that they'll forget passwords and need to reset them.

Can you please add a password reset flow for local accounts (eg "Forgotten password")?

PR #144, opened 2018-07-10
Allow the user to recover from a lost password

hughsie:
Fixes https://github.com/hughsie/lvfs-website/issues/141

### DIFF
--- a/app/templates/email-recover-password.txt
+++ b/app/templates/email-recover-password.txt
@@ -1,0 +1,16 @@
+Dear {{ user.display_name }},
+
+Your password has been reset Please click on the following link to log in to the LVFS:
+
+{{ url_for('.index', _external=True) }}
+
+Your credentials have been set as:
+
+Username: {{ user.username }}
+Password: {{ password }}
+
+You should also change your password when you log in by clicking menu items 'Profile' then 'User'.
+
+Regards,
+
+The LVFS admins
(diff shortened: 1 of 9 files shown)


app/templates/email-recover-password.txt
```diff
@@ -0,0 +1,16 @@
+Dear {{ user.display_name }},
+
+Your password has been reset Please click on the following link to log in to the LVFS:
+
+{{ url_for('.index', _external=True) }}
+
+Your credidentials have been set as:
```
superm1: spelling error

app/views_user.py
```diff
@@ -223,6 +223,65 @@ def user_modify_by_admin(user_id):
...
+        return _error_permission_denied('Unable to recover user as no username')
+
+    # check exists
+    username = request.form['username']
+    user = db.session.query(User).filter(User.username == username).first()
+    if not user:
+        flash('Unable to recover password as no username %s found' % username, 'warning')
+        return redirect(url_for('.index'), 302)
+
+    # password is stored hashed
```
superm1: I think there needs to be a check in this flow if their email is configured as local or oauth and abort if they're oauth.

app/templates/email-recover-password.txt
hughsie: Fixed, thanks.

app/views_user.py
hughsie: ...and if they're a robot, or a locked account. Fixed, thanks.

superm1:
Flow looks good to me now, but I think it would be good to include a test suite test for this if possible since it's a multi step process that's potentially fragile.  I think test suite test could avoid sending/receiving parsing the email and just snoop the token though.

Issue #142, opened 2018-07-10
When creating a new account there should be a checkbox to set it as local or oauth immediately


superm1:
Yesterday I created a new account for someone.  After it was created I clicked the details button and changed it to OAuth only.

They told me that when the account was created they received an email indicating what their password was so they tried to login with that.  It of course failed since they were supposed to login with OAuth.

So I would like if either the OEM policy can be configured to default to OAuth account or to have a slider as part of the account creation that lets me set it to OAuth immediately.

PR #145, opened 2018-07-10
If OAuth is configured don't create local users

hughsie:
Fixes: https://github.com/hughsie/lvfs-website/issues/142

### DIFF
--- a/app/views_vendor.py
+++ b/app/views_vendor.py
@@ -330,20 +330,27 @@ def vendor_user_add(vendor_id):
             return redirect(url_for('.vendor_users', vendor_id=vendor_id), 302)
 
     # add user
-    password = _generate_password()
-    user = User(username=request.form['username'],
-                display_name=request.form['display_name'],
-                auth_type='local',
-                password=_password_hash(password),
-                vendor_id=vendor.vendor_id)
+    if g.user.vendor.oauth_domain_glob:
+        user = User(username=request.form['username'],
+                    display_name=request.form['display_name'],
+                    auth_type='oauth',
+                    vendor_id=vendor.vendor_id)
+    else:
+        password = _generate_password()
+        user = User(username=request.form['username'],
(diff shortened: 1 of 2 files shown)


app/views_vendor.py
```diff
@@ -330,20 +330,27 @@ def vendor_user_add(vendor_id):
...
     # send email
-    send_email("[LVFS] An account has been created",
-               user.username,
-               render_template('email-confirm.txt',
-                               user=user, password=password))
+    if user.auth_type == 'local':
+        send_email("[LVFS] An account has been created",
+                   user.username,
+                   render_template('email-confirm.txt',
+                                   user=user, password=password))
```
superm1: So I think I'd still like to see an email go out if the user was OAuth, particularly notifying them to log in using the big OAuth button instead of trying a password.

hughsie: Doesn't that mean you have a race on your hands adding and then modifying the user before they try and login?

superm1: Do you mean like giving them appropriate permissions?  I suppose yes that's true.
They won't get a welcome email though unless it happens here though right?

I don't think it's a big deal if they log in before they get appropriate permissions, they just can't do as much :)

My main concern is getting people to stay away from the local area, so an email telling them to login with oauth would be useful in that sense.

hughsie: Wouldn't they get an email when you modify the account and have added the right permissions?

superm1: I guess that's true.  Could you have the template for `email-modify.txt` mention how to login with oauth when applicable perhaps then?

hughsie: Something like `You can access the LVFS by clicking the 'Login' button then using the 'Log in with Azure AD' option.` ?

superm1: Yeah that would be perfect.

Issue #143, opened 2018-07-10
Potentially allow an overwrite action on upload


superm1:
I'm exploring with a team the option to have LVFS connected to some build server automation.  Builds released to customers will always have increasing version numbers.  For internal development purpose however its not uncommon multiple builds with the same version number.  Once they're released to a wider audience than the will be incremented for the next build.

For example there might be 10 builds that are "candidate" builds for the `1.0.0` firmware that a product may launch with.

So because of this I'm wondering what you would think about a hidden checkbox as part of the upload POST that would allow overwrite (if in embargo channel).  That overwrite would actually perform like a delete and then upload the new one.

hughsie:
Wouldn't it be better to just 'delete' the previous firmware from the automated script? We could provide API to get the firmware_id from the firmware checksum or GUID+version. I'm hesitant to encode too much magic into the upload process. Perhaps a semi-hidden `replace()` method might be a better compromise.

superm1:
The problem so far has been being able to get `firmware_id` to know what to delete in advance, so an API to get that would work too.  The slight problem with getting the firmware ID from the checksum or GUID is that there might be a bunch of different payloads in the cab so it's going to be unclear which one is the one that LVFS pivots from.  I think it's usually the first alphabetically but then that's hidden magic too.

That's why I was thinking `overwrite/replace` makes sense.

PR #146, opened 2018-07-10
Autodelete duplicate firmware uploaded by robots

hughsie:
Fixes: https://github.com/hughsie/lvfs-website/issues/143

110% untested.

### DIFF
--- a/app/views_firmware.py
+++ b/app/views_firmware.py
@@ -135,19 +135,7 @@ def firmware_undelete(firmware_id):
     flash('Firmware undeleted', 'info')
     return redirect(url_for('.firmware'))
 
-@app.route('/lvfs/firmware/<int:firmware_id>/delete')
-@login_required
-def firmware_delete(firmware_id):
-    """ Delete a firmware entry and also delete the file from disk """
-
-    # check firmware exists in database
-    fw = db.session.query(Firmware).filter(Firmware.firmware_id == firmware_id).first()
-    if not fw:
-        return _error_internal("No firmware file with ID %s exists" % firmware_id)
-
-    # security check
-    if not fw.check_acl('@delete'):
-        return _error_permission_denied('Insufficient permissions to delete firmware')
+def _firmware_delete(fw):
 
     # find private remote
     remote = db.session.query(Remote).filter(Remote.name == 'deleted').first()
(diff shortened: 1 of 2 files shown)


superm1:
Looks good to me, if you want to put it on staging and mark the thunderbolt user as a robot on staging I'll give it a test run. (My Oauth manager account can't log into staging IIRC to do this).

hughsie:
Both done.

superm1:
I'm getting 502 errors both from uploader script on staging and trying to login from browser.

hughsie:
Urgh, the same issue we had on production a few days ago. Fixed, sorry.

superm1:
Seems passing `auto-delete` isn't working still.

Passing like this:

``` 
r = requests.post(uri, data={"target": target, "auto-delete" : "true"},
                                   files={'file': f},
                                   cookies=self.cookies)
```

And still getting `Failed to upload file: A file with hash ...`

hughsie:
> And still getting Failed to upload file: A file with hash ...

Right, as expected. If you change the description in the metainfo, recompress the .cab file, then submit it should auto-delete; there is no point deleting firmware `foo` and then re-uploading the exact same firmware `foo` -- I thought the requirement was for firmware that had changed without bumping the version number...

superm1:
> Right, as expected. If you change the description in the metainfo, recompress the .cab file, then submit it should auto-delete; there is no point deleting firmware foo and then re-uploading the exact same firmware foo -- I thought the requirement was for firmware that had changed without bumping the version number...

Oh duh - sorry brain fart in my testing.  I changed a bit in the firmware, rebuilt the CAB an re-uploaded and it worked properly, saw the old file deleted and new one created.

hughsie:
Ahh, good, thanks for testing. I've added an eventlog item (as well as the existing `FirmwareEvent`) so that the admin knows what's happened. I'm also wondering if we want to actually delete (rather than archive) when we have a large number of "archived" firmwares. In which case, it might be good to add metadata to the firmware to say that it was autodeleted due to a better version coming up, and put those first in line to be deleted.

I also think we want to add a check to the autodelete that the firmware is indeed in embargo or private, correct? I suppose returning with a error if we tried to autodelete something in a public remote is the right thing to do, right?

superm1:
> Ahh, good, thanks for testing. I've added an eventlog item (as well as the existing FirmwareEvent) so that the admin knows what's happened. I'm also wondering if we want to actually delete (rather than archive) when we have a large number of "archived" firmwares. In which case, it might be good to add metadata to the firmware to say that it was autodeleted due to a better version coming up, and put those first in line to be deleted.

That's good on the events.  Actually I think auto-deleted firmware doesn't need to stay around at all.  I think it should be permanently removed.  The use case for it is such that if tests failed, the build is rerun until they pass.  You only want the "latest" of that "version".

> I also think we want to add a check to the autodelete that the firmware is indeed in embargo or private, correct? I suppose returning with a error if we tried to autodelete something in a public remote is the right thing to do, right?

Robots by definition shouldn't be able to demote firmware, but I guess that's a good safeguard to keep in place.

hughsie:
> I think it should be permanently removed

If that's the case then I think we need to be a lot more careful about matching old firmwares and thinking about things like OEM and ODM ownership and who gets to permanently delete firmware, especially when potentially "owned" by someone else.

superm1:
I think to me only robots should be able to do this and only to firmware they uploaded.  That should help prevent cross oem/odm contamination.

hughsie:
I've rebased the patch. I'm not sure if permanently deleting is a good idea; we need the firmware archives for audit reasons and they'll be referenced in the metadata until the server cron job runs, and the user downloads new metadata. I'm fairly happy with the current patch, so I'm happy to deploy that with your blessing.

hughsie:
I'll deploy this at the weekend -- thanks for the review.

hughsie:
I've just deployed this. Yell if there are any problems.

PR #147, opened 2018-07-11
Provide functionality to anonymize an existing dataset

hughsie:
I often get asked to demo the LVFS, but it's hard to show the admin interface
without accidentally disclosing embargoed firmware and any vendors who are
trialling the LVFS in secret.

With all the real data anonymized this is a much simpler thing to do.


superm1:
I suppose you would only be using this for "local" demos right?  Looks good to me.

hughsie:
> I suppose you would only be using this for "local" demos right?

Correct; I think staging should stay as it is now.

PR #150, opened 2018-08-08
Wip/hughsie/analytics

hughsie:


hughsie:
screenshot from 2018-08-08 10-04-09
screenshot from 2018-08-08 10-05-14
screenshot from 2018-08-08 10-05-18
screenshot from 2018-08-08 10-05-55

superm1:
If you're going to be adding graphs like this, I think one that would be really interesting to the public would be a mix of OSes contacting LVFS for downloads.

hughsie:
Would that effect which hardware they chose to buy? I'm also hesitant to do this for unauthed users as we've only been collecting the full useragent for a few months. Perhaps it makes more sense per-firmware for logged in users.

superm1:
I guess probably wouldn't make a buying decision there, but would show where fwupd is heavily used and better supported (but I guess we already know that - Fedora, Arch and Ubuntu).

hughsie:
I assume you've not got any ideological or business issues with merging this kind of thing?

superm1:
I haven't commented yet because I wanted to discuss it with some others and haven't  had chance yet.
I think the one thing I'm not crazy about is publicly showing downloads per OEM with a number associated. I expect others wouldn't like that either.

What do you think about showing that pie chart with % of total LVFS downloads? It shows similar information then but is more obfuscated.

The other graphs I don't think are problematic though.

hughsie:
screenshot from 2018-08-11 10-31-08

PR #151, opened 2018-08-10
Show the newest supported hardware in the devicelist

hughsie:


hughsie:
screenshot from 2018-08-10 14-55-18

hughsie:
> I think you want to show new FW versions rather than new devices.

I think this is a different thing; this PR is trying to address the question of "what new hardware is supported on the LVFS" rather than "what new firmware was released" which is subtly different but means we only include the "first" time the device has a firmware update on the LVFS.

The idea was to promote new devices, rather than new firmware. I think for a firehose of new firmware an RSS feed does make sense, but I do wonder who would read it.

PR #154, opened 2018-08-12
Fix broken link in vendors.html

abitrolly:


app/templates/vendors.html
```diff
@@ -94,7 +94,7 @@ <h1>Working with ODMs</h1>
 </p>
 <p>
   The ODM vendor account also doesn't have to appear in the
-  <a href="/devicelist">search results</a> or the <a href="/vendorlist">vendor list</a>,
+  <a href="/lvfs/devicelist">search results</a> or the <a href="/vendorlist">vendor list</a>,
```
hughsie: The original plan was to restrict `/lvfs/` prefixed paths to logged in users, and have `/random` for non-auth'd users. I think I've broken that rule a few times already :) I think the proper way to fix this is to use `url_for('.device_list')` and then let flask work out the path to use. There are a ton of these to fix in the templates, and I really don't expect you to fix them all :)

abitrolly: Fixed in 5d16c24cd2bb. No worries. :D

I tried to read how `{{url_for('.device_list')}}` works in Flask. Looks like 'device_list` should be a function name, and if this function is defined in different modules or blueprints? Also I am not sure what `.` dot means here. Because I don't use Flast it is kind of hard to find time to read its docs completely.

Why restrict device list and vendor list to logged in users? This information is totally awesome being open! :D

hughsie: A lot of flask is reading the documentation and then going "that makes no sense!" and then after using it for a few days it's completely awesome. :)

> Why restrict device list and vendor list to logged in users? 

They're not anymore; the ability to modify the vendor list is obviously admin-only but most of the data is now public. I don't think the `/lvfs/` and `/` split is terribly useful anymore.

If you'd like to, I'd really like to see all the templates using `url_for` -- you might have to start a flask instance locally to test those tho.

abitrolly: After reading https://stackoverflow.com/questions/19261833/what-is-an-endpoint-in-flask it makes sense, but `endpoint` is a poor name for this kind of abstraction. The common understanding is that [`endpoint` is an URL](https://stackoverflow.com/questions/9807382/what-is-a-web-service-endpoint). `routename` could make it all easy.

Still can not figure out how `.` dot is used?

hughsie: > Still can not figure out how . dot is used?

I'm not sure myself, sorry.

abitrolly: Filled RFE - https://github.com/pallets/flask/issues/2885

`.` is now more or less clear - like a relative route for Blueprints.

PR #156, opened 2018-08-14
<pre>-format update descriptions

abitrolly:
Because they are plain text - https://fwupd.org/lvfs/device/52dd29cd-24b9-4d60-8a45-e72670901924

### DIFF
--- a/app/templates/device.html
+++ b/app/templates/device.html
@@ -74,7 +74,7 @@
   </tr>
   <tr class="row">
     <th class="col-sm-2">Description</th>
-    <td class="col-sm-10">{{fw.mds[0].release_description|safe}}</td>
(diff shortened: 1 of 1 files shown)


hughsie:
Well, they're not meant to be plain text; Lenovo are just not doing a very good job at writing AppStream description markup. I'll talk to them on Friday about this

abitrolly:
Is that the specification - https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-description? What is the benefit if that format over the plain text? I see only downside of not very effectively spending developers time.

hughsie:
I think this is fixed in a better way with https://github.com/hughsie/lvfs-website/pull/187

abitrolly:
That's a huge alternative piece of code. :D If there is a specification for description format, it is nice to see it referenced.

PR #158, opened 2018-08-20
Wip/hughsie/hwinfo

hughsie:


superm1:
So my knee jerk reaction is this is something that probably shouldn't be in the production LVFS but only ever be activated in staging.  Otherwise isn't it inviting a DOS attack that tries to fill up storage?

hughsie:
It's certainly a reasonable reaction. Let me see what I can do.

hughsie:
@superm1 what about something like the fixup commit above? The idea would be that the testing instance would have the default (nvme) and the stable instance would have an empty string.

superm1:
That looks better to me yes.

PR #168, opened 2018-10-06
Wip/hughsie/nuke

hughsie:
This probably needs some careful review, especially on the model changes.

### DIFF
--- a/app/templates/firmware-details.html
+++ b/app/templates/firmware-details.html
@@ -19,6 +19,9 @@
 {% if not fw.is_deleted and fw.check_acl('@delete') %}
       <button type="button" class="btn btn-block btn-danger" data-toggle="modal" data-target="#deleteModal">Delete</button>
 {% endif %}
+{% if fw.is_deleted and fw.check_acl('@nuke') %}
+      <button type="button" class="btn btn-block btn-danger" data-toggle="modal" data-target="#nukeModal">☠ Nuke ☠</button>
+{% endif %}
     </td>
   </tr>
   <tr class="row">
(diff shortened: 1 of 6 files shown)


app/views_firmware.py
```diff
@@ -176,6 +177,37 @@ def firmware_delete(firmware_id):
...
+        return _error_internal("No firmware file with ID %s exists" % firmware_id)
+
+    # security check
+    if not fw.check_acl('@nuke'):
+        return _error_permission_denied('Insufficient permissions to nuke firmware')
+
+    # really delete firmware
+    path = os.path.join(app.config['DOWNLOAD_DIR'], fw.filename)
+    if os.path.exists(path):
+        os.delete(path)
```
superm1: What kind of circumstances would cause the file to not exist?  I almost wonder if you should flash a message when removing the entry from the database but the file was already gone for example.

hughsie: I don't think this can happen on the live instance. When testing the only way this is reproducable is when I've loaded a newer snapshot of the mysql database but not rsynced the firmware files.

app/templates/firmware-details.html
```diff
@@ -18,6 +18,9 @@
     <td class="col-sm-2 text-right">
 {% if not fw.is_deleted and fw.check_acl('@delete') %}
       <button type="button" class="btn btn-block btn-danger" data-toggle="modal" data-target="#deleteModal">Delete</button>
+{% endif %}
+{% if fw.check_acl('@nuke') %}
```
superm1: I think it's worth adding another check to only show the nuke button when the fw is already deleted.  You shouldn't be able to nuke something unless you've gone through the steps to "delete" it first.

superm1:
You're missing some U/I to give permissions to users for "nuke" permission in `views_user.py` aren't you?

app/templates/firmware-details.html
hughsie: That's a good question. I didn't think we'd delete and /then/ nuke -- I figured nuke would be something you can do on a firmware in any stage. Maybe deleted makes sense.

hughsie:
> You're missing some U/I to give permissions to users for "nuke" permission in views_user.py aren't you?

This was going to be admin only; do you think this should be available for users? Seems a bit dangerous to remove things like the firmware event log.

superm1:
> This was going to be admin only

If it's admin only, then why add nuke permissions in the first place?

hughsie:
> If it's admin only, then why add nuke permissions in the first place?

We need to be able to check is the user is in the admin group; it seemed a good idea to push all the logic in the same place.

hughsie:
Okay, self tests added (and bugs fixed!) -- I've also changed it so you have to delete first, then you can nuke as you suggested. If nothing else, it means less skull-and-crossbones buttons when logged in as the admin :)

PR #171, opened 2018-10-12
Set a ReportURI in the embargo remote

hughsie:
This allows users and QA testers to upload reports to the LVFS when the firmware
archive is downloaded from an embargoed remote.


superm1:
I had been wondering about why reports didn't work for embargo targets but wasn't bothered enough to investigate, thanks.

PR #172, opened 2018-10-13
Adding KDE Discover Information on End User Page 

abhijeet2096:
From KDE  Plasma 5.14 release (https://www.kde.org/announcements/plasma-5.14.0.php), KDE Discover has added fwupd support and this P.R adds this info on the end user page.


hughsie:
Can you do this without the merge commit please.

superm1:
I'm curious does it use libfwupd or directly talk over dbus?

Also would be good to mention here:
https://github.com/hughsie/fwupd/blob/master/README.md

PR #177, opened 2018-10-18
Allow components to have a priority

hughsie:
This allows us to set the priority of a specific component to be largest, which
allows fwupd to order the results with the 'main' entity listed first.

The other way to do this is to simply add the files into the source .cab file
in a specific order, but I think that's fragile. It's better to be explicit
rather than relying on an assumption that the file order won't be changed when
the archive is repacked.

### DIFF
--- a/app/metadata.py
+++ b/app/metadata.py
@@ -40,6 +40,7 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri=''):
                 component.set_metadata_license(md.metadata_license)
                 component.set_project_license(md.project_license)
                 component.set_developer_name(None, md.developer_name)
+                component.set_priority(md.priority)
                 components[md.appstream_id] = component
             else:
                 component = components[md.appstream_id]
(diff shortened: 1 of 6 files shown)


hughsie:
> this PR doesn't actually influence which one is listed first though

Listed first in the firmware view, or in the .cab file itself?

hughsie:
> I think some documentation about what priority levels are accepted and how they are used would be good

in the fwupd project, or in the lvfs-website project? It kinda applies to both... In this case priority 1 is better than priority 0 (the default) -- so you want the "main" firmware to have a larger number than the "sub" firmwares.

hughsie:
> Firmware view

Good catch, fixed.

superm1:
> in the fwupd project, or in the lvfs-website project?

I agree it should be documented in both.

> Good catch, fixed.

Can you also deploy this to staging?  I can experiment to make sure everything is working how I expected as well.

PR #179, opened 2018-10-19
Add some more documentation to the LVFS

hughsie:
I can't keep telling billion-dollar companies to read my assorted ramblings on
https://blogs.gnome.org/hughsie


docs/Affiliates.md
```diff
@@ -0,0 +1,37 @@
...
+The LVFS administrator can mark other vendors as *affiliates* of other vendors.
+This gives the ODM permission to upload firmware that is *owned* by the OEM to
+the LVFS, and that appears in the OEM embargo metadata.
+The OEM QA team is also able to edit the update description, move the firmware
+to testing and stable (or delete it entirely) as required.
+The ODM vendor account also doesn't have to appear in the search results or the
+vendor table, making it hidden to all users except ones from the OEM.
+
+This also means if an ODM like Foxconn builds firmware for two different OEMs,
+they also have to specify which vendor should own the firmware at upload time.
```
superm1: You should also mention the ability to change this later in the event it's internal testing that should be made available to OEM later.

docs/CompositeFirmware.md
```diff
@@ -0,0 +1,41 @@
...
+are better. For example:
+
+    <?xml version="1.0" encoding="utf-8"?>
+    <component priority="1" type="firmware">
+      <id>com.hughski.dock.firmware</id>
+      <name>Hughski Dock Update</name>
+      ...
+    </component>
+
+and also:
```
superm1: I think you need to make it clear that each of these is metainfo XML describing one individual component.  This reads to me like you might need to put them all in the same XML.

docs/Telemetry.md
```diff
@@ -0,0 +1,113 @@
...
+
+We obviously have to be careful with users privacy and trust. We can just enable
+automated collection, and this document outlines what we did for fwupd >= 1.0.4
+This functionality should be acceptable to even the most paranoid of users.
+
+The `fwupd` daemon stores the result of each attempted update in a local SQLite
+database.
+In the event there is a firmware update that has been attempted, we now ask the
+user if they would like to upload this information to the LVFS.
+Using GNOME this would just be a slider in the control center privacy panel,
```
superm1: I think to most people who aren't familiar with the code this will be confusing.  Maybe just say that none of GUI tools currently support this feature.

docs/Telemetry.md
```diff
@@ -0,0 +1,113 @@
...
+in the LVFS web-console:
+
+![Firmware view showing the report](report-fw.png)
+
+Which expands out to the report below:
+
+![Report details](report-details.png)
+
+This means vendors using the LVFS know first of all how many downloads they
+have, and also the number of success and failures.
```
superm1: Should make it clear that number of downloads doesn't require the opt in.

docs/Telemetry.md
```diff
@@ -0,0 +1,113 @@
...
+This allows us to offer the same kind of staged deployment that Microsoft
+Update does, where you can limit the number of updated machines to 10,000/day
+or automatically pause the specific firmware deployment if > 1% of the reports
+come back with failures.
```
superm1: Should make it clear that these are optional behaviors configured by the OEM.

hughsie:
> Or you want to bubble it up to https://fwupd.org/vendors ?

I think we need to link to it on the fwupd page, but I don't want all this technical content in the main vendor page. It's scary enough for new vendors to get their heads around this stuff, without making them understand all the subtle nuances of little-used and optional parts of the process.

docs/Affiliates.md
hughsie: I've added a new section and an extra screenshot.

docs/CompositeFirmware.md
hughsie: Agreed, fixed, thanks.

docs/Telemetry.md
hughsie: Done, thanks.

docs/Telemetry.md
hughsie: Fixed, and added a screenshot.

superm1:
Looks good to me now

hughsie:
I did wonder about putting the images in 'img' to avoid cluttering the directory, i.e make it just the .md files in doc/ -- also, do you have any ideas where we should link to this content on the LVFS HTML pages?

superm1:
> make it just the .md files in doc/

Sure no strong opinions there for me.

> do you have any ideas where we should link to this content on the LVFS HTML pages?

I'm still of the opinion it's going to be confusing to come to https://fwupd.org/vendors and not get all the content you need there.  It's already jarring enough to new folks that some is based on tribal knowledge (or analyzing code in fwupd).  So my gut tells me it should be an "advanced topics" or FAQ at that page that links to all this stuff.

hughsie:
> bubble it up to https://fwupd.org/vendors ? I think it makes more sense

I've changed my mind, and I think you're completely right. We've got a website that people read, we don't want to point them at some other place to get the in-depth documentation. I'll respin this, and add a Documentation tab to the main LVFS website.

hughsie:
@superm1 -- can you have a look at https://staging.fwupd.org/ and tell me what you think please? There's now a new "Documentation" top level item with a few more things filled in.

PR #185, opened 2018-10-31
Clarify a statement about signing

hughsie:
It confused one ODM how fwupd was able to get around the SecureBoot signing
requirements.

### DIFF
--- a/app/templates/docs-users.html
+++ b/app/templates/docs-users.html
@@ -120,7 +120,13 @@ $ $ gdbus call --system --dest org.freedesktop.fwupd --object-path / --method or
       By default, any users are able to update signed firmware for removable hardware.
       The logic here is that if the hardware can be removed, it can easily be moved to
       a different device that the attacker owns.
-      For non-removable devices only admin users are able to update signed firmware.
+      For non-removable devices only user accounts with administrative rights
+      are able to update using firmware signed by the LVFS.
+    </p>
+    <p class="card-text">
+      Administrative users (i.e. <code>root</code>) can install firmware that
+      does not originate from the LVFS (e.g. locally built) but fwupd does
+      not affect the separate secure boot requirements in any way.
     </p>
   </div>
 </div>


app/templates/docs-users.html
```diff
@@ -120,7 +120,14 @@ <h3 class="card-header list-group-item-info">Do I always need the root password
...
       The logic here is that if the hardware can be removed, it can easily be moved to
       a different device that the attacker owns.
-      For non-removable devices only admin users are able to update signed firmware.
+      For non-removable devices only admin users are able to update using firmware
+      signed by the LVFS.
+    </p>
+    <p class="card-text">
+      Administrative users (i.e. <code>root</code>) can install firmware that
+      does not originate from the LVFS (e.g. locally built) but the hardware
+      signing key still has to be valid &mdash; fwupd does not affect the
```
superm1: This is actually true?  I seem to recall that daemon doesn't care about unsigned firmware currently.

app/templates/docs-users.html
```diff
@@ -120,7 +120,14 @@ <h3 class="card-header list-group-item-info">Do I always need the root password
       By default, any users are able to update signed firmware for removable hardware.
       The logic here is that if the hardware can be removed, it can easily be moved to
       a different device that the attacker owns.
-      For non-removable devices only admin users are able to update signed firmware.
+      For non-removable devices only admin users are able to update using firmware
```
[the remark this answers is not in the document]
hughsie: Agreed, thanks.

app/templates/docs-users.html
hughsie: In the original design documentation I wrote I think it said that firmware not from the LVFS would only be installable on non-removable hardware as root. Looking at the code `fu_main_install_with_helper()` loads the .cab file, and then we look at each component in turn. For each component with a match we call `fu_install_task_get_action_id()` which tests for `trust_flags & FWUPD_TRUST_FLAG_PAYLOAD` (i.e. signed by something I trust) which either means you get `org.freedesktop.fwupd.update-internal-trusted` (which is `yes` for an active user and `auth_admin` for SSH) and `org.freedesktop.fwupd.update-internal` (which is `auth_admin_keep` for active user and `auth_admin` for SSH).

Does this differ from how you interpreted the paragraph above?

superm1: I think the confusing part about that above paragraph is the "hardware signing key".  I think removing that would make it clearer.

superm1: If the question that prompted this change was around secure boot I think what is needed is to explain how fwupd fits into the secure boot picture (IE distro's shim is signed by a key recognized by the firmware and fwupd efi binary is signed by key recognized by shim)

hughsie: What about ".. can install firmware that does not originate from the LVFS () but fwupd does not affect the separate secure boot requirements in any way"

superm1: Yes that makes it much clearer.

PR #186, opened 2018-10-31
Use lxml to produce the XML remote metadata

hughsie:
Using appstream-glib was convenient, but behind the scenes it was merging
elements like releases in each components in ways that we didn't want.

It's also using a slegdehammer to crack a nut; so lets make things simpler.

Fixes: https://github.com/hughsie/fwupd/issues/818

### DIFF
--- a/app/metadata.py
+++ b/app/metadata.py
@@ -6,10 +6,7 @@
 
 import os
 import hashlib
-
-from gi.repository import AppStreamGlib
-from gi.repository import Gio
-from gi.repository import GLib
+from lxml import etree as ET
 
 from app import app, db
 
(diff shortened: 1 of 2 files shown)


app/metadata.py
```diff
@@ -18,141 +15,208 @@
...
+        component.set('type', 'firmware')
+        ET.SubElement(component, 'id').text = md.appstream_id
+        ET.SubElement(component, 'name').text = md.name
+        ET.SubElement(component, 'summary').text = md.summary
+        ET.SubElement(component, 'developer_name').text = md.developer_name
+        if md.description:
+            desc = ET.SubElement(component, 'description')
+            for c in ET.fromstring('<tmp>' + md.description + '</tmp>'):
+                desc.append(c)
+        ET.SubElement(component, 'project_license').text = md.project_license
```
superm1: Missing metadata license too right?

hughsie: Good question. I did originally have it there, but noticed that appstream-glib doesn't emit per-component metadata-license lines; IIRC because the *document* is a composite of the smaller components and had it's own metadata license that is a superset of all the components (which in this case is basically `CC0 or GFDL or Public Domain` and a couple more that are uninteresting.

superm1: OK, thanks!

superm1:
Nothing else stands out to me, thanks!

hughsie:
> Nothing else stands out to me

I've pushed this branch to the staging server and also increased the metadata regeneration to be every minute rather than every 20 minutes. Could you try with a composite firmware there please? Thanks.

superm1:
I uploaded two composite firmwares.  Trying to open the metadata in the browser I see this:
```
This page contains the following errors:
error on line 1 at column 1: Encoding error
Below is a rendering of the page up to the first error.
```
Not sure what's up with that though because xmllint seems to parse it just fine.

After gunzipping it, the part that we were concerned about initially seems to work properly now - multiple releases are generated when the same binary was needed.  At least immediately it looks right to me.

superm1:
I noticed when I downloaded the file it downloaded as `xml.gz.xml` and same thing happens on regular LVFS instance for embargo channel too.

hughsie:
I've fixed up a few things (custom->metadata) and also re-added the missing XML utf-8 header which you have to manually "turn on" for lxml. I'll merge this now, and deploy on Sunday. Thanks for the testing.

PR #188, opened 2018-11-05
Add a per-firmware is_dirty boolean flag

hughsie:
At the moment moving a firmware from testing to stable shows a message

    'The stable metadata will be signed in 16 minutes'

This tells the vendor that the firmware will be soon included in the metadata.

This *also* shows on any firmware currently in stable, which means vendors are
not sure if older firmwares are actually included at that moment. Now there are
lots of different vendors using the LVFS every day, it means it's frequently
unclear what firmware is actually in the metadata at the present time.

To fix this, include a per-firmware flag that is supposed to represent 'is this
firmware pending for the requested metadata for the selected target'.

When the firmware changes remote it is 'dirtied' and once the metadata has been
successfully regenerated and signed it is marked as clean. This means we can
show a more accurate message in the per-firmware details panel, and also not
show the message for unrelated firmware that happens to already be in the same
remote.

### DIFF
--- a/app/metadata.py
+++ b/app/metadata.py
@@ -242,6 +242,10 @@ def _metadata_update_targets(remotes):
                                 fws_filtered,
                                 firmware_baseuri=settings['firmware_baseuri'])
 
+        # all firmwares are contained in the correct metadata now
+        for fw in fws_filtered:
+            fw.is_dirty = False
+
 def _hashfile(afile, hasher, blocksize=65536):
     buf = afile.read(blocksize)
     while len(buf) > 0:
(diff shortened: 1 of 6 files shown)


superm1:
I think `views_delete` and `firmware_affiliation_change` needs to have the FW marked dirty too.

hughsie:
> I think views_delete and firmware_affiliation_change needs to have the FW marked dirty too.

Agreed. Fixed.

hughsie:
I'll deploy tomorrow when it's quiet. Thanks for the review!

PR #190, opened 2018-11-08
Wip/hughsie/better content

hughsie:


app/templates/firmware-problems.html
```diff
@@ -84,7 +107,20 @@
...
+      Unknown problem <code>{{problem.kind}}</code>!
+    </td>
+    <td class="col-sm-9">
+      <p class="card-text">
+        Please report this issue to the LVFS administrator.
+      </p>
+    </td>
+  </tr>
 {% endif %}
+{% endfor %}
```
superm1: the number of for's and if's are getting a little unweidly.  Maybe add a comment ad the `endfor` and `endif` blocks to make it more obvious what the block is ending.

hughsie: I think I need to sit down and work out how to do Flask templates properly. I'm pondering redoing the problems page, as I agree it's all quite inelegant.

PR #191, opened 2018-11-08
Add some documentation about GUID generation

hughsie:


app/templates/docs-metainfo.html
```diff
@@ -61,6 +61,53 @@ <h1>Introduction</h1>
...
+<p>
+  GUID means 'Globally Unique Identifier' and is a 128-bit integer number used
+  to identify a device.
+  GUIDs are often formatted as strings such as <code>84f40464-9272-4ef7-9399-cd95f12da696</code>.
+  Another name for GUID is UUID ('Universally Unique Identifier') and the two
+  terms can be used interchangably.
+</p>
+<p>
+  In fwupd the GUID is generated from the <code>DeviceID</code> strings, so for
+  a single USB device the GUIDs would be generated like this:
```
superm1: This particular wording is going to cause confusion.  `DeviceID` has a specific meaning now in fwupd (based on the changes for physical ID and logical ID being used to "build" a device ID).

You should be using the term `DeviceInstanceId` instead I think.

app/templates/docs-metainfo.html
```diff
@@ -61,6 +61,53 @@ <h1>Introduction</h1>
...
+</p>
+<pre>
+$ appstream-util generate-guid "USB\VID_0A5C&PID_6412&REV_0001"
+52fd36dc-5904-5936-b114-d98e9d410b25
+$ appstream-util generate-guid "USB\VID_0A5C&PID_6412"
+7a1ba7b9-6bcd-54a4-8a36-d60cc5ee935c
+$ appstream-util generate-guid "USB\VID_0A5C"
+ddfc8e56-df0d-582e-af12-c7fa171233dc
+</pre>
```
superm1: I think it would be better to show python example as well (and first).  Most build systems won't have appstream-util available, and this is most valuable when integrating into a build system.

app/templates/docs-metainfo.html
```diff
@@ -61,6 +61,53 @@ <h1>Introduction</h1>
...
+<p>
+  In the case for PCI devices and other technologies like NVMe, you can dump
+  the GUIDs generated by fwupd using this tool:
+</p>
+<pre>
+sudo /usr/libexec/fwupd/fwupdtool --plugin-whitelist nvme get-devices --verbose
+...
+using e22c4520-43dc-5bb3-8245-5787fead9b63 for NVME\VEN_1179&amp;DEV_010F&amp;REV_01
+using 83991323-9951-5adf-b743-d93e882a41e1 for NVME\VEN_1179&amp;DEV_010F
+using ad9fe8f7-cdc4-52c9-9fea-31b6f4988ffa for NVME\VEN_1179
```
superm1: I wonder if this is indicative that we should have a "standard" documentation table in README.md for each of the plugins showing how they create and use GUIDs.  Then this documentation can link directly back to fwupd saying look at how your plugin you're interested in generates GUIDs.

Alternatively a master table in project wide fwupd documentation or LVFS documentation would suffice for this purpose.  I just think doing it in plugin specific README.md will be more likely to remember to get updated each time a new plugin is created or a new use of GUIDs is introduced to existing plugins.

app/templates/docs-metainfo.html
```diff
@@ -61,6 +61,53 @@ <h1>Introduction</h1>
...
+...
+</pre>
+<p>
+  Firmwares do not have to prodide just <em>one</em> GUID.
```
superm1: typo

app/templates/docs-metainfo.html
hughsie: Agreed, thanks.

app/templates/docs-metainfo.html
hughsie: Do we have a python example in tree?

superm1: Yeah two places:
https://github.com/hughsie/fwupd/blob/2a172e8dd2a9ccb7662be60c6a355f5286eac054/src/fu-common-guid.c#L120

https://github.com/hughsie/fwupd/blob/04d6965a146acbdbd5653c56f895bc30015905fb/plugins/thunderbolt/README.md

hughsie: Thanks, I've added the python example first:

app/templates/docs-metainfo.html
hughsie: I think that makes sense. I've opened https://github.com/hughsie/fwupd/issues/837 -- I can work on this and drop the lvfs-website bits if you think that makes sense.

superm1: Sounds good to me.

hughsie: I've added a note about the new README content.

PR #194, opened 2018-11-15
Make the action more obvious when managing vendor users

hughsie:
On the LVFS we usually put the 'action widget' as a button on the right hand
side of a table. In the per-vendor users view the common action was actually
going to the details, which meant you had to know you could click on the email
address. The uncommon action was a big red scary 'Disable' button that made it
look even more like this was the only way to interact with the table row.

Use a less scary 'Details' button like the rest of the LVFS, and remove the
one-click-disable misfeature.

### DIFF
--- a/app/templates/vendor-users.html
+++ b/app/templates/vendor-users.html
@@ -15,9 +15,7 @@
   </tr>
 {% for u in v.users %}
   <tr class="row">
-    <td class="col-sm-3">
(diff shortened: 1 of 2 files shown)


superm1:
I think you still want to be able to disable the user, I don't remember off hand is that on the details page?

If so then that seems like a good change to me.

hughsie:
> is that on the details page

Yes, you can select "disabled" on the combo dropdown.

superm1:
Lgtm

PR #196, opened 2018-11-21
Strip leading and trailing whitespace before converting to XML

hughsie:


hughsie:
> It seems that they would just be ignored now.

I think they were ignored before, no?

superm1:
Possibly, just comes to mind as something to consider

Issue #117, opened 2018-05-31
Should prompt to change password on initial login


hughsie:
We create a random password for new users, and ask them to change it on first log in. It would be better if on first logging in to the LVFS they are taken straight to the profile page rather than to the index, along with some text about how to choose a good password.

PR #204, opened 2018-12-06
Prompt the user to change thier password when required

hughsie:
This will be used on initial login from account creation, and also if they've
used the password reset feature.

Fixes https://github.com/hughsie/lvfs-website/issues/117

### DIFF
--- a/app/views_user.py
+++ b/app/views_user.py
@@ -76,8 +76,13 @@ def user_modify(user_id):
         flash('Failed to modify profile: Name invalid', 'warning')
         return redirect(url_for('.profile'), 302)
 
+    # hash the provided password
+    password_hash = _password_hash(password)
+    if password_hash != user.password:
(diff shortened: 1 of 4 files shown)


app/views_user.py
```diff
@@ -70,14 +70,19 @@ def user_modify(user_id):
     if not _password_check(password):
         return redirect(url_for('.profile'), 302)
 
+    # hash the provided password
+    password_hash = _password_hash(password)
+    if password_hash != user.password:
+        user.password = password_hash
+        user.password_ts = datetime.datetime.utcnow()
```
superm1: You'll probably want to move this to after the name verification, no?  Otherwise there is a chance that the password change timestamp gets recorded if the display name was set invalid.

hughsie: Good catch.

PR #206, opened 2018-12-07
Allow vendors to disallow firmware downloads from specific countries

hughsie:
This is using the free GeoIP data and is optional, and best-effort.

To test, you need to do: yum install geolite2-country python-GeoIP

### DIFF
--- a/app/templates/docs-vendors.html
+++ b/app/templates/docs-vendors.html
@@ -65,4 +65,19 @@
   </div>
 </div>
 
+<div class="card mb-3">
+  <h3 class="card-header list-group-item-info">Can I restrict downloads from embargoed and sanctioned countries?</h3>
+  <div class="card-body">
+    <p class="card-text">
+      The LVFS administrator can configure the policy for all firmware owned by
+      the vendor to be blocked from download in embargoed or otherwise
+      sanctioned countries.
+    </p>
+    <p class="card-text">
+      The blocked ISO 3166 country codes can also be specified in the firmware
+      itself, using the <code>LVFS::BannedCountryCodes</code> metadata key.
+    </p>
+  </div>
+</div>
+
 {% endblock %}
--- a/app/templates/vendor-details.html
+++ b/app/templates/vendor-details.html
@@ -46,6 +46,12 @@
            value="{{ v.username_glob if v.username_glob }}">
   </div>
   <div class="form-group">
+    <label for="display_name">Banned countries:</label>
+    <input type="text" class="form-control" name="banned_country_codes"
+           placeholder="ISO 3166 country codes, e.g. KP,CU,IR,SU,SY"
+           value="{{ v.banned_country_codes if v.banned_country_codes }}">
+  </div>
+  <div class="form-group">
     <label for="email">Description:</label>
     <input type="text" class="form-control" name="description" value="{{v.description}}" required>
   </div>
--- a/app/models.py
+++ b/app/models.py
@@ -339,6 +339,7 @@ class Vendor(db.Model):
     username_glob = Column(Text, default=None)
     version_format = Column(String(10), default=None) # usually 'triplet' or 'quad'
     url = Column(Text, default=None)
+    banned_country_codes = Column(Text, default=None) # ISO 3166, delimeter ','
 
     # magically get the users in this vendor group
     users = relationship("User",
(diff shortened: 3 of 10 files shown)


app/templates/docs-vendors.html
```diff
@@ -65,4 +65,15 @@ <h3 class="card-header list-group-item-info">Why does the LVFS project sign the
...
 </div>
 
+<div class="card mb-3">
+  <h3 class="card-header list-group-item-info">Can I restrict downloads from embargoed and sanctioned countries?</h3>
+  <div class="card-body">
+    <p class="card-text">
+      The list of embargoed and sanctioned countries can be provided to the
+      LVFS administrator in confidence and set as policy for all firmwares
+      in your LVFS vendor group.
+    </p>
```
hughsie: This certainly sounds inelegant -- ideas welcome.

app/templates/vendor-details.html
```diff
@@ -41,6 +41,12 @@
     <input type="text" class="form-control" name="username_glob" placeholder="*@domain.com"
            value="{{ v.username_glob if v.username_glob }}">
   </div>
+  <div class="form-group">
+    <label for="display_name">Banned countries:</label>
+    <input type="text" class="form-control" name="banned_country_codes"
+           placeholder="ISO 3166 country codes, e.g. KP,CU,IR,SU,SY"
```
superm1: This should be mentioned it is comma delimited ISO 3166 country codes

superm1:
This PR is getting into murky waters.
Is sending the metadata OK in these instances though?  I sorta feel like if you're going to be blocking it should be blocking everything and for all countries embargoes/sanctioned based on where LVFS is hosted as a basis.

app/templates/docs-vendors.html
superm1: The LVFS administrator can configure the policy for all firmware owned by an LVFS vendor to be blocked from download in embargoes or sanctioned countries.

hughsie:
> This PR is getting into murky waters.

Totally agree. It's a request from a moderately large OEM as a condition for joining the LVFS. I'm not sure how hard I should push back, if at all.

> Is sending the metadata OK in these instances though?

Yes, I've asked for clarification, but it seems the binary code (capsule) that contains executable code that can't be distributed due to various legal sanctions. I've got no problem redistributing ColorHug firmware to Sudan (etc.), but I can understand why the OEM might be a little more hesitant.

> based on where LVFS is hosted as a basis

Not sure I follow -- can you expand?

superm1:
> but it seems the binary code (capsule) that contains executable code that can't be distributed due to various legal sanctions

This sounds like pandora's box.  It does make me wonder if we've all been turning a blind eye because plenty of firmware images do carry encryption algorithms.

> Not sure I follow -- can you expand?

Really it's LVFS distributing not the OEM distributing it isn't it?  So doesn't the onus of legal requirements on distribution for encryption (etc) fall on the locale that LVFS is hosted?  I might be entirely wrong here, it's probably worth passing this to RH or LF legal to decide what to do.

hughsie:
> This sounds like pandora's box.

I agree.

> It does make me wonder if we've all been turning a blind eye because plenty of firmware images do carry encryption algorithms.

I think all the "upstream" vendor websites that allow you to download firmware only ask you to read an agreement or EULA which states you can't download the file is you live in Cuba, Iran, North Korea, Sudan or Syria for example. I can't believe all of them actually are doing ip-block selection for this -- and it's a trust exercise on the end user. It's tricky also to test this, as all the VPN services for a place like Cuba are in the US so people from Cuba can escape the IP blacklisting, rather than _in_ Cuba to actually _test_ the IP blacklisting...

> it's probably worth passing this to RH or LF legal to decide what to do.

I've sent email to the LF legal team. This is the first time this issue has even been mentioned, let alone discussed.

app/templates/vendor-details.html
hughsie: I think it's only me that can see that texbox, and I've put commas in the placeholder text. :)

app/templates/docs-vendors.html
hughsie: Thanks.

superm1:
Well unless you want to (potentially) back pedal if legal says not something to put it there, put it in a TOS somewhere my opinion is wait.

hughsie:
> my opinion is wait

Right, agree. Thanks for the sanity check.

hughsie:
> my opinion is wait

**Unofficial** feedback from legal, not legal advice, yada yada yada: Firmware updates distributed without a fee on hardware that is not specifically designed and built for military use is okay to distribute, even to embargoed countries. If we want to allow filtering by country code that's perfectly fine, but if the hardware is built for military use then it's required.

PR #208, opened 2018-12-12
Allow specifying a device checksum in the metainfo.xml file

hughsie:
Some firmware has a different on-device checksum to the hash of the firmware
file itself. This may be because:

 * The content is not a binary file, e.g. Intel HEX or SREC
 * Only part of the firmware is flashed, e.g. ignoring the bootloader section
 * The device checksum is calculated using another method entirely, e.g. PCR0

### DIFF
--- a/app/metadata.py
+++ b/app/metadata.py
@@ -160,6 +160,13 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri=''):
                 csum.set('filename', md.filename_contents)
                 csum.set('target', 'content')
 
+            # add device checksum
+            if md.checksum_device:
+                csum = ET.SubElement(rel, 'checksum')
+                csum.text = md.checksum_device
+                csum.set('type', 'sha1')
+                csum.set('target', 'device')
+
             # add long description
             if md.release_description:
                 rel.append(_xml_from_markdown(md.release_description))
--- a/app/views_upload.py
+++ b/app/views_upload.py
@@ -133,6 +133,12 @@ def _create_fw_from_uploaded_file(ufile):
         md.checksum_contents = csum.get_value()
         md.filename_contents = csum.get_filename()
 
+        # from the device checksum
+        if hasattr(AppStreamGlib.ChecksumTarget, 'DEVICE'):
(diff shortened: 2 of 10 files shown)


app/metadata.py
```diff
@@ -160,6 +160,13 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri=''):
                 csum.set('filename', md.filename_contents)
                 csum.set('target', 'content')
 
+            # add device checksum
+            if md.checksum_device:
+                csum = ET.SubElement(rel, 'checksum')
+                csum.text = md.checksum_device
+                csum.set('type', 'sha1')
```
superm1: sha1? Since this is new I feel it should be sha256

app/views_upload.py
```diff
@@ -133,6 +133,12 @@ def _create_fw_from_uploaded_file(ufile):
         md.checksum_contents = csum.get_value()
         md.filename_contents = csum.get_filename()
 
+        # from the device checksum
+        if hasattr(AppStreamGlib.ChecksumTarget, 'DEVICE'):
```
superm1: Weren't you trying to remove the need for appstream glib in lvfs-website?

app/metadata.py
hughsie: We should really move to a better hash format at some point. We do need to support /both/ as some hardware will only spit out SHA1 for the PCR0 case. I think supporting "both" means "any", so it's quite a bigger change to rearchitect the databases to split out the checksums to a different table and to add a proper `Checksum` object in `models.py`. Certainly needs doing, but not today.

app/views_upload.py
hughsie: I think here it makes sense still in `uploadedfile.py`; we can re-use a lot of the validation stuff from the library without re-implementing all the rules in python.

app/metadata.py
superm1: As long as from the OEM side building the metainfo we can just fill in both now without LVFS getting upset pushing that back is fine.

hughsie:
Looks good now? It's getting complicated juggling PRs :)

PR #209, opened 2018-12-12
Allow setting the update protocol on each firmware

hughsie:
These can either be set on the web console or by including this extra data to
the metadata file:

    <custom>
      <value key="LVFS::UpdateProtocol">uefi</value>
    </custom>

At the moment the LVFS doesn't actually know what kind of protocol the firmware
blob is going to use, which makes things like automatic checking at upload time
impossible.
This extra data would allow us to detect a missing UEFI UpdateCapsule header at
upload time, rather than at deployment time.

Having this information also hints about the relative safety of doing mechanism,
for instance unsigned DFU being much worse than UpdateCapsule or CSR.


superm1:
I would say it's better to use `LVFS::UpdatePlugin` since realistically this is going to map out to fwupd plugins.

hughsie:
Doesn't this mean that the fwupd plugin name becomes somewhat like ABI? I'm not saying this is a bad thing necessarily, but it does mean we have to put some compat hooks in place if we ever rename a plugin like we did with `wacomhid`->`wacom-usb`. I suppose it also lets us do other clever things in the future if we decide this is a good idea e.g. restricting the firmware to a specific plugin in fwupd. It also means we don't have to debate the protocol values in the LVFS, e.g. `uefi` or `UpdateCapsule`

superm1:
Yes it does mean that fwupd plugin becomes an ABI, that's true.  And would need to be careful for future plugin renames.

I think those two reasons you mentioned are great reasons to do this.  However I'll note there are some instances that the same firmware applies in two plugins, so this situation doesn't work there.

What we're talking about is really binary formats they can use so maybe this slight modification:
1. Plugins automatically advertise the name of themselves as the plugin name for the purpose of this PR.
2. Plugins can opt into supporting other formats during startup.  So `dell-dock` can advertise that it also supports `synapticsmst` and `wacom-usb` can also advertise it supports `wacomhid`.  That gets you the compatibility hooks and handles the rename scenario effectively.

hughsie:
> that the same firmware applies in two plugins

This is per-component, not per-firmware -- is that still true?

> wacom-usb can also advertise it supports wacomhid

Something like:

    fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_ALIAS_FOR_PLUGIN, "wacomhid");

My worry here is we're blurring the lines between the protocol itself and the plugin name. I'd maybe suggest we should have a dell-dock `Protocol` version if that's the plugin that's going to be applying the SyanpticsMST firmware. Although it's SyanpticsMST firmware being applied to a SyanpticsMST device, you can't actuall use the `synapticsmst` plugin for dock hardware, right? If you can, then we need to do the `ALIAS_FOR_PLUGIN` feature before we limit the firmwares to a specific plugin.

superm1:
> fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_ALIAS_FOR_PLUGIN, "wacomhid");

Yeah

> Although it's SyanpticsMST firmware being applied to a SyanpticsMST device, you can't actuall use the synapticsmst plugin for dock hardware, right? 

You can use `synaptics-mst` to install the update as well, the priority is just set for `dell-dock` if it's compiled in.  

>  If you can, then we need to do the ALIAS_FOR_PLUGIN feature before we limit the firmwares to a specific plugin

So yeah I think we would need to have plugin aliases sorted first.

New name proposal: `LVFS::FwupdPlugin`

superm1:
> This is per-component, not per-firmware -- is that still true?

Missed this one.  Yes this needs to be per-component.  At least with docks it can be flashed through 2-3 plugins in concert.

superm1:
One thought that came to mind after your reverse dns idea for fwupd:

How about using the identifier as a hint to set protocol in LVFS initially?

IE if it started with org.uefi then LVFS picks the org.uefi.UpdateCapsule?

hughsie:
> if it started with org.uefi

The component ID is under the OEM control, it's supposed to indicate where the firmware is from, rather than how it's being deployed. We could certainly have some rules (hardcoded if required) like we have in the migration script:

        if md.appstream_id.startswith('com.dell.uefi'):
            md.protocol_id = proto_id_for_value['org.uefi.UpdateCapsule']

superm1:
I do think rules like that would be useful.  Especially since it will take time to get a new key added into building scripts running in production.

app/templates/docs-metainfo.html
```diff
@@ -180,6 +180,43 @@ <h2>Restricting direct downloads</h2>
   &lt;/custom&gt;
 </pre>
 
+<hr/>
+<h2><a id="upload-protocol"></a>Settting the Upload Protocol</h2>
+<p>
+  The LVFS now needs to know what protocol is being used to flash the firmware
```
superm1: There isn't really a concept of time in documentation right now

app/templates/firmware-md-overview.html
```diff
@@ -41,12 +41,33 @@
...
+{% endif %}
+  </tr>
+  </form>
   <form class="form-inline" method="post" action="/lvfs/component/{{md.component_id}}/modify" method="POST">
   <tr class="row">
     <th class="col-sm-3">Device Checksum</th>
 {% if md.fw.remote.name != 'stable' or g.user.check_acl('@admin') %}
     <td class="col-sm-7">
       <input type="text" class="form-control fixed-width" name="checksum_device" value="{{md.checksum_device if md.checksum_device}}"/>
+{% if md.protocol and md.protocol.value == 'org.uefi.UpdateCapsule' %}
```
superm1: should be adjusted to lower case now

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+            md.protocol_id = proto_id_for_value['com.qualcomm.DFU']
+        elif md.appstream_id.startswith('com.acme.'):
+            md.protocol_id = proto_id_for_value['org.usb.DFU']
+        elif md.appstream_id.startswith('fakedevice'):
+            md.protocol_id = proto_id_for_value['org.usb.DFU']
+        elif md.appstream_id.startswith('com.jabra.'):
+            md.protocol_id = proto_id_for_value['org.usb.DFU']
+        elif md.appstream_id.startswith('com.dell.mst'):
+            md.protocol_id = proto_id_for_value['com.synaptics.MST']
+        elif md.appstream_id.startswith('com.dell.salomon'):
```
superm1: please drop this one. just don't set protocol for it.

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+                                name='Wacom (USB devices)',
+                                is_signed=False) # FIXME?
+        db.session.commit()
+
+    # find the IDs for each value
+    proto_id_for_value = {}
+    for pr in db.session.query(Protocol).all():
+        proto_id_for_value[pr.protocol_value] = pr.protocol_id
+
+    # convert the existing components
```
superm1: Case needs to be adjusted for all of these.

I think ideally this code should be moved into another function that can be imported here so that you can make the rules effective (at least for a while) to automatically set these protocol IDs until building tools get updated.

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+            md.protocol_id = proto_id_for_value['org.usb.DFU']
+        elif md.appstream_id.startswith('fakedevice'):
+            md.protocol_id = proto_id_for_value['org.usb.DFU']
+        elif md.appstream_id.startswith('com.jabra.'):
+            md.protocol_id = proto_id_for_value['org.usb.DFU']
+        elif md.appstream_id.startswith('com.dell.mst'):
+            md.protocol_id = proto_id_for_value['com.synaptics.MST']
+        elif md.appstream_id.startswith('com.dell.salomon'):
+            md.protocol_id = proto_id_for_value['com.dell.Dock']
+        elif md.appstream_id.startswith('com.dell.wd'):
```
superm1: likewise for this one, you can just drop it.

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+            md.protocol_id = proto_id_for_value['unknown']
+        elif md.appstream_id.startswith('com.tw.supermicro.'):
+            md.protocol_id = proto_id_for_value['org.uefi.UpdateCapsule']
+        else:
+            print('unknown protocol for', md.appstream_id)
```
superm1: In this case don't you also want to set
```
md.protocol_id = proto_id_for_value['unknown']
```

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+                                is_signed=False)
+        db.session.add(Protocol(value='org.nvmexpress',
+                                name='NVMe',
+                                is_signed=False) # FIXME?
+        db.session.add(Protocol(value='org.dmtf.Redfish',
+                                name='Redfish',
+                                is_signed=True)
+        db.session.add(Protocol(value='com.realtek.RTS54',
+                                name='Realtek RTS54',
+                                is_signed=False)
```
superm1: I expect this to be signed

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+        db.session.add(Protocol(value='com.acme.Test',
+                                name='Test protocol DO NOT USE',
+                                is_signed=False)
```
superm1: It might be worth special casing this one to prevent it getting out to the field.  if something has this, throw up the problems flag and don't let it migrate to stable.

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+        db.session.add(Protocol(value='com.intel.Thunderbolt',
+                                name='Intel Thunderbolt',
+                                is_signed=False) # FIXME?
```
superm1: In production hardware this is always signed.

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+        db.session.add(Protocol(value='org.uefi.UpdateCapsule',
+                                name='UEFI UpdateCapsule',
+                                is_signed=True)
```
superm1: Generally true - but it's actually not mandatory that it's signed.

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+                                is_signed=False)
+        db.session.add(Protocol(value='org.altusmetrum.AltOS',
+                                name='AltOS Update',
+                                is_signed=False)
+        db.session.add(Protocol(value='com.qualcomm.DFU',
+                                name='Qualcomm (Cambridge Silicon Radio) DFU',
+                                is_signed=True)
+        db.session.add(Protocol(value='com.dell.Dock',
+                                name='Dell Salomon Dock',
+                                is_signed=False) # FIXME?
```
superm1: correct :(

superm1: Should be True

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+        db.session.add(Protocol(value='com.synaptics.MST',
+                                name='Synaptics MST',
+                                is_signed=False) # FIXME?
```
superm1: correct :(

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,162 @@
...
+                                name='Hughski ColorHug',
+                                is_signed=False)
+        db.session.add(Protocol(value='org.altusmetrum.AltOS',
+                                name='AltOS Update',
+                                is_signed=False)
+        db.session.add(Protocol(value='com.qualcomm.DFU',
+                                name='Qualcomm (Cambridge Silicon Radio) DFU',
+                                is_signed=True)
+        db.session.add(Protocol(value='com.dell.Dock',
+                                name='Dell Salomon Dock',
```
superm1: "Dell dock" please

app/models.py
```diff
@@ -1280,6 +1296,48 @@ def __init__(self, value, datestr=0, cnt=1):
...
+class Protocol(db.Model):
+
+    # sqlalchemy metadata
+    __tablename__ = 'protocol'
+    __table_args__ = {'mysql_character_set': 'utf8mb4'}
+
+    protocol_id = Column(Integer, primary_key=True, nullable=False, unique=True)
+    value = Column(Text, nullable=False)
+    name = Column(Text, default=None)
+    is_signed = Column(Boolean, default=False)
```
superm1: Some of these protocols whether they are signed is implementation specific.  For example UEFI capsules don't need to have signatures, it's up to an OEM to decide if they want to do that.

So it almost feels like this should be a trilean field of `True`, `False`, `Unknown`

migrations/versions/6dde8aa078e5_add_protocol_table.py
hughsie: Case adjusted. I didn't move the importer, as these are only mostly true, a few manual fixups are required. I also don't want to suggest there's a link between the `<id>` and the protocol going forward, it's really only something I'm using for this import as a hacky heuristic.

migrations/versions/6dde8aa078e5_add_protocol_table.py
hughsie: Fixed, good catch.

migrations/versions/6dde8aa078e5_add_protocol_table.py
hughsie: What about if we hide this from the dropdown unless you're the admin user?

app/models.py
hughsie: Can you create an unsigned capsule? I understand you can turn the signing requirement "off" on the hardware, but that's not something we can capture on the LVFS or fwupd even.

migrations/versions/6dde8aa078e5_add_protocol_table.py
superm1: What happens though if someone puts it in `LVFS::UpdateProtocol`?  Although I guess you have to be really determined to do that so hiding for admin only is probably good enough.

hughsie: Valid point about `LVFS::UpdateProtocol` -- I'll add the problem thing you suggested initially.

hughsie:
Some screenshots of the new stuff:

screenshot from 2018-12-13 14-07-59
screenshot from 2018-12-13 14-08-31
screenshot from 2018-12-13 14-06-06
screenshot from 2018-12-14 15-41-17

superm1:
Personal preference; I don't like the badges.  I feel that they don't convey anything (you have to know what colors mean what, what shape means what)

app/dbutils.py
```diff
@@ -136,7 +136,9 @@ def anonymize_db(db):
             continue
         for md in fw.mds:
             md.checksum_contents = hashlib.sha1(os.urandom(32)).hexdigest()
-            md.checksum_device = hashlib.sha1(os.urandom(32)).hexdigest()
+            for csum in md.device_checksums:
+                csum.kind = 'SHA1'
```
superm1: I'd think you don't want to hardcode this here do you since you can fill the metadata with other types of checksums.

app/models.py
superm1: Well let me put this  this way.  I expect that you can go append garbage to a Dell UEFI capsule and as long as the header at the front of the capsule reflects the proper size the `UpdateCapsule()` routine should accept it (and actually work).

The _capsule itself_ is not the signed part.  The signed part is "within" the capsule and evaluated later on after it's scanned and extracted.

So by the above comments there is no requirement whatsoever that the capsule itself have anything signed.

app/templates/docs-metainfo.html
```diff
@@ -194,6 +194,43 @@ <h2><a id="custom-direct-downloads"></a>Restricting direct downloads</h2>
   &lt;/custom&gt;
 </pre>
 
+<hr/>
+<h2><a id="custom-update-protocol"></a>Settting the Update Protocol</h2>
+<p>
+  The LVFS needs to know what protocol is being used to flash the device.
+  The protocol value is used to provide information about the security of the
+  firmware update to end users.
+</p>
```
superm1: Is it worth including that fwupd will also use this information too in some situations?

migrations/versions/6dde8aa078e5_add_protocol_table.py
```diff
@@ -0,0 +1,180 @@
...
+                                can_verify=False,
+                                is_signed=False))
+        db.session.add(Protocol(value='com.qualcomm.dfu',
+                                name='Qualcomm (Cambridge Silicon Radio) DFU',
+                                can_verify=False,
+                                is_signed=True))
+        db.session.add(Protocol(value='com.dell.dock',
+                                name='Dell Dock',
+                                can_verify=False,
+                                is_signed=False)) # FIXME?
```
superm1: This should be set to `True`

app/dbutils.py
hughsie: Well, this is just antonymous data. It could be MD5, I don't think it's matters.

app/templates/docs-metainfo.html
hughsie: Yes, when we add the fwupd bit. There are quite a few edge cases which need thinking about.

hughsie:
> Personal preference; I don't like the badges

I just used the default icons for the various security states in GNOME. Other ideas welcome -- I do think we need a concise way to show the user the secure from the insecure, although I'm not wedded to the shields.

hughsie:
Another idea:

screenshot from 2018-12-18 14-00-51

hughsie:
Or maybe we just don't use any kind of image, e.g.:

screenshot from 2018-12-18 14-11-14

hughsie:
I talked a bit with a co-worker, and dissuaded myself from the icons:

    <hughsie> the general idea is we try to get vendors to get all green so they're more secure
    <hughsie> and the problem with shields
    <hughsie> is we either add a new level for the next best practice thing we ask for, e.g. "crested shield with parrot on top"
    <hughsie> or we demote people with an existing silver shield to a wooden shield
    <hughsie>which might cause problems in some marketing departments...

I've switched the PR to use no icons, and just show the red/green checkbox thing.

hughsie:
I've had one valid comment from a person worried about the implication that the signed requirement is always a good thing, and that we should display hardware that allows "support for other signing keys" -- for instance enrolling your keys into the hardware so that you could write or build your own firmware. I guess for UEFI devices this is typical, but not guaranteed, for unifying devices this is impossible by design. For devices like ColorHug this is possible by opening the device and powering on with a jumper set, and for a Chromebook it takes removing a screw. Ideas?

superm1:
I think that's starting to conflate different security mechanisms.  

When you enroll certificates into the hardware that's referring to loading UEFI binaries signed with the associated private keys for the purposes of secure boot.  It has no implication on the signature of the UEFI capsule payload used for updates.  I think we should be very careful that we refer only to that to avoid extra confusion. 

>  so that you could write or build your own firmware

The only systems I know of that will allow this are chromebooks.

If you're going to start talking about hardware modifications such as opening a case, removing a screw or similar to allow you to flash "unsigned" payloads you run into the situations that you can probably remove the SPI, take something like dediprog to flash it and return it.  It's much more difficult to defend against this type of "physical" attack.

> Ideas?

I think stick to the stuff that fwupd and LVFS does and can potentially verify - not the security architecture of the system or contents of the payload.  If the payload is signed and the firmware on the chip it's being flashed to can verify it, indicate that.  If it's not, indicate that.  Anything beyond that is going down a rabbit hole.

hughsie:
> you can probably remove the SPI, take something like dediprog to flash

Fair point...

> Anything beyond that is going down a rabbit hole.

Yes, I think you're right. Does the patch as-is look acceptable?

app/templates/docs-metainfo-protocol.html
```diff
@@ -42,4 +42,44 @@
...
+  This is stored in the TPM chip as <code>PCR0</code>.
+</p>
+<p>
+  To list the various <code>PCRs</code> on the running system you can use
+  <code>cat /sys/class/tpm/tpm0/pcrs</code> for TPMs using protocol 1.2, or
+  <code>tpm2_listpcrs</code> for TPMs using later protocol versions.
+  The PCR0 can be included in the vendor-supplied <code>firmware.metainfo.xml</code>:
+</p>
+<pre>
+    &lt;releases&gt;
```
superm1: for best practice as this may lead to templates and script I think you should include an example with both sha256 and sha1.

superm1:
Just one minor nit in your template, but otherwise I think it's fine now.

app/templates/docs-metainfo-protocol.html
hughsie: Fixed, good idea.

PR #212, opened 2018-12-14
Add device name to page title

wiktor-k:
This change adds currently displayed device name to the page title. Having device name there makes it easy to look for a previously visited device in browser history. This may also have a positive impact on search engine positioning.

I was considering an alternative where the most specific string is on the left (e.g. "XPS 13 9350 System Update - Device - LVFS") for when a page title is displayed in a shrunk page tab.

To be honest I didn't test this change but I reused the placeholder from device name so I hope it works :)

Thank you for taking the time to review this PR!

### DIFF
--- a/app/templates/device.html
+++ b/app/templates/device.html
@@ -1,5 +1,5 @@
 {% extends "default.html" %}
-{% block title %}Device{% endblock %}
+{% block title %}{{fws[0].mds[0].name if fws else 'No firmware for device found'}}{% endblock %}
 
 {% block nav %}{% include 'device-nav.html' %}{% endblock %}
 


hughsie:
I think you need a `if fws|length == 0` in there -- fws doesn't always have to have an children. In the case it's not supported, I guess a better title than `Device` makes sense too. Thanks!

wiktor-k:
Ah yes, I see the travis failure and the docker scripts to test it now. Give me a moment to fix this up!

wiktor-k:
Okay, passes sanity checks, sorry for the initial failure! If you have a suggestion for better titles I'm happy to hear it :)

app/templates/device.html
```diff
@@ -1,5 +1,5 @@
 {% extends "default.html" %}
-{% block title %}Device{% endblock %}
+{% block title %}{% if fws|length == 0 %}No firmware for device found{% else %}Device: {{fws[0].mds[0].name}}{% endif %}{% endblock %}
```
hughsie: I'm really not an expert at templates, but can do do something like??:

    {{fws[0].mds[0].name if fws else 'No firmware for device found'}}

wiktor-k: Looks a lot cleaner! :+1:

Issue #221, opened 2019-01-03
Show more complicated component requirements


hughsie:
At the moment there is no way to add a GUID firmware requirement to an existing component, or to show the requirement type in the admin console. There is also no way of specifying a range version requirement, for instance `>= 1.0.0` and `<= 1.2.0` -- both are not easily solved using the "simple" requirement panel.

superm1:
One other thought that comes to mind with this issue is how do you prevent someone from accidentally locking out updates from a majority of versions?  `>= 1.1.4` and `<= 1.2.0` would mean that only the `1.1.5` and later releases of `1_1_X `worked and then all the `master` `1.2.x` releases after `1.2.0` wouldn't work. 

Maybe this is a little far fetched, but bear with me.  Similarly what if it's a bug that existed in fwupd `1.2.1` and fwupd `1.1.4`.  it was patched in fwupd `1.2.2` and that patch backported to `1.1.5`.  if you set `>= 1.1.5 `you might run into problems with it in `1.2.1` and `1.2.0` still.

It's almost like you need to indicate which branch it's tracking at that point. to handle complex requirements?

hughsie:
> is how do you prevent someone from accidentally locking out updates 

I don't think we can stop people doing this, and it might be exactly what they want to do.

> It's almost like you need to indicate which branch it's tracking at that point

I guess we could define `org.freedesktop.fwupd.branch-1_0_X` and depend on that, but it seems a bit fragile and magical. I think my advice there would just to depend on the newest released version of fwupd that has the bugfix, e.g. `>=1.2.3` otherwise we're getting into the complicated area of tracking each behavior change or bugfix -- e.g. depending on commit 831eb7ef7e60c91485a3953267cbfb416018a8d3 being included. That's not even enough, as the git commits might be different SHA1's... Distros are also free to revert or cherry pick anything into any branch, and so the actual version number or build hash starts to break down totally.

hughsie:
I'll merge this to unstick the ODM affected, but I'm happy to continue discussing this stuff.

superm1:
I think you're right, the only safe way to do it would be ignore backports.  In terms of distros reverting and cherry picking anything on a branch, I do think we should discourage that behavior and slap wrists if we see it happen though.

PR #222, opened 2019-01-03
Show a detailed component requirement view when required

hughsie:
For the uncommon case of a firmware with a GUID requirement, or where a version
range is required switch to an 'advanced' table view that allows the user to
both view all existing requirements and to add more.

Fixes https://github.com/hughsie/lvfs-website/issues/221

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -736,6 +736,19 @@ class Component(db.Model):
                                   page='update')
         return problems
 
+    @property
+    def has_complex_requirements(self):
+        seen = []
+        for rq in self.requirements:
+            if rq.kind == 'firmware':
+                if rq.value not in ['firmware', 'bootloader']:
+                    return True
+            key = rq.kind + ':' + rq.value
+            if key in seen:
+                return True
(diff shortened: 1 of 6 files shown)


hughsie:
screenshot from 2019-01-03 16-16-46

superm1:
When "Hardware" is selected I think you need to prevent choosing some of the compare types and any values in the version field.

Or alternatively don't allow adding "hardware" here to avoid that set of bad values happening.

hughsie:
> I think you need to prevent choosing some of the compare types and any values in the version field

I think we need `ne` too -- although I agree `lt` and `ge` don't make a lot of sense for GUIDs.

hughsie:
> I think we need ne too

Actually, that's nonsense. CHIDs have to just be specified as OR text.

hughsie:
Okay, reworked to split out each as sections. 

screenshot from 2019-01-08 15-22-36

This works a lot better as you can put the right kind of placeholder in each section...

Issue #225, opened 2019-01-08
Importing a metainfo file with multiple CHIDs stores an invalid GUID


hughsie:
Setting:

    <hardware>foo|bar|baz</hardware>

Should store 3 requirements, not a string like `foo|bar|baz`. It needs to be split at upload time.

PR #226, opened 2019-01-08
Split hardware requires in the metainfo file

hughsie:
Fixes: https://github.com/hughsie/lvfs-website/issues/225

### DIFF
--- a/app/views_upload.py
+++ b/app/views_upload.py
@@ -111,12 +111,18 @@ def _create_fw_from_uploaded_file(ufile):
 
         # from requires
         for req in component.get_requires():
-            rq = Requirement(md.component_id,
-                             AppStreamGlib.Require.kind_to_string(req.get_kind()),
-                             req.get_value(),
-                             AppStreamGlib.Require.compare_to_string(req.get_compare()),
-                             req.get_version())
-            md.requirements.append(rq)
+            req_kind = AppStreamGlib.Require.kind_to_string(req.get_kind())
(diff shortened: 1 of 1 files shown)


app/views_upload.py
```diff
@@ -111,12 +111,18 @@ def _create_fw_from_uploaded_file(ufile):
...
-                             AppStreamGlib.Require.kind_to_string(req.get_kind()),
-                             req.get_value(),
-                             AppStreamGlib.Require.compare_to_string(req.get_compare()),
-                             req.get_version())
-            md.requirements.append(rq)
+            req_kind = AppStreamGlib.Require.kind_to_string(req.get_kind())
+            req_compare = AppStreamGlib.Require.compare_to_string(req.get_compare())
+            req_value = req.get_value()
+            if req_kind == 'hardware':
+                req_values = req_value.split('|')
```
superm1: I think you should also add this to documentation so it doesn't become a magic delimiter that you only find out about by checking source.

hughsie: Documentation in fwupd, or lvfs-website (e.g. metainfo pages)?

superm1: I think LVFS website, that's where the example for this is right?

hughsie: Already there I think: https://fwupd.org/lvfs/docs/metainfo/restrict

superm1: Great!

PR #229, opened 2019-01-09
Wip/hughsie/device checksums

hughsie:
screenshot from 2019-01-09 11-27-28

### DIFF
--- a/app/templates/component-checksums.html
+++ b/app/templates/component-checksums.html
@@ -46,6 +46,42 @@
   </tr>
 {% endif %}
 </table>
+
+<h3 class="mt-3">User-Submitted Reports</h3>
+
+{% if checksum_counts|length %}
+<table class="table">
+  <tr class="row">
+    <th class="col-sm-1 text-center">Count</th>
+    <th class="col-sm-9">Checksum</th>
+    <th class="col-sm-2">&nbsp;</th>
+  </tr>
+{% for (cnt, checksum) in checksum_counts %}
+  <tr class="row">
+    <td class="col-sm-1 text-center">
+{% if checksum in device_checksums %}
+      <span class="badge badge-success">{{cnt}}</span>
+{% else %}
+      <span class="badge badge-danger">{{cnt}}</span>
+{% endif %}
+    </td>
+    <td class="col-sm-9">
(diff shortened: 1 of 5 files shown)


app/views_component.py
```diff
@@ -130,6 +130,32 @@ def firmware_component_modify(component_id):
...
+    md = db.session.query(Component).filter(Component.component_id == component_id).first()
+    if not md:
+        return _error_internal('No component matched!')
+
+    # security check
+    fw = md.fw
+    if not fw:
+        return _error_internal('No firmware matched!')
+    if not fw.check_acl('@view'):
+        return _error_permission_denied('Unable to view other vendor firmware')
```
superm1: "other vendor" might not be true. It could be just w user missing this permission right?

app/templates/component-checksums.html
```diff
@@ -47,6 +47,38 @@
...
+<h3 class="mt-3">User-Submitted Reports</h3>
+<table class="table">
+  <tr class="row">
+    <th class="col-sm-2">Timestamp</th>
+    <th class="col-sm-2">Machine</th>
+    <th class="col-sm-6">Checksum</th>
+    <th class="col-sm-2">&nbsp;</th>
+  </tr>
+{% for (rpt, checksum) in reports_data %}
+  <tr class="row">
```
superm1: Assuming this takes off, this page is going to be huge eventually showing a lot of the same data.

I would wonder if it's more valuable to show
1) the identified checksums
2) number of machines that reported them.
3) "first" reported checksum details button (or maybe only show signed ones or something)

app/views_component.py
hughsie: right, will fix.

app/templates/component-checksums.html
hughsie: > this page is going to be huge eventually showing a lot of the same data

Agreed. I'll have another play and see what I can come up with.

hughsie: More like this?
screenshot from 2019-01-09 16-58-18

it's green if it's matching an existing checksum, red otherwise.

superm1:
Looks great to me

PR #232, opened 2019-01-10
Perform 'offline' runtime checks on uploaded firmware

hughsie:
Initially, add a simple plugin that checks the UEFI capsule header. Later we
could add further firmware analysis like uefi-firmware-parser or perhaps even
something like CHIPSEC.

Running this as a cron job every 5 minutes means the main uWSGI thread is not
waiting for the result, which means we could even do a malware scan which takes
a minute or two to complete.

Failures in the assay means a problem is added to the firmware, which prevents
it being moved to testing and stable. Some failures are waivable by a QA user
which means a problem is no longer generated.

### DIFF
--- a/app/templates/firmware-problems.html
+++ b/app/templates/firmware-problems.html
@@ -82,6 +82,24 @@
         can be moved to a different target.
       </p>
     </td>
+{% elif problem.kind == 'assay-pending' %}
+    <td class="col-sm-7">
+      <p class="card-text">
+        The LVFS runs tests on certain types of firmware to check they are valid.
+        Some tests are still pending and will be completed shortly.
+        You can refresh this page to find out when the firmware has been tested.
+      </p>
+    </td>
+{% elif problem.kind == 'assay-failed' %}
+    <td class="col-sm-7">
+      <p class="card-text">
+        A check on the firmware has failed.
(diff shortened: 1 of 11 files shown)


app/templates/firmware-problems.html
```diff
@@ -82,6 +82,24 @@
         can be moved to a different target.
       </p>
     </td>
+{% elif problem.kind == 'assay-pending' %}
+    <td class="col-sm-7">
+      <p class="card-text">
+        The LVFS runs tests on certain types of firmware to check they are valid.
+        Some tests are still pending and will be completed shortly.
```
superm1: Since you have this croned, can you also indicate when they're estimated to be done?  Similar to how you do things with the signing timeout.

superm1: Oh right... I guess that makes sense then how it is.

hughsie:
> Have you already ran it on the whole LVFS archive?

Yes, which took a ~2 minutes to complete.

> It would be good to notify those that it's going to be failing to those firmware owners and they'll be needing to waive before it goes live.

I'll write something to lvfs-announce and BCC people as required. I think Dell is all good, but there's one firmware from one of your ODMs that's currently in `private` -- so nothing to worry about IMHO. Lenovo and HP have some fixing to do :)

PR #233, opened 2019-01-10
Make local.py work better

superm1:


app/views_upload.py
```diff
@@ -55,7 +55,7 @@ def _user_can_upload(user):
     # works for us
     return True
 
-def _create_fw_from_uploaded_file(ufile):
+def _create_fw_from_uploaded_file(ufile, local=False):
```
hughsie: Is this hunk meant to be here?

app/views_upload.py
```diff
@@ -102,7 +102,11 @@ def _create_fw_from_uploaded_file(ufile):
         # from the release
         rel = component.get_release_default()
         md.version = rel.get_version()
-        md.install_duration = rel.get_install_duration()
+        #allows this to work on older versions of appstream-glib
+        try:
+            md.install_duration = rel.get_install_duration()
+        except AttributeError:
```
hughsie: Maybe `hasattr` might be cleaner here?

app/views_upload.py
superm1: Yes, it's to avoid the DB call for trying to match protocol.

app/views_upload.py
```diff
@@ -163,11 +163,12 @@ def _create_fw_from_uploaded_file(ufile, local=False):
             md.version_format = metadata['LVFS::VersionFormat']
 
         # allows OEM to specify protocol
-        if 'LVFS::UpdateProtocol' in metadata:
-            pr = db.session.query(Protocol).\
-                    filter(Protocol.value == metadata['LVFS::UpdateProtocol']).first()
-            if pr:
-                md.protocol_id = pr.protocol_id
+        if not local:
```
hughsie: could we just move this code to `upload()` then we don't need to tell the function whether to use the database?

app/uploadedfile.py
```diff
@@ -322,7 +322,7 @@ def _load_metainfos(self):
         for cf in cfs:
             self._load_metainfo(cf)
 
-    def parse(self, filename, data):
+    def parse(self, filename, data, local=False):
```
hughsie: I think rather than the _reason_, I think this should document the _behaviour_. e.g. switch around the logic and call it `use_hashed_prefix` or something.

app/metadata.py
```diff
@@ -145,9 +147,13 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri=''):
             ET.SubElement(rel, 'location').text = firmware_baseuri + md.fw.filename
 
             # add container checksum
-            if md.fw.checksum_signed:
+            if md.fw.checksum_signed or local:
                 csum = ET.SubElement(rel, 'checksum')
-                csum.text = md.fw.checksum_signed
+                #metadata intended to be used locally won't be signed
+                if local:
+                    csum.text = hashlib.sha1 (md.fw.filename).hexdigest()
```
hughsie: The filename? Shouldn't it be the archive data? i.e. just use `checksum_upload`

local.py
```diff
@@ -28,18 +28,19 @@
...
+            print('Processing %s...' % fn)
+            ufile = UploadedFile()
+            ufile.parse(fn, open(os.path.join(root, fn), 'r').read(), local=True)
+            fw = _create_fw_from_uploaded_file(ufile, local=True)
+            fws.append(fw)
 
     # write metadata
     print('Writing %s' % metadata_fn)
-    _generate_metadata_kind(metadata_fn, fws)
+    _generate_metadata_kind(metadata_fn, fws, firmware_baseuri="./", local=True)
```
hughsie: single quotes please.

superm1:
Sorted out the CI failures for this when you get another chance to look.

PR #234, opened 2019-01-11
Wip/hughsie/component checks

hughsie:
This adds a plugin to scan all firmware for trojans, viruses, malware and other malicious threats.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1081,6 +1081,15 @@ class Firmware(db.Model):
                 sc_lowest = md.security_claim
         if not sc_lowest:
             sc_lowest = SecurityClaim()
+
+        # been virus checked
+        test = self.find_test_by_plugin_id('clamav')
+        if test and test.ended_ts:
+            if test.success:
+                sc_lowest.add_attr('virus-safe', 'Virus checked using ClamAV')
+            else:
+                sc_lowest.add_attr('no-virus-safe', 'Virus check using ClamAV failed')
+
         return sc_lowest
 
     @property
--- a/app/models.py
+++ b/app/models.py
@@ -63,9 +63,9 @@ class Problem(object):
             return 'No valid update description'
         if self.kind == 'no-protocol':
             return 'No update protocol set'
-        if self.kind == 'assay-failed':
+        if self.kind == 'test-failed':
             return 'Firmware is not valid'
-        if self.kind == 'assay-pending':
+        if self.kind == 'test-pending':
             return 'Firmware tests are pending'
         return 'Unknown problem %s' % self.kind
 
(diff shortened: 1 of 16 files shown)


hughsie:
screenshot from 2019-01-11 13-00-42

lvfs_test.py
```diff
@@ -73,6 +73,10 @@ def setUp(self):
             value='com.hughski.colorhug',
         ), follow_redirects=True)
         assert b'Added protocol' in rv.data, rv.data
+        rv = self.app.post('/lvfs/settings/modify', data=dict(
+            clamav_enable='disabled',
```
superm1: Doesn't this take binary true/false?

plugins/clamav/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+        s.append(PluginSettingBool('clamav_enable', 'Enable scanning', True))
+        s.append(PluginSettingBool('clamav_detect_pua', 'Detect Possibly Unwanted Applications', True))
+        s.append(PluginSettingBool('clamav_use_daemon', 'Use clamd daemon', True))
+        return s
+
+    def ensure_assay_for_md(self, md):
+
+        # get settings
+        settings = _get_settings('clamav_enable')
+        if settings['clamav_enable'] != 'enabled':
```
superm1: Likewise you should use true/false here

plugins/clamav/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+                    '--nocerts',
+                    '--no-summary',
+                    fn]
+            if settings['clamav_detect_pua'] == 'enabled':
+                argv.append('--detect-pua=yes')
+        try:
+            ps = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            rc = ps.wait()
+            if rc == 2:
+                assay.add_fail('Scanning', ps.stderr.read())
```
superm1: On stderr do you see the version of clamav that claimed this failed? I think it would be useful to have in the problems report particularly to diagnose false positives relative to upsteam.

lvfs_test.py
hughsie: Confusingly not. The PluginSettingBool was originally just used for enabling/disabling the plugin -- at some point I need to revisit the settings stuff and should probably just use python types...

plugins/clamav/__init__.py
hughsie: Agreed. I'll just add it as a AssayAttribute, e.g.
screenshot from 2019-01-11 13-49-05

lvfs_test.py
superm1: Oh okay

hughsie:
> Did you already run this through to see if it flags anything in the archives?

Yup:

    ..../ab7afd17a258d0eacedffc7486d0b151c1c2677d-hughski-colorhug2-2.0.3.cab: Eicar-Test-Signature FOUND
    ----------- SCAN SUMMARY -----------
    Infected files: 1

superm1:
I think you need to have some documentation that reminds the administrator whenever the "system" clamav ruleset is upgraded to check all the binaries in LVFS for more false positives. 

Once a false positive is detected, the admin will need to have a button somewhere to waive them too.

hughsie:
> the admin will need to have a button somewhere to waive them too

I wondered about that too. Unfortunate, I think if fwver 1.2.3 is a false positive, 1.2.4 is going to be one too -- so that means the admin needs to get involved too. I that case I think it's better to blacklist the specific signature in `scan.conf` rather than do this per-firmware.

hughsie:
> I that case I think it's better to blacklist the specific signature in scan.conf

We can do better than that. I've been talking to the VT people: https://www.virustotal.com/#/monitor-overview -- basically we can help each other, but we'd have to use their private "Monitor" service rather than the public service that would leak the files out of embargo. They've offered enough private storage space that would last us a long time. The monitor service is used by AV companies to reduce or eliminate false positives and the **trusted** data from the LVFS would help them with that. We'd get access to all the scanners (not just clamav) for free, although it's not going to be integrated in the plugin as the notification of problems is very async, and also only via email. I think it solves the false positive problem in a much nicer way.

My concern is uploading files in private and embargo would leak details but the files would only get shared with the AV company if there is a false positive. The alternative is we only include files in testing or stable, but I think that's trying to get back the horse that's bolted.

Feedback welcome.

superm1:
So to avoid leaking data out of embargo and private, I'd propose another way you can use this other service.

Run it on the public stuff only.  If you throw up some flags, send an email the uploader and to the admin and you can make a decision together whether to bring it back to embargo for investigation.
If it's left public a statement under the security section that this was "scanned by VT" and may have this virus/malware.

hughsie:
I've re-worked things a bit (see first commit) -- I'm much happier with this now.

app/models.py
```diff
@@ -1077,6 +1081,15 @@ def security_claim(self):
                 sc_lowest = md.security_claim
         if not sc_lowest:
             sc_lowest = SecurityClaim()
+
+        # been virus checked
+        test = self.find_test_by_plugin_id('clamav')
```
superm1: Rather than hardcoded to clamav how about having plugins advertise the arrtibute 'virus-checker' or similar?

plugins/virustotal-monitor/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+            headers['User-Agent'] = settings['virustotal_user_agent']
+            files = {'file': ('filepath', file_obj, 'application/octet-stream')}
+            args = {'path': remote_path}
+            r = requests.post(settings['virustotal_uri'], files=files, data=args, headers=headers)
+            if r.status_code != 200:
+                test.add_fail('Uploading', r.text)
+                return
+
+        # success
+        test.add_pass('Uploaded', 'All OK')
```
superm1: Do you never check back in for a result?

hughsie: It's not really how this service works. The public scan means you can do just that, i.e. it's like a virus scanner that just does the work for you offline. The VT Monitor service is more of a "don't flag this file as a false positive" service, and if there is any disagreement about if it's a valid virus or a false positive it's all done over email with the AV vendor concerned, as I understand it. Of course, we could certainly scan files in public/testing with the VT public scanner much like we do with clamav, although that's a separate plugin I think.

app/models.py
hughsie: If we scanned with VT public (which used basically *all* of the virus scanners available on the market) would we still show the 'Scanned with ClamAV text'? I'd argue the future' Scanned with VT' text supersedes the ClamAV scan completely. If so, I'd rather just hardcode this here, rather than assigning `provides` and `priorities` to the plugins for this general case.

hughsie:
I'll merge to avoid vendors waiving things that need to be waived again. I happy to follow up on all unresolved points as follow up commits -- tanks.

superm1:
I think that your comments regarding how scanning are fine, it was just weekend and I didn't get to them yet.  No other follow ups needed.

Issue #239, opened 2019-01-16
Vendor accounts to use Two Factor Authentication


Sean-StarLabs:
To add another level of protection beyond a username and password

@hughsie mentioned flask as a solution
https://blog.miguelgrinberg.com/post/two-factor-authentication-with-flask

PR #243, opened 2019-01-22
Add 2 factor authentication using OTP

hughsie:
Fixes https://github.com/hughsie/lvfs-website/issues/239

### DIFF
--- a/app/templates/profile.html
+++ b/app/templates/profile.html
@@ -37,10 +37,67 @@
     </div>
 {% endif %}
     <div class="form-group row">
+      <div class="col-sm-2">Two Factor Auth</div>
+      <div class="col-sm-10">
+        <span class="switch mt-3 mb-3">
+          <input class="switch mt-3" type="checkbox" name="is_otp_enabled" id="is_otp_enabled" value="1" {{'checked' if u.is_otp_enabled}} {{'disabled' if not u.is_otp_working}}>
+          <label for="is_otp_enabled">
+            Using OTP secures your account more than using just using a password.
+          </label>
+        </span>
+{% if g.user.is_otp_working %}
+        <button type="button" class="btn btn-secondary float-right" data-toggle="modal" data-target="#otpSetupModal">
+          Show QR Code
+        </button>
+{% else %}
+        <button type="button" class="btn btn-secondary float-right" data-toggle="modal" data-target="#otpSetupModal">
+          Set up using QR code
+        </button>
+{% endif %}
+      </div>
+    </div>
+    <div class="form-group row">
       <div class="col-sm-2"></div>
       <div class="col-sm-10">
     <input type="submit" class="btn btn-primary btn-large" class="submit" value="Modify">
       </div>
     </div>
   </form>
+
+<!-- Modal -->
+<div class="modal" id="otpSetupModal" tabindex="-1" role="dialog" aria-labelledby="otpSetupModalLabel" aria-hidden="true">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title" id="otpSetupModalLabel">Setup OTP</h5>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        <p>
+          Please use <a href="https://freeotp.github.io/">FreeOTP</a>,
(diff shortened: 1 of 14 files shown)


app/templates/profile.html
```diff
@@ -36,11 +36,66 @@ <h1>Modify Local User</h1>
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title" id="otpSetupModalLabel">Setup OTP</h5>
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        <p>
+          Please start FreeOTP and scan the following QR Code:
```
superm1: I suspect most people are going to use something like Google Authenticator or something else they already have for other sites maybe it's worth making it clear that they can use any client that supported this.

hughsie: `Please start FreeOTP, Google Authenticator or any other suitable application and scan the following QR Code` ? I could hyperlink FreeOTP and Google Authenticator too.

superm1: Yeah I think that's good

superm1:
I feel like this kind of attribute is actually something the manager should be able to "mandate" but then that means a new state in state machine for enabled but not working.

hughsie:
> this kind of attribute is actually something the manager should be able to "mandate" 

This gets really hard, as a manager adding accounts for users would have to get the user to scan the QR code *before* the initial login. Or I suppose we could deal with the 2FA setup on first login, although that needs some careful design to avoid it being insecure. I think it's a new PR for new functionality, not this one IMHO.

superm1:
Yeah that's what I was thinking too.  It's definitely a new state to define on doing it for first login, similar to a mandatory password change request.  I don't know that it's necessary, just wanted to make sure it's discussed.

app/templates/profile.html
hughsie: Fixed.

hughsie:
Other than the organisation-policy for 2FA, no other objections? I'm not 100% sure about the login box now:

screenshot from 2019-01-22 15-22-50

I'm quite keen on keeking things simple and not do client-side JS to hide/show the OTP if it's not configured. Hopefully longer term more local accounts than not will have 2FA enabled, or it'll be a constant reminder to do so.

hughsie:
> 2FA should still work with OAuth right?

2FA and OAuth are mutually exclusive from the LVFS point of view. If you want to use 2FA using OAuth you need to configure it Azure-side so-to-speak.

> What about making it a 2 stage login screen

I thought a lot about this; I didn't find any way it would be secure. I'd have to store a login cookie that indicated that the user was "half" logged on, and that broke quite a few assumptions in the flask code that I didn't quite understand.

superm1:
> 2FA and OAuth are mutually exclusive from the LVFS point of view. If you want to use 2FA using OAuth you need to configure it Azure-side so-to-speak.

Right, that's what I was thinking. So it needs to be pretty explicit which fields go to which login.

> I thought a lot about this; I didn't find any way it would be secure. I'd have to store a login cookie that indicated that the user was "half" logged on, and that broke quite a few assumptions in the flask code that I didn't quite understand.

Alright so what about re-organizing the page so that it's a "left and right" split with a top delimitter.

```
Header text
------------
      |
OAuth | Password login
      |
```

Then in the password login side of the window you make the 2FA thing indented and a different color.
The indentation will look better with the horizontal view.

hughsie:
> Alright so what about re-organizing the page

Maybe this is the straw that breaks the camels back. What about:

* Making the `Login` button turn into a dropdown, with two sub-entries, `Azure AD` and `Local Account`, possibly with the local option first.
* The `Azure AD` button would take you to the SSO if applicable
* The `Local Account` would show the username/password/otp, perhaps remaining as a modal dialog.

superm1:
I can tell you that's _likely_ to prompt anyone who does use Azure to start trying to enter domain credentials into LVFS.  Until azure was made more prominent I recall that we were having that happen.

**Counter-proposal:**
Make the login page two stage.
1. user enters email address
2. LVFS either redirects to oauth provider or returns back a page to do login process.
3. On returned login process page LVFS can return 2FA section in the page if it's needed.

I would think that should avoid breaking security (needing to track a cookie or anything like that).

hughsie:
> for your ODMs?

Ignore me, they're probably not using OAuth.

hughsie:
@superm1  -- something like this?

screenshot from 2019-01-23 11-58-31
screenshot from 2019-01-23 11-58-22

hughsie:
Okay, comments welcome ^^

superm1:
That LGTM

PR #246, opened 2019-01-24
Add support for a per-release source and details URL

hughsie:
The source URL allows us to comply with our various obligations when shipping
firmware built from GPL licensed sources. The details URL allows vendors to
include a link to a full HTML details page about the specific release.


hughsie:
> recommend to switch "source url" to explicitly "source code url"

I really want to match the AppStream nomenclature here, then we can just point to someone elses documentation not our own.

PR #258, opened 2019-02-15
Make the firmware page more beautiful

hughsie:
I spend a lot of time looking at this, so make it look nice.


hughsie:
screenshot from 2019-02-15 13-05-01

PR #259, opened 2019-02-15
Make the event logs nicer to read

hughsie:


hughsie:
screenshot from 2019-02-15 16-37-20

PR #271, opened 2019-03-10
Simplify cross/tick icons to make them display in Firefox

### LINKED ISSUE #270: Cross / Check icons not displayed correctly in Firefox
In Firefox, the vendor list is unusable because the icons are not displayed correctly:

Screenshot_2019-03-09_13-27-01

In Chromium and Inkscape, they look like they should.

FloEdelmann:
Closes #270.

Apparently, the `clip-path` inside the embedded styles are considered harmful in Firefox's CSRF protection. Since the icons are really simple, those embedded styles are not really needed either.

This simplification should make them display correctly. (I couldn't test that because CSRF protection is disabled for `http://localhost` / `file://`.)


hughsie:
Sounds plausible to me, lets give this a go. Thanks!

FloEdelmann:
Looks fine now! :smiley: 

Screenshot_2019-03-10_14-43-34

hughsie:
Many thanks for fixing that!

PR #272, opened 2019-03-10
Wip/hughsie/certificates

hughsie:


hughsie:
Screenshot from 2019-03-11 15-05-22

hughsie:
Screenshot from 2019-03-11 15-06-44

app/templates/profile-crts.html
```diff
@@ -0,0 +1,69 @@
...
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        <p>
+          You can upload the client certificate from <code>/var/lib/fwupd/pki/client.pem</code> or generate one yourself using:
+        </p>
+        <pre>
+certtool --generate-privkey --outfile secret.key
+certtool --generate-self-signed --load-privkey secret.key --outfile client.pem
```
superm1: I think it's probably better to just ask people to use a particular version of fwupd.  Asking for independent generation won't be super useful other than to confuse people about how to put it back into fwupd when fwupd automatically generates a "different" certificate.

app/templates/profile-crts.html
```diff
@@ -0,0 +1,69 @@
...
+        </pre>
+        <form method="post" action="{{url_for('.user_certificate_add')}}">
+          <div class="form-group">
+            <label for="certificate">Paste certificate here:</label>
+            <textarea id="text" rows="25" cols="65" class="form-control fixed-width" type="password" name="text" placeholder="-----BEGIN CERTIFICATE-----"></textarea>
```
superm1: I'm curious - why even ask for a copy/paste?  Isn't that asking for whitespace errors when you compare Windows/Linux uploaders etc?

How about instead just a button to upload and browse for a file?

app/templates/profile-crts.html
hughsie: Right, I didn't make this very clear. If you wanted to use a corporate-issued PKCS-7 certificate signed by Dell Inc. you could use that in place of the client-generated one. If you copied the personalized certificate and private key around to all the systems you were testing on you only need to upload one certificate to the LVFS.

app/templates/profile-crts.html
hughsie: Hmm, I guess uploading the file is a lot simpler. Will try again.

app/templates/profile-crts.html
superm1: I think in this case it's still clearer to say to use the same certificate being used by fwupd by pulling it from `/var/lib/fwupd/...`.  Maybe separately in fwupd documentation mention that a corporate cert can be used instead of generating one by putting it in that directory with proper permissions etc?

app/templates/profile-crts.html
hughsie: > the default permissions on /var/lib/fwupd/ certificate file work out too though

Is this sane?

    $ ls /var/lib/fwupd/pki/ -l
    -rw-r--r--. 1 root root 1383 Mar  8 09:21 client.pem
    -rw-------. 1 root root 2459 Mar  8 09:21 secret.key

superm1: Yep, that looks sane to me.

app/templates/profile-crts.html
hughsie: Okay, I think I've got something better now. I've added a note in case the client.pem does not already exist as it's only created on demand now. Better text very welcome. Thanks.

app/templates/profile-crts.html
```diff
@@ -0,0 +1,59 @@
...
+
+<div class="card mb-3">
+  <h3 class="card-header list-group-item-info">Where can I find the client certificate?</h3>
+  <div class="card-body">
+    <p class="card-text">
+      If <code>/var/lib/fwupd/pki/client.pem</code> does not already exist it
+      can be created using <code>fwupd >= 1.2.6</code>:
+    </p>
+    <pre class="card-text">
+sudo /usr/libexec/fwupd/fwupdtool self-sign xxx
```
superm1: This looks a lot better!
Just something to note;
* what `xxx` is isn't clear
* that path isn't the same for all distros (debian/ubuntu don't use `/usr/lib/libexec`)

So that makes me wonder if there should just be an extra command in fwupd `generate-cert` that calls self-sign with essentially `/dev/null`?

Then instead of referring to that path like you did, maybe like this to hit the two common ones.
```
sudo /usr/libexec/fwupd/fwupdtool generate-cert
```
or
```
sudo /usr/lib/fwupd/fwupdtool generate-cert
```

hughsie: I think every time we make this even harder with extra steps we're going to get fewer people actually uploading certificates. I think the best thing is to patch fwupd to generate the certificate on startup, but not failing startup in the event of an older GnuTLS.

superm1: Great idea.

hughsie: https://github.com/hughsie/fwupd/pull/1089

Issue #236, opened 2019-01-14
Need to downgrade the test failure for 4kB-aligned UpdateCapsule header


hughsie:
Lots of Lenovo ThinkPad firmware doesn't align the UpdateCapsule payload to a multiple of 4kB. The original justification was in 8612318158c9d199d539dd4d0dfcc33d4264d8fa which references problems on other architectures, so it makes sense to choose 4kB when we're generating a header to be as portable as possible. When we're downloading a capsule onto x64-only hardware it seems overkill to enforce the header is 4kB in size, possibly. I think we should relax the check to be `hdrsz < 4kB | hdrsz % 4kB == 0`. @superm1 agree?

superm1:
But regarding what to enforce from lvfs-website side of things, that proposed relaxation is probably good enough.

hughsie:
> The FSP provided by Intel will mandate page alignment and reject anything else

That kinda suggests we should stick to the 4kB alignment check then, to me at least. It's waivable afterall.

superm1:
It's a little baffling to me that this behavior hasn't raised flags for anything pre-production, one of our IBVs hit it on several platforms in the upcoming generations.

You might want to double check with the IBVs on Windows behavior here.  Maybe Windows unpacks the capsule header and repacks it into page aligned chunks and that's why it's not an issue there?

hughsie:
> Maybe Windows unpacks the capsule header and repacks it into page aligned chunks

That seems a semi-sane thing to do. At the moment fwupd doesn't do anything like that (if the header is present, but not 4kB aligned), and I'm also wondering if this is some of the unexplained failures we've been getting from various ODMs over the last few months.

hughsie:
> The FSP provided by Intel will mandate page alignment and reject anything else

Lenovo are saying the page alignment is important for the capsule (i.e. for the start of the header) and not important for the actual payload section. Could you find out more details about this please, and maybe we can find out what Windows does. I'm not against repacking the capsule if we have to, but I am worried about adding a heuristic for the presence of a capsule header given there is no _magic signature_ we can check.

superm1:
I'm not sure I'm going to be able to get a lot of help here.  There are a few black boxes at play..

> but I am worried about adding a heuristic for the presence of a capsule header given there is no magic signature we can check.

The heuristic of look for the GUID in XML to match the correct location in the capsule header seems reasonable to me.

PR #273, opened 2019-03-11
Don't enforce a 4kB capsule header size

hughsie:
Only the capsule header has to be aligned to the 4kB page size.

Fixes: https://github.com/hughsie/lvfs-website/issues/236


superm1:
I wish there was more clarity here; but I don't have an opposition to this.

hughsie:
> I wish there was more clarity here

You and me both. I'm going to continue advising people implementing this stuff to use a 4k header to be on the safe side.

PR #275, opened 2019-03-11
Test the DFU footer for firmwares with protocol org.usb.dfu

hughsie:
These are similar to the checks done for UEFI capsules.

### DIFF
--- a/plugins/dfu/__init__.py
+++ b/plugins/dfu/__init__.py
@@ -1,0 +1,138 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 Richard Hughes <richard@hughsie.com>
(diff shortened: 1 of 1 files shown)


plugins/dfu/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        return s
+
+    def ensure_test_for_fw(self, fw):
+
+        # get settings
+        settings = _get_settings('dfu_check_footer')
+        if settings['dfu_check_footer'] != 'enabled':
+            return
+
+        # only run for capsule updates
```
superm1: capsule?

superm1:
Just that nit, and LGTM

plugins/dfu/__init__.py
hughsie: Fixed, thanks.

PR #276, opened 2019-03-11
Add metadata about whether a firmware has a testable header

hughsie:
FIXME: needs Alembic migration script


superm1:
Concept looks good, documentation seems fine.

PR #278, opened 2019-03-13
Add a page showing new firmware on the LVFS

hughsie:
I'm often asked this, and from a personal point of view it's easier in a list
than sorting through the event log and checking which firmware is in testing or
stable.

### DIFF
--- a/app/templates/firmware-new.html
+++ b/app/templates/firmware-new.html
@@ -1,0 +1,38 @@
+{% extends "default.html" %}
+{% block title %}New Firmware{% endblock %}
+
+{% block content %}
+<h1>New Firmware</h1>
+<p>
+  This list show the last {{limit}} public firmwares available on the LVFS.
+</p>
+
+{% if fwevs|length == 0 %}
+<p class="text-muted">
+  No firmware has been uploaded yet.
+{% else %}
+<table class="table">
+  <tr class="row">
+    <th class="col-sm-2">When</th>
+    <th class="col-sm-5">Hardware</th>
+    <th class="col-sm-3">Version</th>
+    <th class="col-sm-2">&nbsp;</th>
+  </tr>
+{% for fwev in fwevs %}
+  <tr class="row">
+    <td class="col-sm-2">{{format_humanize_naturaltime(fwev.timestamp)}}</td>
+    <td class="col-sm-5">
+      {{fwev.fw.mds[0].developer_name_display}}
+      {{fwev.fw.mds[0].name_display}}
+    </td>
+    <td class="col-sm-3">{{fwev.fw.version_display}}</td>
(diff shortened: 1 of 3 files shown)


hughsie:
Screenshot from 2019-03-13 10-32-39

app/templates/firmware-new.html
```diff
@@ -0,0 +1,38 @@
+{% extends "default.html" %}
+{% block title %}New Firmware{% endblock %}
+
+{% block content %}
+<h1>New Firmware</h1>
+<p>
+  This list show the latest public firmware being made available on the LVFS.
```
superm1: Maybe worth mentioning it's only showing last 200…

app/templates/firmware-new.html
```diff
@@ -0,0 +1,38 @@
...
+  <tr class="row">
+    <td class="col-sm-2">{{format_humanize_naturaltime(fwev.timestamp)}}</td>
+    <td class="col-sm-5">
+      {{fwev.fw.mds[0].developer_name_display}}
+      {{fwev.fw.mds[0].name_display}}
+    </td>
+    <td class="col-sm-3">{{fwev.fw.version_display}}</td>
+    <td class="col-sm-2">
+      <a href="/lvfs/device/{{fwev.fw.mds[0].guids[0].value}}"
+         class="btn btn-info btn-block">Details</a>
```
superm1: A little asterisk or something to indicate which are in testing vs stable is probably good too.

app/views_firmware.py
```diff
@@ -61,6 +61,25 @@ def firmware(show_all=False):
 def firmware_all():
     return firmware(True)
 
+@app.route('/lvfs/firmware/new')
+@app.route('/lvfs/firmware/new/<int:limit>')
+def firmware_new(limit=200):
+
+    # get a sorted list of vendors
+    fwevs = db.session.query(FirmwareEvent).\
+                order_by(FirmwareEvent.timestamp.desc()).\
```
superm1: So this means it's going to jump the list again when it goes from testing -> stable right? Expected that some get to jump twice?

app/templates/firmware-new.html
hughsie: `This list show the last {{limit}} public firmwares available on the LVFS.`?

app/views_firmware.py
hughsie: I did ponder only including the ones just in stable -- most users won't know how to enable the testing remote.

superm1: I think that makes more sense too

app/templates/firmware-new.html
superm1: Yeah that's good

hughsie:
Agreed, only public firmware works better.

PR #281, opened 2019-03-14
Wip/hughsie/certificates

hughsie:


hughsie:
RHEL 7 CI failing as GnuTLS is too old. Not sure how to fix that yet; it's not exactly self contained.

hughsie:
> With CI failure, maybe move it over to Debian or Fedora or something instead

I really wanted it to match what we deploy onto to minimize suprises. I was tempted to use flatpak, and just use gcab, appstream-glib and gnutls from that, rather than trying to backport new stuff into RHEL. Is using flatpak insane?

superm1:
For me flatpak has always taken way longer to spin up than native. On that reason I'd say not a good idea for a service.

hughsie:
> flatpak has always taken way longer to spin up than native

Yes, perhaps a little. But it does also give us the Bubblewrap sandbox, which for crypto code like GnuTLS is probably no bad thing. I spent about 2 hours this morning trying to create a static version of `certtool` that could be deployed into RHEL 7, but on deploying I found the libnettle library also has been API bumped and thus would need linking-in to. This descended into chaos pretty quick.

superm1:
Yikes.  Well if you do end up with success down this path, I think the fwupd snap would benefit from this too.  I found that bumping gnutls also leads to chaos that doesn't work (eg https://github.com/hughsie/fwupd/commit/84b1877b9091f3c8439ba2050758c3d558601f9a)

hughsie:
CI seems to like it. Any reservations? I guess the other way is to use python-cryptography but then we have two crypto implementations and I quite like the fact we can use the same libray in the fwupd daemon and on the LVFS. I tried to get the Python bits into GnuTLS a while ago but no luck: https://github.com/AGProjects/python-gnutls/issues/6

superm1:
As long as you're happy with the performance of how it runs this way I don't have any reservations.  I prefer to see the same bits on both ends too.

PR #287, opened 2019-03-27
correct the number of splits when parsing certtool

ezr:
This should only break the line into two parts, so 1 should be used as the number of splits. Set to 2, the code raises ValueError when parsing "Signature status: verification failed: Public key signature verification has failed.". Without this change I think it would be slightly easier to pull off something like CVE-2018-12020 as an attacker.

I haven't tested running the app with this change. I really just wanted to comment - feel free to deny this PR.


hughsie:
Many thanks for the patch. In GLib, the integer is the max number of _sections_, not the max number of _splits_, so this is why I got confused.

PR #290, opened 2019-03-29
Add a component property to express the firmware scope

hughsie:
Some firmwares only update one part of the system, e.g. the EC or ME firmware.
Other updates include all the updates needed for the whole system, and vendors
have been doing different things with the component name due to this.

Some examples of the brokenness:

* `Logitech Unifying Receiver` -> `Logitech Logitech Unifying Receiver`
* `Thunderbolt Firmware Update for AKiTiO Node Lite` -> `AKiTiO Thunderbolt Firmware Update for AKiTiO Node Lite`
* `Star LabTop Mk III Embedded Controller Firmware` -> `Star Labs Star LabTop Mk III Embedded Controller Firmware`
* `ChengMing 3967 System Update`
* `Thunderbolt Firmware Update for 0x07E6, XPS9370`
* `Z440/Z640/Z840 Workstation System BIOS`

The translator teams are also frustrated that they cannot translate terms like
`System Update` and also provide the correct versions of `Device`.

To fix, add an enumerated set of firmware 'scopes' that can be set by the
uploader in the metainfo.xml file (or changed the LVFS) which automatically
set the name suffix.

On firmware upload, remove any of the words we're going to suffix and provide
a nag warning to the uploader only. For existing firmware convert them to having
the correct scope and fix the names used in the AppStream metadata.

In the future we can set the LVFS:Scope entry in the metadata and provide
translations in the frontend. This obviously needs changes to fwupd and more
importantly gnome-software and the KDE software center so gently phase this in.


app/metadata.py
```diff
@@ -232,6 +233,7 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
                 child.set('key', 'LVFS::VersionFormat')
                 child.text = md.version_format
                 elements['LVFS::VersionFormat'] = child
+            # deliberately not including LVFS::Scope here...
```
hughsie: Maybe this should be `LVFS::UpdateScope` to match the others?

hughsie:
Screenshot from 2019-03-29 11-41-22
Screenshot from 2019-03-29 11-40-11

superm1:
This has happened of course because the field is just `name` and there is no guidance on what is supposed to go in name, and what order it should be presented.

Have you tried to run this across the device list to see what it actually does to everything?

Another way to do it is only show the translated strings fetched from the fwupd device where you have full control.  So in `gnome-software` rather than show
`Precision Tower 3440 System Update version 00.01.00.03` you would pull from the fwupd device release and show it as `Precision Tower 3430 System Firmware version 00.01.00.03`.  So you predictably know that:
* `System Firmware` was inserted from fwupd
* `Precision Tower 3430` came from the SMBIOS tables
* for the locale where putting version makes sense

Then if you click the details you just see the wild west title of whatever was in the release, which likely isn't going to be translated anyways.

hughsie:
> Have you tried to run this across the device list to see what it actually does to everything

Yes, this is how I did the migration script

> only show the translated strings fetched from the fwupd device where you have full control

I think the way to do this is to let the translators join up the strings together. They hate it when we try to play tricks with "_string surgery_" as it breaks in all kinds of interesting ways.

migrations/versions/b59b7d74ed0f_add_component_scope.py
```diff
@@ -0,0 +1,89 @@
...
+                    md.scope_id = apxs['system'].scope_id
+                    break
+            for suffix in ['Embedded Controller Firmware']:
+                if md.name.endswith(suffix) or md.summary.endswith(suffix):
+                    md.scope_id = apxs['ec'].scope_id
+                    break
+            for suffix in ['ME Firmware']:
+                if md.name.endswith(suffix) or md.summary.endswith(suffix):
+                    md.scope_id = apxs['me'].scope_id
+                    break
```
superm1: rather than only putting this in the migration script, what would you think about this part remaining automatic based on these magic migration strings?

hughsie: I did think of doing that, but I really want to annoy the uploader into fixing it in the metainfo file template. If it's not changed before upload time then if the user just double-clicks on the .cab file then we don't have the appstream metadata and fall back to the untranslated name. This is the "stick" side of the LVFS I'm afraid.

superm1: I guess i'm at odds because this really comes down to adjusting some build scripts which are of course only a 3 line change but take forever to push out to all the platforms.
So I don't want to slow down the people who have nothing to do with that build script process.

I guess you really want to avoid automatically rewriting the metainfo.xml from LVFS side though, right?

hughsie: > I guess you really want to avoid automatically rewriting the metainfo.xml from LVFS side 

Right, shall we say an "agency" really wanted the vendor to sign the metainfo files long term, and modifying them on the LVFS obviously breaks that.

superm1: And keep them signed all the way to client?

At least for UEFI - well they're typically already signed in detached file if you're trying to reuse the files on WU.  Just with something that can't be parsed yet by LVFS :)

hughsie: > Just with something that can't be parsed yet by LVFS 

There is a huge amount of legal, licensing and political problems with just reading the .cat file. I'm trying to sort out at least two of those at the moment...

superm1: Right..

So another thought then; the client side has to change anyway to support this scope stuff.  Can the client also take a simple heuristic like this?  If no scope set, you see these words or this protocol set do the best you can?

hughsie: > the client side has to change anyway to support this scope stuff

Right, I have a patch waiting for fwupd if you like the concept here.

> If no scope set

I think the only thing we can do is build untranslated <name> entries here, and then in about 6 months have a flag day where we stop adding "System Update" as a suffix and start setting `LVFS::Scope` and expecting the front ends to do it.

I suppose a halfway house would be to strip off the known suffixes in the client too.

superm1: I think flag days for this kind of change are fine.

app/metadata.py
superm1: I do think it should be `UpdateScope` yeah to match the others, agree.

migrations/versions/b59b7d74ed0f_add_component_scope.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        op.add_column('components', sa.Column('scope_id', sa.Integer(), nullable=True))
+        op.create_foreign_key(None, 'components', 'scopes', ['scope_id'], ['scope_id'])
+
+    if 1:
+        db.session.add(Scope(value='unknown', name='Unknown'))
+        db.session.add(Scope(value='system', name='System Firmware'))
+        db.session.add(Scope(value='device', name='Device Firmware'))
+        db.session.add(Scope(value='ec', name='EC Firmware'))
+        db.session.add(Scope(value='me', name='ME Firmware'))
```
superm1: So I think you're missing a few common-ish ones that don't really make sense with the device acronym in a composite context:
`hub`: "Hub Firmware"
`tbt`: "Thunderbolt firmware"

app/views_upload.py
```diff
@@ -348,6 +348,13 @@ def upload():
     fw.checksum_signed = hashlib.sha1(cab_data).hexdigest()
     fw.is_dirty = True
 
+    # fix name
+    for md in fw.mds:
+        name_fixed = _fix_component_name(md.name, md.developer_name_display)
```
superm1: What happens with 2+ word developer names?  Do these need to be split and only look at first word (or is `developer_name_display` handling that)?

hughsie: `developer_name_display` does that for us.

migrations/versions/b59b7d74ed0f_add_component_scope.py
hughsie: Aren't hub and thunderbolt devices just "devices"?

app/metadata.py
hughsie: Fixed

migrations/versions/b59b7d74ed0f_add_component_scope.py
superm1: Well I guess let me use the dell-dock plugin as an example to illustrate.

If you opened up a composite CAB for a dock (and gnome-software was changed to show all the components) here are some of the current titles:
```
USB 3.1 Generation 2 controller Update
USB 3.1 Generation 1 controller Update
Thunderbolt controller update
```

I feel like those should have been covered by this, but I don't see how they would be.

hughsie: > Thunderbolt firmware

ThunderBolt is a trademark and shouldn't be translated, so in my eyes this is where the "line" is for deciding that `Embedded` should be a scope and not a `Thunderbolt` -- perhaps the word `Controller` is super useful to an end user; from a 40,000ft view it's really just another _device_ on their system. That said, `Thunderbolt Controller Device Update` isn't awesome, and `Thunderbolt Device Update` is indeed misleading. Maybe if we had `controller` as a scope enum that fixes things also like `USB 3.1 Generation 1 controller Update` -> `USB 3.1 Generation 1 {Controller Firmware}`. Not sure I 100% like the word "controller" tho.

> If you opened up a composite CAB for a dock

Again, from a 40,000ft point of view, you're updating the dock, rather than component things in it. I guess this makes the case for `Thunderbolt Dock TB16` -> `TB16 Thunderbolt {Dock}` where we have a `dock` scope that's translated.

Maybe the problem is the name `scope` -- if it was called something like `name_suffix` or `component_kind` it would be clearer when to use this feature. I'm certainly open to using a different noun entirely, although I struggled to find a single word to use. Ideas very welcome. Also, no panic on this, as we'll need to give the OEM/ODMs some heads-up before their uploads start failing or we show new UX to unsuspecting interns.

superm1: If you're going back up to the 40,000 foot view, why even tell them what an embedded firmware is or what an ME firmware is?  Bury the details further down and just say "System" or "Device" based on whether it's internal or external and leave it at that.

hughsie: The idea behind scope was really "_update one specific thing_" v.s. "_update a collection of things on the same system_" rather than _internal_ v.s. _external_. For the specific thing case, I was trying to translate the suffix in the name as that's something I get reminded about from translators quite a bit.

hughsie: Okay, I was talking about this feature with a colleague of mine for a good chunk of this afternoon. As he said, it's not "scope", it's a display "style" and I tend to agree. Certainly using `style` feels more right. I've neutered a few things so the style is optional. I'll close this PR and open a new one with the new branch to avoid confusing everything.

PR #291, opened 2019-03-29
Add list of featured vendors on the index page

hughsie:
This replaces the now-dated Spectre information card.


hughsie:
Screenshot from 2019-03-29 15-09-04

superm1:
LGTM

PR #292, opened 2019-03-29
Add quotes to the front page

hughsie:
This makes the front page more engaging. With the theme change we can make
these look better, but cards will do until we have more quotes.


hughsie:
I'm not sure you actually wrote the quote below, but I got it from a slide-deck from Christian. If you didn't, can you write something nice please? :)
Screenshot from 2019-03-29 16-21-21

superm1:
I do recall saying something like that.  Seems fine to me.

PR #305, opened 2019-04-03
Add a component property to express the firmware style

hughsie:
Some firmwares only update one part of the system, e.g. the EC or ME firmware.
Other updates include all the updates needed for the whole system, and vendors
have been doing different things with the component name due to this.

Some examples of the brokenness:

* `Logitech Unifying Receiver` -> `Logitech Logitech Unifying Receiver`
* `Thunderbolt Firmware Update for AKiTiO Node Lite` -> `AKiTiO Thunderbolt Firmware Update for AKiTiO Node Lite`
* `Star LabTop Mk III Embedded Controller Firmware` -> `Star Labs Star LabTop Mk III Embedded Controller Firmware`
* `ChengMing 3967 System Update`
* `Thunderbolt Firmware Update for 0x07E6, XPS9370`
* `Z440/Z640/Z840 Workstation System BIOS`

The translator teams are also frustrated that they cannot translate terms like
`System Update` and also provide the correct versions of `Device`.

To fix, add an enumerated set of firmware 'styles' that can be set by the
uploader in the metainfo.xml file (or changed the LVFS) which automatically
set the name suffix.

On firmware upload, remove any of the words we're going to suffix and provide
a nag warning to the uploader only. For existing firmware convert them to having
the correct scope and fix the names used in the AppStream metadata.

In the future we can set the LVFS:UpdateStyle entry in the metadata and provide
translations in the frontend. This obviously needs changes to fwupd and more
importantly gnome-software and the KDE software center so gently phase this in.

### DIFF
--- a/app/metadata.py
+++ b/app/metadata.py
@@ -4,7 +4,7 @@
 # Copyright (C) 2015-2018 Richard Hughes <richard@hughsie.com>
 # Licensed under the GNU General Public License Version 2
 #
-# pylint: disable=too-many-statements
+# pylint: disable=too-many-statements,too-many-locals,too-many-nested-blocks
 
 import os
 import hashlib
@@ -42,7 +42,8 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
         component = ET.SubElement(root, 'component')
         component.set('type', 'firmware')
         ET.SubElement(component, 'id').text = md.appstream_id
-        ET.SubElement(component, 'name').text = md.name
+        # until all front ends support <category> append the suffix */
+        ET.SubElement(component, 'name').text = md.name_with_category
         ET.SubElement(component, 'summary').text = md.summary
         ET.SubElement(component, 'developer_name').text = md.developer_name
         if md.description:
@@ -200,6 +201,23 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
                 sz = ET.SubElement(rel, 'size')
                 sz.set('type', 'download')
                 sz.text = str(md.release_download_size)
+
+        # deliberately not including <category> here until 2020-01-01
+        if False:                       # pylint: disable=using-constant-test
+            cats = []
+            for md in mds:
+                if not md.category:
+                    continue
+                if md.category.value not in cats:
+                    cats.append(md.category.value)
+                if md.category.fallbacks:
+                    for fallback in md.category.fallbacks.split(','):
+                        if fallback not in cats:
+                            cats.append(fallback)
+            if cats:
+                categories = ET.SubElement(root, 'categories')
+                for cat in cats:
+                    ET.SubElement(categories, 'category').text = cat
 
         # provides shared by all releases
         elements = {}
--- a/app/templates/private.html
+++ b/app/templates/private.html
@@ -61,6 +61,7 @@
           <div class="dropdown-divider"></div>
           <a class="dropdown-item" href="/lvfs/settings">Server Settings</a>
           <a class="dropdown-item" href="{{url_for('.protocol_all')}}">Flashing Protocols</a>
+          <a class="dropdown-item" href="{{url_for('.category_all')}}">Update Categories</a>
           <a class="dropdown-item" href="{{url_for('.agreement_list')}}">Agreements</a>
         </div>
       </li>
@@ -112,6 +113,7 @@
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='restrict')}}">Adding Restrictions</a>
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='protocol')}}">Update Protocol</a>
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='version')}}">Version Format</a>
+          <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='category')}}">Update Category</a>
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='urls')}}">Source URLs</a>
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='style')}}">Style Guide</a>
           <a class="dropdown-item" href="{{url_for('.docs_composite')}}">Composite Devices</a>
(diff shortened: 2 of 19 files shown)


app/templates/docs-privacy.html
```diff
@@ -17,7 +17,7 @@ <h2>Introduction</h2>
   ensure that relevant compliance steps are addressed.
 </p>
 
-<h2>Scope</h2>
+<h2>Style</h2>
```
superm1: Probably an overzealous find and replace?

app/templates/private.html
```diff
@@ -112,6 +113,7 @@
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='restrict')}}">Adding Restrictions</a>
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='protocol')}}">Update Protocol</a>
           <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='version')}}">Version Format</a>
+          <a class="dropdown-item" href="{{url_for('.docs_metainfo', page='style2')}}">Update Style</a>
```
superm1: Maybe should this be in the Style guide specifically rather than it's own page?  If not, then I think Style guide probably should rename too.

app/templates/docs-privacy.html
hughsie: Ha, indeed!

app/views_upload.py
```diff
@@ -171,6 +171,15 @@ def _create_fw_from_uploaded_file(ufile):
             if pr:
                 md.protocol_id = pr.protocol_id
 
+        # allows OEM to specify style
+        if 'LVFS::UpdateStyle' in metadata:
+            sty = db.session.query(Style).filter(Style.value == metadata['LVFS::UpdateStyle']).first()
+            if sty:
+                md.style_id = sty.style_id
+        # fallback to device -- DROP THIS IN 2020
```
superm1: I think a prescriptive date is better than a year.  So maybe 1/2/2020 or something?
At some point as that gets closer it will be a good idea to flash a message too.

app/templates/private.html
hughsie: Yes, you're completely right. I did ponder just putting the style guide into https://fwupd.org/lvfs/docs/metainfo -- would that work?

app/views_upload.py
hughsie: > I think a prescriptive date is better than a year

Agreed, will fix.

> a good idea to flash a message too.

I think we do that if the user includes the name with the suffix already.

app/metadata.py
```diff
@@ -42,7 +42,8 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
         component = ET.SubElement(root, 'component')
         component.set('type', 'firmware')
         ET.SubElement(component, 'id').text = md.appstream_id
-        ET.SubElement(component, 'name').text = md.name
+        # until all front ends support LVFS::UpdateStyle append the suffix */
+        ET.SubElement(component, 'name').text = md.name_with_style
```
superm1: Well this is going to be tough.  You either need to reject really old frontends or keep two different strings in the metadata to stay compatible.  I'd say it's actually better to distribute two different strings.

hughsie: Well, the worst case scenario here is that we display "Hughski ColorHug" in the UI rather than "Hughski ColorHug Device Update" -- I didn't feel that was worth the extra pain.

superm1:
Sorry to bikeshed; but when you discussed with your colleague did you guys also discuss something like `ComponentType`?  Style to me has an inference that it's modifiable attribute of the device, but this is talking about strictly the fixed type of device that gets flashed.

hughsie:
> ComponentType

We had exactly this discussion. The `<component>` already has a type (`firmware`) and I think it would be confusing. I also really wanted a single word if possible.

app/metadata.py
superm1: OK

app/templates/private.html
superm1: Yeah that's a good place for it I think.

superm1:
Single word; `subtype`?

hughsie:
> Single word; subtype

Initially that worked, but GLib has subtle problems with things that end in `type` -- got any other ideas?

hughsie:
It appears _type is the problem, so maybe subtype works afterall. Maybe devtype works better, given its talking about the fwupd device as exposed through fwupd, not the release component...

superm1:
Especially when you map it back to a fwupd device, fwupd can have a concept of this devtype/subtype/class etc.

hughsie:
Although I see the logic to mapping back to the fwupd device (which solves other translating problems in various places), `fwup_device_get_devtype` would be weird, no? Maybe we want `fwup_device_get_kind` and `LVFS::UpdateKind` or `_get_kind` and `LVFS::DeviceKind`?

hughsie:
The other idea I had overnight was just use the generic AppStream way to tag components with enumerated things, for instance:

    <categories>
      <category>X-EmbeddedController</category>
      <category>X-Controller</category>
    </categories>

    or

    <categories>
      <category>X-ManagementEngine</category>
      <category>X-System</category>
    </categories>

This has the advantage that we can add new categories without "breaking API" -- the front end would just fall-back to the enumerated group that it does support, and carry translations for.

superm1:
Yes, I do like LVFS::DeviceKind or Appstream categories much better than style or scope.

hughsie:
I've respun things to use `<category>X-System</category>`

hughsie:
This pull request **introduces 1 alert** when merging 798d8080a1260b06c6b087b62685539a4eae5dc4 into 7b280156937976fadfece2eab0581a9754541634 - [view on LGTM.com](https://lgtm.com/projects/g/hughsie/lvfs-website/rev/pr-7369918b61c1ca1153a79e556fab83e863792fd3)

**new alerts:**

* 1 for Unreachable code

---

*Comment posted by [LGTM.com](https://lgtm.com)*

hughsie:
This pull request **introduces 1 alert** when merging edfe8b4cba0cf2aa2a74dcaa78e95ee74200d210 into 7ef94396304c5347d806ac26fb9c16c92abcb40b - [view on LGTM.com](https://lgtm.com/projects/g/hughsie/lvfs-website/rev/pr-b5157b23126e2423195742655df2628920635a0a)

**new alerts:**

* 1 for Unreachable code

---

*Comment posted by [LGTM.com](https://lgtm.com)*

PR #318, opened 2019-04-11
Make the firmware details page more visually engaging

hughsie:
Screenshot from 2019-04-11 13-23-56


hughsie:
Comments welcome; we also show cards for reports that match and if the firmware is embargoed from different countries.

hughsie:
I'm going to merge this to unstick another vendor (for the telemetry and reports performance improvements) but please tell me what you think of the firmware details page -- it still needs more work.

superm1:
The firmware reports page is obviously much more snappy.

I notice that it's wrapping funny for me.
This is from Chrome on Windows:
image

I notice that it's not obvious how to find the "failed" reports anymore on the reports tab (other than visiting $URL/analytics/reports/3)

hughsie:
> I notice that it's wrapping funny for me.

My first instinct was to break out `text-nowrap` but that's going to break in other more subtle ways. I'll have a think how I can get rid of that big expanse of table.

> I notice that it's not obvious how to find the "failed" reports anymore

I hid the tab is there were failed reports... Was this unexpected or is there a bug?

superm1:
> My first instinct was to break out text-nowrap but that's going to break in other more subtle ways. I'll have a think how I can get rid of that big expanse of table.

One thought is to just show the aggregate information and leave an CSV link to get the details.

> I hid the tab is there were failed reports... Was this unexpected or is there a bug?

I see that the link moved and it's only in details page.  I think it was just unexpected that it moved around to me.  No problem.

hughsie:
> One thought is to just show the aggregate information

I've improved the filtering quite a bit and deployed some new UI -- can you tell me if the new UI does what you need it to? I'm not against exporting a big table of CSV, but that's kinda the "we gave up" school of design :)

superm1:
Hmm https://fwupd.org/lvfs/telemetry/30 doesn't have anything now and I'm kinda confused what https://fwupd.org/lvfs/telemetry/0/success/down is supposed to be showing.  It seems like just a random hand ful of devices.

hughsie:
> I'm kinda confused what https://fwupd.org/lvfs/telemetry/0/success/down is supposed to be showing

I've fixed some of that in 3e3da35d02eeaf417e1078d5cc472422cf997142 -- I agree not showing what we're sorting by isn't easy to wrap your head around. I'll deploy when I have fixed the below.

> https://fwupd.org/lvfs/telemetry/30 doesn't have anything now 

Hmm, a bug. Will debug and fix.

hughsie:
> Will debug and fix

Can you try now? You can alter the 30 at the end of the URL if you need to make it go back a bit more.

superm1:
I guess i'm still really confused what i'm looking at.  the 30 link gets me one system and there were definitely downloads for more than just that system in the last 30 days.

For the "success" link, can you explain what that should be showing now too?  It still seems like an unobvious list of what "Success" means.

Also notice this:
image

hughsie:
> and there were definitely downloads for more than just that system in the last 30 days

It's only showing firmwares uploaded in the last 30 days.

> It still seems like an unobvious list of what "Success" means

Percentage of reports that were successful.

> Also notice this

Ya, I have to fix a whole load of those. Thanks.

superm1:
> It's only showing firmwares uploaded in the last 30 days.

Ah thanks...  I think that needs to be made clear somewhere that the 30 days represents 30 days of uploads.  Also I think it would be clearer to make it 30 days since promotion rather than upload as firmware might sit in embargo for a while in testing.  I definitely know of some right now that will be more than 30 days old before it's turned to stable.

> Percentage of reports that were successful.

Similarly I think some text to explain this will help.

PR #325, opened 2019-04-30
Ensure the firmware-flashed value is a valid lowercase GUID

hughsie:
Fixes https://github.com/hughsie/fwupd/issues/1103


superm1:
Code changes look good to me, but I believe you should also update documentation to mention that GUID needs to be lowercase.

hughsie:
> should also update documentation to mention that GUID needs to be lowercase

Added, thanks!

PR #326, opened 2019-05-02
Wip/hughsie/semver

hughsie:


hughsie:
Screenshot from 2019-05-02 17-26-39

hughsie:
Cool, thanks for reviewing. I did wonder if we should put the current deployed version in a custom.cfg.in type file, so that the footer would "know" the deployed version without me hardcoding it each time? The other thing we could do is show a toaster reminder to users if they log in and there's been a few changes. This would mean storing the last LVFS version in the user table and making the news page a little more interesting. I don't know if that's over-engineering things a bit tho.

superm1:
Could you maybe look at git describe for dynamic filling that out?

On May 3, 2019 02:43, Richard Hughes <notifications@github.com> wrote:

[EXTERNAL EMAIL]

Cool, thanks for reviewing. I did wonder if we should put the current deployed version in a custom.cfg.in type file, so that the footer would "know" the deployed version without me hardcoding it each time? The other thing we could do is show a toaster reminder to users if they log in and there's been a few changes. This would mean storing the last LVFS version in the user table and making the news page a little more interesting. I don't know if that's over-engineering things a bit tho.

-

PR #329, opened 2019-05-10
Use CHIPSEC to add component shards for UEFI firmware

hughsie:
This allows us to show what modules make up each firmware volume, and in the
future means we can do attestation and verification of the installed modules.

At the moment it's really useful to see what modules are shared between models
and vendors, and would allow us to construct a blocklist in the future for
known broken firmware.

### DIFF
--- a/plugins/chipsec/__init__.py
+++ b/plugins/chipsec/__init__.py
@@ -1,0 +1,146 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 Richard Hughes <richard@hughsie.com>
+# Licensed under the GNU General Public License Version 2
+#
+# pylint: disable=no-self-use
+
+import tempfile
+import glob
+import hashlib
+import os
+import re
+import shutil
+import subprocess
+
+from app import db
+from app.pluginloader import PluginBase, PluginError, PluginSettingBool, PluginSettingText
+from app.models import Test, ComponentShard, ComponentShardChecksum, ComponentShardInfo
+
+def _add_component_shards(md, files):
+
+    # remove any old shards
+    for shard in md.shards:
+        db.session.delete(shard)
+
+    # parse each EFI binary as a shard
+    for fn in files:
+        sections = fn.rsplit('/')
+        name = sections[-1].split('.')[0]
+        guid = None
+        for section in reversed(sections[:-1]):
+            if section.find('GUID_DEFINED') != -1:
+                continue
+            if section.find('COMPRESSION') != -1:
+                continue
+            guid = section.split('_')[1].split('.')[0].lower()
+            break
+        if not guid:
+            continue
+
+        shard = ComponentShard(component_id=md.component_id)
+        shard.info = db.session.query(ComponentShardInfo).\
+                            filter(ComponentShardInfo.guid == guid).first()
+        if shard.info:
+            shard.info.cnt += 1
+        else:
+            shard.info = ComponentShardInfo(guid, name)
+        with open(fn, 'rb') as f:
+            data = f.read()
+
+            # SHA1 is what's used by researchers, but considered broken
+            csum = ComponentShardChecksum(hashlib.sha1(data).hexdigest(), 'SHA1')
+            shard.checksums.append(csum)
+
+            # SHA256 is now the best we have
+            csum = ComponentShardChecksum(hashlib.sha256(data).hexdigest(), 'SHA256')
+            shard.checksums.append(csum)
+
+        # add shard to component
+        md.shards.append(shard)
+
+def _run_chipsec_on_blob(self, test, md):
+
+    # write firmware to temp file
+    cwd = tempfile.mkdtemp(prefix='lvfs')
+    src = tempfile.NamedTemporaryFile(mode='wb',
+                                      prefix='lvfs_',
+                                      suffix=".bin",
+                                      dir=cwd,
+                                      delete=False)
+    src.write(md.blob)
+    src.flush()
+
+    # log file output
+    log = tempfile.NamedTemporaryFile(mode='wb',
+                                      prefix='lvfs_',
+                                      suffix=".log",
+                                      dir=cwd,
+                                      delete=False)
+
+    # run chipsec
+    cmd = self.get_setting('chipsec_binary', required=True)
+    argv = [cmd, '--no_driver', '--log', log.name, 'uefi', 'decode', src.name]
+    ps = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=cwd)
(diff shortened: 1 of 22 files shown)


hughsie:
Screenshot from 2019-05-10 16-17-46

superm1:
> in the future means we can do attestation and verification of the installed modules

How?  You mean with a locally running chipsec instance on a machine?

plugins/chipsec/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def name(self):
+        return 'CHIPSEC'
+
+    def summary(self):
+        return 'Add firmware shards for UEFI capsules'
+
+    def settings(self):
+        s = []
+        s.append(PluginSettingBool('chipsec_enabled', 'Enabled', True))
```
superm1: As I can tell there isn't actually anything in this PR that uses this variable to enable/disable the plugin?

superm1:
So a Shard is a UEFI firmware volume?  Why the different nomenclature?

superm1:
When this is merged you'll probably want to PR to https://github.com/chipsec/chipsec to update the README to mention you're using it.

plugins/chipsec/__init__.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # decompress firmware
+        fn = _get_absolute_path(fw)
+        try:
+            istream = Gio.File.new_for_path(fn).read()
+        except GLib.Error as e: # pylint: disable=catching-non-exception
+            raise PluginError(e)
+        cfarchive = GCab.Cabinet.new()
+        cfarchive.load(istream)
+        cfarchive.extract(None)
```
superm1: I wonder if it's worth doing this automatically when the plugins launch rather  than in each plugin.  Otherwise you're going to be doing this at least twice for UEFI capsule now with the GUID checker.

plugins/chipsec/__init__.py
hughsie: It's magic in the plugin loader. The 'Enabled' string is special.

hughsie:
> Why the different nomenclature

This functionality isn't uefi specific. I was going to do the same for DfuSe files too. I'm not attached to 'shard' if you have better ideas.

plugins/chipsec/__init__.py
superm1: OK thanks.

plugins/chipsec/__init__.py
hughsie: Yes, I think you're right. Some archived firmware is kinda large. It'll need some rework of the loader. I'll think about this.

superm1:
>This functionality isn't uefi specific. I was going to do the same for DfuSe files too. I'm not attached to 'shard' if you have better ideas.

Ah OK, that makes sense then.  No concerns.

plugins/chipsec/__init__.py
hughsie: Okay, I think I've done this now. Feedback welcome!

hughsie:
> You mean with a locally running chipsec instance

Yes, although I've not worked out the details yet. There's a lot of work to do before this becomes possible.

hughsie:
This pull request **introduces 1 alert** when merging 5f84def322b22e8aa424872295e9caad1ec90a95 into 8ce20a0ea28cfcd37208736e392d091e5c54deb0 - [view on LGTM.com](https://lgtm.com/projects/g/hughsie/lvfs-website/rev/pr-5ea5796ab71fa470fbf695e11aae70ad8148f01c)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

superm1:
Lgtm

hughsie:
I've broken out the checksums into a new table; security researchers ironically use SHA1, but if I use that I get pinged by the security reviewers... So we just do both now...

hughsie:
Also, this information is now available to **public** users, and thus indexable by Google... It's no more information than users are able to get already using chipsec on the decompressed .cab file (hence the logic to making it public), but I didn't want to surprise anyone... The "diff" between versions I think is hugely useful.

Screenshot from 2019-05-12 09-13-45

plugins/chipsec/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+                                      suffix=".log",
+                                      dir=cwd,
+                                      delete=False)
+
+    # run chipsec
+    cmd = self.get_setting('chipsec_binary', required=True)
+    argv = [cmd, '--no_driver', '--log', log.name, 'uefi', 'decode', src.name]
+    ps = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=cwd)
+    if ps.wait() != 0:
+        raise PluginError('Failed to decode file: %s' % ps.stderr.read())
```
superm1: So this is going to make chipsec passing extraction mandatory, no?
Right now Dell uses a compressed file within binary but if that switched to a new method chipsec would fail on Dell binaries, and this test fails.

And what if a vendor uses a format unsupported by  chipsec in general?

I think my preference here is that this not be a requirement to promote but just FYI type thing.

plugins/chipsec/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # look for shards
+    outdir = src.name + '.dir'
+    files = glob.glob(outdir + '/FV/**/*.efi', recursive=True)
+    _add_component_shards(md, files)
```
superm1: If you don't find files don't you want to log that? As in chipsec had 0 return code but nothing found.

plugins/chipsec/__init__.py
hughsie: > So this is going to make chipsec passing extraction mandatory, no?

No, but I agree it's not obvious. If `uefi decode` fails to find any volumes, chipsec exits with 0 (which arguably is a bug in itself) and we just don't add any shards. I've reported https://github.com/chipsec/chipsec/issues/558 so we can get a more sensible error code in this case.

plugins/chipsec/__init__.py
hughsie: Agreed, good idea. I don't want to fail this even on no FV's found as there's too much firmware that fails to parse -- when chipsec is python3 I'll be becoming a thorn in their side a bit more.

hughsie:
This pull request **introduces 1 alert** when merging 436e244755fede44562e546490351872042f1d02 into 8ce20a0ea28cfcd37208736e392d091e5c54deb0 - [view on LGTM.com](https://lgtm.com/projects/g/hughsie/lvfs-website/rev/pr-a8a602e29825ce4644397f6dcb4443b7c7829ca9)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

PR #334, opened 2019-05-21
Wip/hughsie/shards

hughsie:


hughsie:
Screenshot from 2019-05-21 15-02-57

app/models.py
```diff
@@ -978,6 +989,11 @@ def __init__(self, value, kind='SHA1'):
     def __repr__(self):
         return "ComponentShardChecksum object %s(%s)" % (self.kind, self.value)
 
+def _calculate_entropy(s):
+    probabilities = [n_x / len(s) for x, n_x in collections.Counter(s).items()]
+    e_x = [- p_x * math.log(p_x, 2) for p_x in probabilities]
+    return sum(e_x)
```
superm1: I'm not really sure why this is valuable to anyone. Care to comment?

plugins/intelme/__init__.py
```diff
@@ -0,0 +1,242 @@
...
+        # only until required
+        return True
+    if md.protocol.value != 'org.uefi.capsule':
+        return False
+
+    # match on category
+    if not md.category:
+        # only until required
+        return True
+    return md.category.matches(['X-System', 'X-ManagementEngine'])
```
superm1: Didn't you have some other new ones that got added recently that should match here too?

hughsie: Sure, they both have `X-ManagementEngine` as a fallback which means `category.matches()` returns True, I think.

app/models.py
hughsie: It's really only useful to me. For some segments that are nearly all `0xff`'s they have really low entropy. This is the case of the empty MFS for example. We can also see modules that switched from being uncompressed to compressed as the entropy goes way up.

superm1: Thanks for explaining.

Issue #335, opened 2019-05-21
LVFS XML file is failing on ancient fwupd


superm1:
Ubuntu 16.x still have fwupd 0.8.3 and recently there was an issue reported where XML stopped parsing properly.  The issue doesn't exist in newer fwupd versions.

```
#fwupdmgr refresh
failed to update metadata: Failed to parse XML: Mismatched XML
```

I looked at the XML with `appstream-util validate` and noticed this but don't know if it's related:
```
• markup-invalid        : com.lenovo.ThinkStationP520/P520c.firmware: <id> has invalid character [/]
```

The issue was reported here: https://bugs.launchpad.net/ubuntu/+source/appstream-glib/+bug/1829813

hughsie:
Is it weird that `appstream-util dump ~/.cache/fwupd/lvfs-metadata.xml.gz` still works -- in 0.8.x it's parsing the XML with the same `AsNode` functionality.

hughsie:
Hah, that's not it. The LVFS has got too successful:

    gzip -d /home/hughsie/.cache/fwupd/lvfs-metadata.xml.gz
    du -h /home/hughsie/.cache/fwupd/lvfs-metadata.xml
    1.1Mb

and of course, the limit in fu-main.c post decompression is `0x100000` (1.0Mb)

Two ways to solve this -- prune the metadata a bit and possibly remove some cruft (and then the problem will only come back later...) or raise the limit in `fu-main.c` to something "huge" like 10Mb. I'd be fine for a patch against the old branch if that helps.

superm1:
Hah.  Good find!
Well how about both?  I think pruning some really old stuff will help the immediate term, and then a patch to ancient branch will have some time to get deployed through various processes.

PR #336, opened 2019-05-21
Only include the latest 5 releases for a component

hughsie:
This reduces the uncompressed metadata size from 1.1Mb to 916Kb which fixes
https://github.com/hughsie/lvfs-website/issues/335 for a little while.


superm1:
Looks good to me, lets plan to revert this later.

PR #338, opened 2019-05-23
Wip/hughsie/chipsec

hughsie:


plugins/chipsec/__init__.py
```diff
@@ -100,18 +100,21 @@ def settings(self):
         s.append(PluginSettingText('chipsec_binary', 'CHIPSEC executable', 'chipsec_util'))
         return s
 
-    def ensure_test_for_fw(self, fw):
+    def _require_test_for_md(self, md):
+        if not md.protocol:
```
superm1: Even if you've got a local instance of the database and files shouldn't you always have a protocol set?

hughsie: The `Protocol:protocol_id` is `nullable=False` and some older `Components` have no `protocol_id` set (with a `NULL` default) --0 so if you don't guard against this you get:

      File "plugins/chipsec/__init__.py", line 106, in _require_test_for_md
        return md.protocol.value == 'org.uefi.capsule'
    AttributeError: 'NoneType' object has no attribute 'value'

superm1: Gotcha okay

PR #340, opened 2019-05-28
Wip/hughsie/pefile

hughsie:


plugins/pecheck/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+        for cert in certs:
+            if cert.not_before and cert.not_before > dtnow:
+                test.add_fail('Certificate Invalid',
+                              '{}: {} is invalid before {}'.format(shard.info.name, cert.desc, cert.not_before))
+            elif cert.not_after and cert.not_after < dtnow:
+                test.add_fail('Certificate Invalid',
+                              '{}: {} is invalid after {}'.format(shard.info.name, cert.desc, cert.not_after))
+            else:
+                test.add_fail('Certificate',
+                              '{}: {}'.format(shard.info.name, cert.desc))
```
superm1: Don't you want to add a pass in the situation that you had certs and no failures?

hughsie: That's handled above:

        if not certs:
            test.add_pass('Scanned', 'No certificates')
            return

superm1: That handles the no certificates case - but I'm saying specifically the case you had certs you checked and didn't find failures on them.

plugins/pecheck/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+                              'Extracted: {} bytes, expected: {} bytes'.format(len(signature), security.Size - 8))
+                return
+        except pefile.PEFormatError as _:
+            # not a PE file, which is fine
+            return
+
+        # get all the certificates and signer
+        certs = _extract_certs_from_authenticode_blob(signature)
+        if not certs:
+            test.add_pass('Scanned', 'No certificates')
```
superm1: Since this is likely to run on 10's or 100s of .EFI files, you probably will want to have put the `shard.info.name` in the message.

plugins/pecheck/__init__.py
hughsie: Aha! good catch, that last `add_fail` is supposed to be `add_pass` -- thanks!

plugins/pecheck/__init__.py
hughsie: Good idea, added.

PR #341, opened 2019-05-31
Wip/hughsie/save shards

hughsie:


plugins/pecheck/__init__.py
```diff
@@ -118,6 +118,7 @@ def order_after(self):
     def settings(self):
         s = []
         s.append(PluginSettingBool('pecheck_enabled', 'Enabled', True))
+        s.append(PluginSettingInteger('pecheck_allowable', 'Number of years to relax failure', 3))
```
superm1: If you have intention to change this in the future, it is probably worth flagging things that are within "margin" right now as an explicit pass, or flash a warning or something similar.

plugins/chipsec/__init__.py
```diff
@@ -103,6 +100,7 @@ def summary(self):
     def settings(self):
         s = []
         s.append(PluginSettingBool('chipsec_enabled', 'Enabled', True))
+        s.append(PluginSettingBool('chipsec_write_shards', 'Write shards to disk', True))
```
superm1: I guess it's OK that all uploads will essentially take 2x the size by default with this?

plugins/pecheck/__init__.py
superm1: Also, this probably needs to be mentioned explicitly in news that this behavior has changed

plugins/chipsec/__init__.py
superm1: Yes, I think deduping w/ hardlinks is probably a good idea where possible for this.

>Note, I don't think the Dell firmware is extractable using chipsec uefi decode -n foo.bin yet

Right, didn't think so.  Payloads are buried within that binary.

hughsie: >  Payloads are buried within that binary

Do you know how chipsec would go about decoding the Dell binaries? Any docs about how the Dell binary is structured?

superm1: There are certainly some reverse engineering efforts that work on _some_ of the binaries.
However there is currently no public documentation about how it's structured.

plugins/pecheck/__init__.py
hughsie: > probably needs to be mentioned explicitly in news

Fixed, thanks.

hughsie: > things that are within "margin" right now

I'm talking with some security teams right now to find out how bad (or unimportant) expired Authenticode certs are. If the conclusion it's unforgivable, I'll dial this right back.

plugins/chipsec/__init__.py
hughsie: Okay, I think I've got a workaround.

PR #345, opened 2019-06-06
Automatically demote firmware with >5 failures and a success rate of <70%

hughsie:
This was requested by Intel.

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -12,13 +12,16 @@ import sys
 import hashlib
 import datetime
 
+from flask import render_template
+
 from cabarchive import CabArchive
 
 from lvfs import app, db, ploader
 from lvfs.dbutils import _execute_count_star
+from lvfs.emails import send_email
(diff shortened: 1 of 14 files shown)


hughsie:
Screenshot from 2019-06-06 12-12-07

superm1:
In principle this seems like a really sensible thing to do.  I do like that it would only be marked for failures, so it would encourage marking known issues if there are known issues.  I have to start to worry though that this could cause a bad actor to "game" the automatic downgrade system.   If they could make 50 bad reports automatically generated with random machine IDs be uploaded from a script that would trigger this, wouldn't it?

hughsie:
> this could cause a bad actor to "game" the automatic downgrade system

It could, yes. I do think that's somewhat unlikely, and if that happened we could certainly put in limit on the client IP address or even disable it for specific firmwares were required.

hughsie:
Lets give this a go, and if it gets abused we can turn it off easily.

superm1:
Sounds good.

PR #357, opened 2019-06-21
Report invalid <name> values as an upload error

hughsie:
Vendors are just ignoring the nag message to use <cagegory> instead.


hughsie:
@superm1 this won't be applied until you can give the thumbs up -- a lot of the Dell firmware gets a warning message like this `Fixed component name from "Latitude 5X01/Precision 3541 System Update" to "Latitude 5X01/Precision 3541"` which obviously can't be seen my the robot uploader. I think some work on the template is needed.

p.s. I'm don't want a response until you're back from leave. :)

superm1:
This is going to break flows for various teams on older platforms that have sustaining firmware releases being developed.  Those firmwares are built by automation and pushing build process changes is quite painful.  It's likely that these firmwares would have to be hand modified (ie something like `gcab extract; sed $STUFF; gcab create` or it's similar corollary on `makecab.exe`)

Could you just make the message silent when it's adjusted instead?

hughsie:
Could we relax the new rule for robot uploaders?

superm1:
> Could we relax the new rule for robot uploaders?

They're real people uploading, but the firmware is `built` by automation and only uploaded after verification is complete and it's close to real public release.

hughsie:
> pushing build process changes is quite painful

Do you have an EOL for these platforms?

superm1:
> Do you have an EOL for these platforms?

The typical sustaining lifecycle is in the order of multiple years for certain brands.  But I don't have that information to share on a platform by platform basis.  The tooling team told me that various platforms have the build automation tooling forked at a moment in time and then managed by the individual platform teams.  We're talking about updating at least ~50 sets of forked tools with the appropriate number of teams that manage them.  This really is not an option for them.

Let me ask this - why is it a problem for LVFS to just keep silently change the titles?  Technical debt?

hughsie:
>  This really is not an option for them.

Okay, fair enough.

> Let me ask this - why is it a problem for LVFS to just keep silently change the titles? Technical debt?

It because new vendors and models are being added all the time and uploaders are just ignoring the warnings. If we remove the text in the appstream metadata we don't edit the metainfo.xml file, so when we want to turn on the client side translations we'll get something like "foo System Update System Update" when the user double clicks on the standalone .cab file.

Of course, that failure mode isn't worth all the extra work it sounds like it would be for you.

What about, we relax the check if there is even one existing component with the same AppStream ID? Would that work? Or perhaps if the vendor has more than 25 existing uploads?

superm1:
> What about, we relax the check if there is even one existing component with the same AppStream ID? Would that work? Or perhaps if the vendor has more than 25 existing uploads?

The exact timing of in development platforms that already locked tools but don't have firmware on LVFS will cause some fringe cases for the former.  I think the latter would definitely help.

hughsie:
>  I think the latter would definitely help.

What about this version? ^^^

superm1:
Looks fine by me now

PR #363, opened 2019-07-16
Add notes on using <firmware>GUID</firmware>

hughsie:


lvfs/templates/docs-metainfo-restrict.html
```diff
@@ -63,6 +63,25 @@
...
+<div class="card mt-3">
+  <div class="card-body">
+    <div class="card-title"><a id="requires-fw-versions"></a>Restricting using other firmware version</div>
+<p class="card-text">
+  Newer versions of fwupd can restrict updates on one device depending if another
+  firmware GUID exists on the system of any version.
+  This is similar to the CHID method above but uses the GUID of the firmware,
+  not a hardware ID.
+</p>
+<pre>
```
superm1: I think it might be useful to mention an example of _why_ this would be used.  I naively think of the scenario of you have something like a logitech keyboard that connects to the USB dongle but the keyboard firmware can only install if the dongle is updated to some version of the firmware that starts exporting a GUID.  Maybe i'm wrong :)

PR #369, opened 2019-07-31
Show the user capabilities in the headerbar

hughsie:
I got asked today how users are supposed to know what user level they are.

I think it's better to show in the header next to the 'person' rather than in
the profile view, the latter being where things get changed.


hughsie:
Limited upload-only user:

Screenshot_2019-07-31 LVFS Home

All powerful user:

Screenshot_2019-07-31 LVFS Home(1)

superm1:
sure LGTM

PR #370, opened 2019-08-04
Allow firmware or vendors to enable DoNotTrack functionality

hughsie:
This allows either entire vendors to globally opt out of the download and
success reporting, or restrict it from specific firmware objects. Enabling the
LVFS::DoNotTrack feature disables a lot of the useful vendor functionality
(including the per-firmware limits and auto-demotion functionality) and is not
recommended by the LVFS admins.

### DIFF
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -383,6 +383,7 @@ class Vendor(db.Model):
     version_format = Column(String(10), default=None) # usually 'triplet' or 'quad'
     url = Column(Text, default=None)
     banned_country_codes = Column(Text, default=None) # ISO 3166, delimiter ','
+    do_not_track = Column(Boolean, default=False)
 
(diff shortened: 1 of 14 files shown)


hughsie:
Screenshot_2019-08-04 LVFS Firmware Details

superm1:
The PR looks fine, but given the impact of this what do you think of having a warning before enabling it about the not immediate obvious impacts?

hughsie:
You mean at upload time? E.g. "Firmware analytics and reports will not be collected" in an alert box?

superm1:
Yeah.

hughsie:
What about this?

Screenshot_2019-08-06 LVFS Upload

Screenshot_2019-08-06 LVFS Dashboard

superm1:
Yeah that looks good

PR #372, opened 2019-08-05
Wip/hughsie/better errors

hughsie:


lvfs/templates/error.html
```diff
@@ -8,4 +8,18 @@
...
 </div>
+
+        <div class="row flex-center min-vh-100 py-6 text-center">
+          <div class="col-sm-10 col-md-8 col-lg-6 col-xxl-5"><a class="d-flex flex-center mb-4" href="../../index.html"><img class="mr-2" src="../../assets/img/illustrations/falcon.png" alt="" width="58" /><span class="text-sans-serif font-weight-extra-bold fs-5 d-inline-block">falcon</span></a>
+            <div class="card">
+              <div class="card-body p-4 p-sm-5">
+                <div class="display-1 text-200 fs-error">404</div>
+                <p class="lead mt-4 text-800 text-sans-serif font-weight-semi-bold">The page you're looking for is not found.</p>
+                <hr />
+                <p>Make sure the address is correct and that the page hasn't moved. If you think this is a mistake, <a href="mailto:info@exmaple.com">contact us</a>.</p><a class="btn btn-primary btn-sm mt-3" href="../../index.html"><span class="fas fa-home mr-2"></span>Take me home</a>
```
superm1: wrong domain

lvfs/views.py
```diff
@@ -212,17 +212,17 @@ def format_html_from_markdown(tmp):
 
 @lm.unauthorized_handler
 def unauthorized():
-    msg = ''
+    msg = 'Permission denied: '
```
superm1: does this actually need the whitespace?

lvfs/templates/index.html
```diff
@@ -17,8 +17,13 @@
...
   </div>
-  <div class="container">
-    <div class="row justify-content-center align-items-center pt-6 pt-lg-8 pb-lg-9 pb-xl-0">
+  <div class="container pt-6 pt-lg-8">
+{% for category, msg in get_flashed_messages(with_categories=true) %}
+    <div class="alert alert-{{ category }}" role="alert">
+      {{ msg }}
+    </div>
+{% endfor %}
+    <div class="row justify-content-center align-items-center pb-lg-9 pb-xl-0">
```
superm1: Since this is on the landing page, just want to confirm there isn't a large performance impact.  It pretty much means that `wget http://fwupd.org` will run this code.

lvfs/templates/error.html
hughsie: Eeek, good catch, thanks!

lvfs/views.py
hughsie: It's in a funky place, indeed.

lvfs/templates/index.html
hughsie: Good question. The flash messages are already being processed (in the Request) and we're actually showing them already in the `<!-- -->` section to make the self tests possible. tl;dr: no performance impact.

superm1:
LGTM

PR #375, opened 2019-08-12
Wip/hughsie/disable inf

hughsie:


superm1:
Don't some older fwupd versions also parse INF? If so it's probably worth implicitly setting minimum fwupd version when this flag is turned on (although document this).

Otherwise lgtm

hughsie:
> Don't some older fwupd versions also parse INF

Ohh hmm, maybe. I couldn't see anything in the NEWS file for fwupd although I know I wrote an inf parser for libappstream-glib. I can't find anything in the fwupd git history about .inf files, so maybe it was only really used for appstream-util validate?

superm1:
Yeah, I think it was specifically when using appstream-glib w/ fwupd.  If you just set it to have a minimum fwupd version to the one after appstream-glib was dropped that's probably safe.

hughsie:
It was <0.7.3 so a pretty long time ago. I added some notes in the docs, feedback very welcome, thanks.

superm1:
LGTM

superm1:
Side note; anything that old still connecting recently?  Another approach is to indicate minimum version to use LVFS to be 0.8.x or newer.

hughsie:
I think fwupd only started using its version in the user agent in 0.8.x, and yes we still have thousands of very old clients connecting :(

PR #377, opened 2019-08-12
Wip/hughsie/vendor oem

hughsie:


hughsie:
Screenshot_2019-08-12 LVFS Upload

superm1:
This really confuses the permissions model to me.  Let's say you have ODM XYZ that does OEM work for ABC and DEF.  For ABC they only have permissions to upload into ABC's account, but ABC does the final promotion to stable and manages lifecycle.  If there is a problem, ABC is notified, pulls the firmware out and XYZ works on the fix.

For DEF you want to have them do everything end to end.   What does that mean their individual user settings look like?  How do you decide which OEMs they're allowed to represent?  How can OEM get involved if there is a problem?

hughsie:
> This really confuses the permissions model to me

Yes, it needs work. The missing bit I think is adding permissions booleans to the `Affiliation` object, e.g. we specify what the ODM can do on behalf of the OEM, e.g.

    Boolean allow_promote_stable

And then we assume all ODM affiliates can upload for and transfer ownership to the OEM. At the moment giving one shared ODM permission to push to stable would allow it to push any of it's uploaded firmwares to stable, even if the other OEM would rather do the QA itself.

>  How can OEM get involved if there is a problem

They can't -- the OEM can't see any of the ODMs firmware if it's still owned by the ODM, which is somewhat suboptimal I suppose. I guess we want the OEM to have access to the firmware uploaded and managed by the ODM in the embargo remote too.

I think this patch is on the right lines in that we need two vendors insight into the firmware, the OEM group and the ODM group. So I think that just having the vendor_id_oem be used _for display only_ seems a mistake perhaps.

hughsie:
I think an easier to understand way to do this would be to define a write-once `Vendor.vendor_upload` and always allow QA users in the `vendor_upload` group to view the firmware, and only allow pushing to stable as an affiliate QA user if `allow_promote_stable` is set as a condition of the affiliation. I'll prototype that too.

superm1:
> Boolean allow_promote_stable

Yeah I think that's the right direction.

> So I think that just having the vendor_id_oem be used for display only seems a mistake perhaps.

Yes, I think it definitely blurs the lines of ownership.

GavinChengGao:
Hi guys, thanks for working on this. Yes, as OEM we need to audit the LVFS packages in at least embargo/testing/stable states. In addition, in case of demotion by ODM, OEM should also be notified.

superm1:
> Hi guys, thanks for working on this. Yes, as OEM we need to audit the LVFS packages in at least embargo/testing/stable states. In addition, in case of demotion by ODM, OEM should also be notified.

I've had some discussion internally at my company as well and there is some appetite for this type of change as well.

hughsie:
The more I worked on this, the more complex it became. I'll close this now and open a different PR which is much simpler.

PR #387, opened 2019-09-04
Allow the OEM to control what the ODM is able to do

hughsie:
To enforce this, we also need to add the concept of vendor namespaces.
The namespace prefix is the AppStream ID prefix that can be displayed to the
user. The OEM is the 'owner' of the ID prefix, and never the ODM. Add a
firmware 'problems' for existing firmware on the LVFS that hasn't already been
pushed to stable.

This prevents the ODM being 'credited' with the hardware when they are just
managing the entire end-to-end firmware lifecycle for the OEM as part of a
contract. Adding the actions to an ACL list allows us to better explain to the
vendor manager what the ODM is able to do, without hardcoding the logic.

Additionally, add an optional notfication email for ODM or OEM upload so that
the OEM is aware of what the ODM is doing.

Fixes https://github.com/fwupd/lvfs-website/issues/380

### DIFF
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -87,6 +87,10 @@ class Problem:
             return 'Firmware tests are pending'
         if self.kind == 'no-source':
             return 'No source code link'
+        if self.kind == 'no-vendor-namespace':
+            return 'No vendor namespaces set'
+        if self.kind == 'invalid-vendor-namespace':
+            return 'Invalid vendor namespace'
         return 'Unknown problem %s' % self.kind
 
     @property
@@ -359,6 +363,24 @@ class Namespace(db.Model):
     def __repr__(self):
         return '<Namespace {}>'.format(self.value)
 
+class AffiliationAction(db.Model):
+
+    # database
+    __tablename__ = 'affiliation_actions'
+    __table_args__ = {'mysql_character_set': 'utf8mb4'}
+
+    affiliation_action_id = Column(Integer, primary_key=True, unique=True, nullable=False)
+    affiliation_id = Column(Integer, ForeignKey('affiliations.affiliation_id'), nullable=False)
+    ctime = Column(DateTime, nullable=False, default=datetime.datetime.utcnow)
+    user_id = Column(Integer, ForeignKey('users.user_id'), nullable=False)
+    action = Column(Text, default=None)
+
+    user = relationship('User', foreign_keys=[user_id])
+    affiliation = relationship('Affiliation', foreign_keys=[affiliation_id])
+
+    def __repr__(self):
+        return "<AffiliationAction {}>".format(self.action)
+
 class Affiliation(db.Model):
 
     # database
@@ -373,10 +395,17 @@ class Affiliation(db.Model):
     # link using foreign keys
     vendor = relationship("Vendor", foreign_keys=[vendor_id], back_populates="affiliations")
     vendor_odm = relationship("Vendor", foreign_keys=[vendor_id_odm])
+    actions = relationship("AffiliationAction", cascade='all,delete-orphan')
 
     def __init__(self, vendor_id, vendor_id_odm):
         self.vendor_id = vendor_id
         self.vendor_id_odm = vendor_id_odm
+
+    def get_action(self, action):
+        for act in self.actions:
+            if action == act.action:
+                return act
+        return None
 
     def __repr__(self):
         return "Affiliation object %s" % self.affiliation_id
@@ -555,9 +584,11 @@ class Vendor(db.Model):
         if action == '@modify-oauth':
             return False
         if action == '@view-affiliations':
-            return user.is_qa
+            return user.is_vendor_manager
         if action == '@modify-affiliations':
             return False
+        if action == '@modify-affiliation-actions':
+            return user.is_vendor_manager
         if action == '@view-exports':
             return user.is_qa or user.is_vendor_manager
         if action == '@modify-exports':
@@ -1404,6 +1435,23 @@ class Component(db.Model):
                                   page='update')
             problems.append(problem)
 
+        # the OEM doesn't manage this namespace
+        values = [ns.value for ns in self.fw.vendor_odm.namespaces]
+        if not values:
+            problem = Problem('no-vendor-namespace',
+                              'No AppStream namespace values for '
+                              'vendor {}'.format(self.fw.vendor_odm.group_id))
+            problems.append(problem)
+        elif self.appstream_id_prefix not in values:
+            problem = Problem('invalid-vendor-namespace',
+                              'Component ID {} not allowed '
+                              'for vendor {}: {}'.format(self.appstream_id_prefix,
+                                                         self.fw.vendor_odm.group_id,
+                                                         ','.join(values)))
+            problem.url = url_for('.firmware_affiliation',
+                                  firmware_id=self.fw.firmware_id)
+            problems.append(problem)
+
         # set the URL for the component
         for problem in problems:
             if problem.url:
@@ -1456,15 +1504,11 @@ class Component(db.Model):
         # depends on the action requested
         if action == '@modify-updateinfo':
             if not self.fw.remote.is_public:
-                if user.is_qa and self.fw._is_vendor(user):
-                    return True
-                if self.fw._is_owner(user):
+                if user.is_qa and self.fw._is_permitted_action(action, user):
                     return True
             return False
         if action in ('@modify-keywords', '@modify-requirements', '@modify-checksums'):
-            if user.is_qa and self.fw._is_vendor(user):
-                return True
-            if self.fw._is_owner(user) and not self.fw.remote.is_public:
+            if user.is_qa and self.fw._is_permitted_action(action, user):
                 return True
             return False
         raise NotImplementedError('unknown security check action: %s:%s' % (self, action))
@@ -1647,6 +1691,12 @@ class Firmware(db.Model):
     remote = relationship('Remote', foreign_keys=[remote_id], lazy='joined')
 
     @property
+    def vendor_odm(self):
+        if not self.user:
+            return None
+        return self.user.vendor
+
+    @property
     def target_duration(self):
         if not self.events:
             return 0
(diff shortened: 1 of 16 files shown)


hughsie:
Screenshot_2019-09-04 LVFS Vendor Affiliations(1)

hughsie:
Screenshot_2019-09-04 LVFS Vendor Affiliations(2)

lvfs/models.py
```diff
@@ -1369,6 +1433,23 @@ def problems(self):
...
+            problem = Problem('no-vendor-namespace',
+                              'No AppStream namespace values for '
+                              'vendor {}'.format(self.fw.vendor_odm.group_id))
+            problems.append(problem)
+        elif self.appstream_id_prefix not in values:
+            problem = Problem('invalid-vendor-namespace',
+                              'Component ID {} not allowed '
+                              'for vendor {}: {}'.format(self.appstream_id_prefix,
+                                                         self.fw.vendor_odm.group_id,
+                                                         ','.join(values)))
```
superm1: I just want to make sure i'm following this correctly.  So basically in the XML say you have `com.dell` as your appstream prefix and Compal uploads some firmware.  As long as Compal has `com.dell` in their vendor_odm.namespaces they'll be able to do the actions that have been granted to them, correct?

lvfs/models.py
```diff
@@ -1791,10 +1874,32 @@ def _ensure_blobs(self):
...
         return self.vendor_id == user.vendor_id
 
+    def _is_odm(self, user):
+        return self.vendor_odm.vendor_id == user.vendor_id
+
     def mark_dirty(self):
         self.is_dirty = True
         self.remote.is_dirty = True
 
+    def _is_vendor_or_permitted_odm_action(self, action, user):
```
superm1: just a nit, how about maybe `is_permitted_action` and then leave the details to why it's permitted to the function rather than the title?

lvfs/templates/email-firmware-uploaded.txt
```diff
@@ -0,0 +1,15 @@
...
+A firmware has been uploaded to the Linux Vendor Firmware Service by {{user_upload.username}} ({{user_upload.vendor.group_id}}):
+
+{% for md in fw.mds %}
+ * {{md.name_with_category}} version {{md.version_display}}
+{% endfor %}
+
+You can log into the LVFS and view the firmware: {{url_for('.firmware_analytics_reports', firmware_id=fw.firmware_id, state=3, _external=True)}}
+
+If you believe this action should not have been allowed please contact the LVFS administrator by forwarding the email to info@fwupd.org with details.
+
```
superm1: I think this template is great.  I think that a similar template for promotion/demotion actions would be quite interesting too though.

lvfs/models.py
hughsie: >  As long as Compal has com.dell in their vendor_odm.namespaces they'll be able to do the actions

Not quite. What this is saying is that only Dell would have the com.dell id prefix, so the firmware vendor has to be assigned to dell and **never** to compal. This means we'd never refer to the firmware as a "Compal XPS 13" device which could happen in theory now if compal pushed to stable without changing the vendor to dell beforehand. What this patch does is force the ODMs to assign the correct OEM as the vendor, and then uses the user_id which is the uploader to work out the vendor group of the uploader, which is used for the odm fallback. I think it's 100x more clear than the old version where you had three things, the vendor_oem, vendor_odm and the user_uploader.

lvfs/templates/email-firmware-uploaded.txt
hughsie: Agreed. The User table is becoming a bit of a huge boolean mess so I thought I would clean that up at some stage, possibly with a 1:1 helper table just for norifications, or perhaps switch them to string enums so I don't keep having to do db changes when adding email notification types.

lvfs/models.py
hughsie: You mean just rename the function, or move the `_is_vendor()` check back to the `if` in the caller?

superm1: Just personal preference was leaning on renaming it rather than putting the logic in the title.

lvfs/models.py
superm1: OK, the most important thing to me was making sure that Compal didn't have to change the XML in some way, it's literally just a form upload action and database matching then.

superm1:
From what I follow this looks fine to me.  The various permissions that exist, I assume those will all default to not enabled and then will just turn on the individual ones as needed.

hughsie:
> I assume those will all default to not enabled and then will just turn on the individual ones as needed.

For new affiliations, exactly that. For existing affiliations I used: `@add-limit,@delete,@modify,@undelete,@modify-updateinfo,@view` -- is that sane?

Also, do you think vendor managers should be able to control this themselves, or leave this to the admin users?

lvfs/models.py
hughsie: Done.

lvfs/models.py
hughsie: Right, on the upload form they'd just chose Dell rather than Dell or Compal. If they chose Compal they'd get a "problem" appear, and then be able to transfer the firmware to Dell to remove the problem.

superm1:
> add-limit

I don't think this should default for existing ones.  If there is a company using it, they should share it with their affiliations only.

>remove-limit

I think that you may just want to merge these to `modify-limit` so you don't accidentally end up in a situation that an ODM turns on a limit while twisting knobs to see what they do and can't turn it back off.

> modify-keywords, modify-requirements, modify-checksums

I'd think this should default to yes for existing.

>Also, do you think vendor managers should be able to control this themselves, or leave this to the admin users?

Well affiliation has to be set by admin initially right?  I think that vendor manager should be able to at least see read-only see affiliation permission details, but don't see a problem with them adjusting the write permissions too.

hughsie:
>> add-limit
> I don't think this should default for existing ones

Agreed, removed.

> I think that you may just want to merge these to modify-limit

Done.

> Well affiliation has to be set by admin initially right

Yes, as the ODM might not be "public" on the LVFS yet, if ever. The vendors using the LVFS is kinda secret at this point.

> but don't see a problem with them adjusting the write permissions too.

I'll work on that this morning.

hughsie:
> I'll work on that this morning.

Done, I think.

superm1:
LGTM now.

hughsie:
I was thinking of sending something like this to lvfs-announce, ideas welcome:

Planned new functionality for managing the ODM and OEM relationship

At the moment the LVFS has lots of hardcoded assumptions that the ODM uploads the firmware, performs some testing in embargo and then transfers ownership to the OEM which add update details, and moves the firmware to testing and stable. This doesn’t match reality for a number of large vendors on the LVFS (for example, where the ODM manages all parts of the update flow), and so we needed to do something better going forward. There is also the problem where the ODM is being “credited” with the firmware upload, when we really don’t want to leak out the ODM→OEM relationship even accidentally.

To do this I’ve changed slightly how we assign the firmware on upload. Each OEM has a set of appstream prefixes that match the <id> values uploaded to the LVFS. The idea is that, for instance, Compal would never own firmware with com.dell.* and only the Dell account can do that. This would stop the ODM accidentally owning firmware that should be credited to the OEM.

The user who uploaded the firmware is part of a vendor group, and this is how we get the ODM vendor information. If an affiliation has been defined between the OEM and ODM then the ODM can manage the firmware on the LVFS. We can now define what “manage” means, with the most simple permissions being “just view” and the most complete permissions being “view, modify, delete, move to testing, move to stable”.

For all the vendors on the LVFS I will assign what I think are the correct vendor prefixes, based on the existing values uploaded, and for vendors with no uploaded firmware a prefixed based on the reverse DNS of the OEM homepage. For instance, the Dell account will have ‘com.dell’ assigned. OEMs can have multiple vendor prefixes assigned where required, and the vendor prefix doesn’t have to match the domain name in every case.

There are a lot of existing ODM<->OEM vendor relationships on the LVFS. I’ve converted the existing relationships to be a safe subset of the existing permissions on the affiliate relationship. For vendors who are currently managing the end-to-end flow the OEM vendor manager will need to add extra allowed actions for the ODM. Either ask your OEM vendor manager to do this, or me.

I’m aware this changes the way some ODMs are using the LVFS right now, and I’m happy to help both OEMs and ODMs choose actions that reflects the current business relationship. There might be firmware files that I’ve not fixed, or where affiliations need to be added or modified, and please email me if you’re suddenly unable to do an action that you expected to be able to do.

Existing firmware in testing and stable isn’t affected by the new rules, and this will only affect new firmware uploaded. If anyone has any concerns please say. I’m confident the new functionality will allow us to grow the LVFS and also be suitable for the changing landscape of ODM and OEM relationsips.

Richard.

superm1:
> This doesn’t match reality for a number of large vendors on the LVFS (for example, where the ODM manages all parts of the update flow), and so we needed to do something better going forward

This doesn't match the _desired_ flow.  I'd think that no ODMs should be able to promote OEM firmware to stable until this.

> To do this I’ve changed slightly how we assign the firmware on upload. Each OEM has a set of appstream prefixes

I would highly suspect people who receive this message don't know what appstream prefixes are.  If there is a link you can send (or document on LVFS) what they are I think it would be useful.

> For all the vendors on the LVFS I will assign what I think are the correct vendor prefixes, based on the existing values uploaded, and for vendors with no uploaded firmware a prefixed based on the reverse DNS of the OEM homepage

No need to mention what you "think are the correct vendor prefixes".  Maybe instead: "I will assign vendor prefixes based upon existing values uploaded"

> Existing firmware in testing and stable isn’t affected by the new rules, and this will only affect new firmware uploaded. If anyone has any concerns please say. I’m confident the new functionality will allow us to grow the LVFS and also be suitable for the changing landscape of ODM and OEM relationsips.

I think there will be questions about stuff _currently_ in $oem-embargo.  Can ODM promote those all the way to stable now given the right permissions?

hughsie:
> No need to mention what you "think are the correct vendor prefixes". Maybe instead: "I will assign vendor prefixes based upon existing values uploaded"

Right, thanks.
 
> I think there will be questions about stuff _currently_ in $oem-embargo. Can ODM promote those all the way to stable now given the right permissions?

Good question. Either I (probably it'll be me...) or the vendor manager will have to reassign these before pushing. I'll obviously not deploy this until I've heard back from the other two bit OEMs using the affiliates feature (so the ODMs get the right permissions set from day 0) and also when I can view all the firmware problems and fix them manually. I also need to rerun several months of stats calculations with the new vendor's set correctly. (e.g., Wistron will disappear from the graph)

PR #394, opened 2019-09-09
Redesign the user list to use bootstrap cards

hughsie:
This allows the page to be usable on mobile and provides more content without
disclosing vendor secrets.

Fixes: https://github.com/fwupd/lvfs-website/issues/391

### DIFF
--- a/lvfs/templates/vendorlist.html
+++ b/lvfs/templates/vendorlist.html
@@ -3,26 +3,12 @@
 {% block title %}Vendor Status{% endblock %}
 
 {% block content %}
-<div class="card">
-  <div class="card-body">
-  <div class="card-title">Vendors contacted by the LVFS team</div>
-<table class="table">
-  <tr class="row table-borderless">
-    <th class="col col-sm-3">Vendor</th>
-    <th class="col col-sm-1">fwupd</th>
-    <th class="col col-sm-1">Account</th>
-    <th class="col col-sm-1">Uploads</th>
-{% if g.user is defined and g.user.check_acl('@admin') %}
-    <th class="col col-sm-5">Comments</th>
-    <th class="col col-sm-1">&nbsp;</th>
-{% else %}
-    <th class="col col-sm-6">Comments</th>
-{% endif %}
-  </tr>
+
 {% for v in vendors %}
 {% if v.visible or (g.user is defined and g.user.check_acl('@admin')) %}
-  <tr class="row" id="{{v.group_id}}">
-    <td class="col col-sm-3">
+<div class="card mb-3" id="{{v.group_id}}">
+  <div class="card-body">
+    <div class="card-title">
 {% if v.visible %}
 {% if v.url %}
       <a href="{{v.url}}">{{v.display_name_with_team}}</a>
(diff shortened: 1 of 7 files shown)


hughsie:
Screenshot_2019-09-09 LVFS Vendor Status(3)
Screenshot_2019-09-09 LVFS Vendor Status(2)
Screenshot_2019-09-09 LVFS Vendor Status(1)
Screenshot_2019-09-09 LVFS Vendor Status

lvfs/templates/vendorlist.html
```diff
@@ -37,44 +23,56 @@
...
-    <td class="col col-sm-1">
-      <span class="fas fa-check-circle fs-5 text-success"></span>
-    </td>
-{% else %}
-    <td class="col col-sm-1"></td>
+    <li class="list-group-item">
+      <span class="fas fa-check-circle fs-2 text-success"></span>
+      Hardware supported by fwupd
+{% if v.plugins %}
+      ({{v.plugins|join}})
```
superm1: I think it would be better to show protocols here, it would be less manual stuff to fill out that gets out of date wouldn't it?

hughsie: Hmm, good idea. Will try.

hughsie: So, that maps to the protocol, rather than the plugin -- which is maybe more what the user wants. In doing so we lose the `>= 9.9.1` version info in some plugins, and have quite verbose output:

Screenshot_2019-09-09 LVFS Vendor Status(4)

superm1: I would argue the individual versions `>= 9.9.1` isn't super useful anymore.  That could be placed on an individual device page if there are requirements worth showing.

Why is that so verbose?

Shouldn't it be something like this:

```
Hardware supported by fwupd:
* USB Device Firmware Update
* UEFI Update Capsule 
```

hughsie: Something more like this?

Screenshot_2019-09-09 LVFS Vendor Status(6)
Screenshot_2019-09-09 LVFS Vendor Status(5)

superm1: Yes. I would suspect that you might hide a few things though:

Don't show the unknown line if there are other things with cab set (or maybe we need to correct the database for all the stuff already promoted)

Only show the count of public stuff (does Ami have public stuff?)

hughsie: Right, I was forgetting to filter on stable. All fixed I think.

lvfs/templates/vendorlist.html
```diff
@@ -37,51 +23,58 @@
...
+      <span class="fas fa-check-circle fs-2 text-success"></span>
+{% if v.protocols|length == 1 %}
+      Firmware update protocol used: {{v.protocols[0].name}}
 {% else %}
-    <td class="col col-sm-1"></td>
+      Firmware update protocols used:
+      <ul>
+{% for protocol in v.protocols %}
+        <li>{{protocol.name}}</li>
+{% endfor %}
```
superm1: Are you going to fix all existing ones with a DB fix to avoid unknown showing up here?

hughsie: I filtered the list to stable firmware, and you can't have a firmware in stable without a protocol now.

superm1: Stuff that was promoted before protocols went in, those are fixed already then?

hughsie: Yes, I did lots of clicking a while ago. :/

superm1:
OK lgtm then

PR #402, opened 2019-09-13
Add the CVEs to the component in a structured way

hughsie:
100% untested, but I wanted some feedback on the idea.


superm1:
Broadly I like this, and especially that it can pull it out of existing text automagically or let it be filled out.  Now that being said, have you noticed that it's always `CVE` not `cve`?  There's a few instances I noticed that you are using them both, but don't use a `.upper` / `.lower` type thing to distinguish.

lvfs/models.py
```diff
@@ -1212,6 +1213,40 @@ def ensure_info(self, guid, name):
...
+    component_id = Column(Integer, ForeignKey('components.component_id'), nullable=False)
+    kind = Column(Text, nullable=False)
+    value = Column(Text, nullable=False)
+
+    # link back to parent
+    md = relationship("Component", back_populates="issues")
+
+    @property
+    def url(self):
+        return 'https://nvd.nist.gov/vuln/detail/{}'.format(self.value)
```
superm1: I wonder if it's worth having a concept of a prefix that maps to a database URL.  In your example later I noticed the `INTEL-SA` one which presumably Intel puts out a security advisory that talks more specifically about their problems.

hughsie:
> have you noticed that it's always CVE not cve

Right, I used the lower case form in the metadata as it's the preferred form for AppStream.

>  In your example later I noticed the INTEL-SA one

Phase 2 I think.

PR #405, opened 2019-09-17
Blacklist generic GUIDs like 'main-system-firmware'

hughsie:


hughsie:
I didn't do this as a "problem" as the LVFS can't add or remove GUIDs. It's just a hard upload failure now.

PR #407, opened 2019-09-25
Add support for <name_variant_suffix>

hughsie:
Fixes https://github.com/fwupd/lvfs-website/issues/373


lvfs/templates/docs-metainfo-style.html
```diff
@@ -46,6 +46,17 @@
   </div>
 </div>
 
+<div class="card mt-3">
+  <div class="card-body">
+    <div class="card-title"><code>&lt;name_variant_suffix&gt;</code></div>
+    <ul class="card-text">
+      <li>Only use this optional tag if the <code>&lt;name&gt;</code> would be duplicated, e.g. if there are two variants of the same hardware</li>
+      <li>Use a short string, as it will be appended to the visible name with brackets if required</li>
+      <li>Don't duplicate any part of the name</li>
```
superm1: Is it worth adding a check for this and making a problem if it happens?

hughsie: I guess that's a good idea. Break both up into words, look for anything common?

lvfs/metadata.py
```diff
@@ -40,8 +40,9 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
         component = ET.SubElement(root, 'component')
         component.set('type', 'firmware')
         ET.SubElement(component, 'id').text = md.appstream_id
-        # until all front ends support <category> append the suffix */
+        # until all front ends support <category> and <name_variant_suffix> append both */
         ET.SubElement(component, 'name').text = md.name_with_category
+        #ET.SubElement(component, 'name_variant_suffix').text = md.name_variant_suffix
```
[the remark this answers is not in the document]
hughsie: Right, same with the category.

lvfs/views_component.py
```diff
@@ -164,6 +164,8 @@ def component_modify(component_id):
         md.appstream_id = request.form['appstream_id']
     if 'name' in request.form:
         md.name = request.form['name']
+    if 'name_variant_suffix' in request.form:
+        md.name_variant_suffix = request.form['name_variant_suffix']
```
superm1: To encourage usage, I think it would be really useful to raise a problem forcing putting a suffix when two releases with different GUIDs use the same title.

hughsie: I guess that's true. At the moment we calculate problems by looking at the `Component` as a single thing, without all the other objects visible. Computationally it's not an easy thing to work out JIT like the other `problems` as we'd have to consider the version and remote too.

lvfs/templates/docs-metainfo-style.html
superm1: Yeah, except for some prepositions.

hughsie: What prepositions would you use in the tag legitimately? I assumed it would likely just be one word.

superm1: Hard to predict the future :)  I guess the cases right now you're right - it's probably 1-2 words and they're not prepositions.

hughsie: Done.

lvfs/views_component.py
hughsie: I think i'm going to punt this for now, and come back to it is we are ever able to add "transient" test failures (e.g. from a cron job) as a firmware problem.

PR #411, opened 2019-10-10
Use YARA in preference to our dumb blocklist plugin

hughsie:
This will allow us to write more complicated rules in the future that are not
just searching for simple strings.


hughsie:
@matrosov does this look good to you?

matrosov:
> 
> 
> @matrosov does this look good to you?

yes sir, that's exactly what we need :-)

hughsie:
@matrosov I wondered a few things:

 * Is compiling the rules going to be a significant slowdown? We could compile them in an hourly cron-job and then we just load the compiled blob in this plugin if you think that would be a good idea going forwards. I don't think we're going to have hundreds of files or rules, but I'm guessing it's not the thing we want to be doing 1000x if we're re-scanning the entire archive.

* I probably want to scan for *yar files recursively, so we can organize the rules into coherent places -- such as rules that make sense for AMI only. I'm not sure how to "tag" firmware as being suitable for specific tests yet.

* At the moment I'm scanning the .efi files *and* the entire (possibly compressed) image with all the rules. I'm guessing some rules only want to be run on the efi binaries and never on the blob -- as some of the blobs are 50Mb+ in size.

PR #413, opened 2019-10-15
Allow researchers to run YARA queries on the public firmware

hughsie:
Some security researchers want to run YARA queries on the firmware on the LVFS.
They can (and do) this already, but it means:

 * Downloading tens of gigabytes of firmware using PULP_MANIFEST
 * Exploding each cabinet to get the capsule
 * Exploding each capsule into EFI binaries with chipsec
 * Running yara on the command line

Given the LVFS already has the shards cached for each firmware, just allow
authenticated users to run YARA queries on the public firmware. These are done
in an existing cron job as they typically take a few minutes to compile and run.

It's important to note that this functionality isn't providing anything more
than people can do already with wget and a shell script and so I don't think
it's overstepping the mark allowing carefully chosen security researchers
access to this kind of thing. If anything, I can keep an eye on them this way :)

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -11,6 +11,7 @@ import os
 import sys
 import hashlib
 import datetime
+import yara
 
 from flask import render_template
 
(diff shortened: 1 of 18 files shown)


hughsie:
Screenshot from 2019-10-15 15-57-33

superm1:
> It's important to note that this functionality isn't providing anything more
than people can do already with wget and a shell script and so I don't think
it's overstepping the mark allowing carefully chosen security researchers
access to this kind of thing. If anything, I can keep an eye on them this way :)

I think the important comparison to draw between someone `wget`/`bash` scripting and using a logged in instance is that the security boundaries to OEM embargo channels become a lot more blurry.  A small bug that uncovers an OEM's firmware to another logged in OEM is certainly bad, but uncovering it to a security researcher goes from bad to a lot worse potentially.

hughsie:
> security boundaries to OEM embargo channels become a lot more blurry

Agreed -- this is explicitly only scanning stuff in `stable` -- even `testing` is filtered out.

superm1:
Considering this is for researchers does it even make sense to render a UX as the result?  I wonder if a CSV is better suited for their needs.

hughsie:
I think a CSV export is a good idea, but I don't think it's the only thing they need.

PR #417, opened 2019-10-24
Allow the blocklist plugin to add persistent claims

hughsie:
This could be used for instance to detect Computrace, BootGuard or BIOSguard,
and used to provide helpful user information to the device model.

### DIFF
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -51,9 +51,9 @@ class SecurityClaim:
 
     @property
     def rating(self):
-        if 'signed-firmware' in self.attrs and 'device-checksum' in self.attrs:
+        if 'success-signed-firmware' in self.attrs and 'success-device-checksum' in self.attrs:
             return 2
-        if 'signed-firmware' in self.attrs:
+        if 'success-signed-firmware' in self.attrs:
             return 1
         return 0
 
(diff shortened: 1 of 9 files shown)


hughsie:
Screenshot_2019-10-24 LVFS OptiPlex 3050 AIO

superm1:
Lgtm

hughsie:
@superm1 anything else you can think it makes sense to detect and display? I did wonder about BIOSGuard, but @out0xb2 pointed out you can have the DXEs but are not active on a particular CPU generation... Perhaps it makes sense to allow the vendors to "opt in" in the metainfo file for a "manual" claim?

superm1:
I don't think making manual claims in metadata is valuable. There is nothing to stop lying (intentionally or unintentionally).

I do think that things like this can be informative claims but shouldn't be a positive or negative badge.

BootGuard makes great sense to me.

It's do agree that FV don't always get activated so this claim can't be part of a security story, just informative that the firmware contains a security technology.

hughsie:
Do you mean that manually added attributes like BootGuard are okay, OR that we should just show a BootGuard info claim if the firmware has the BG DXE?

superm1:
Sorry if I was unclear.  I don't think manually added attributes are good for this.  These types of claims should only be made if there is some sort of evidence to support the claim (IE tearing apart a FV or ME image).

PR #418, opened 2019-10-28
Wip/hughsie/yara

hughsie:


plugins/chipsec/__init__.py
```diff
@@ -147,13 +147,20 @@ def _run_chipsec_on_md(self, test, md):
         if not shards:
             blob = self._find_dell_pfs(md.blob)
             if blob:
+                shard_zlib = ComponentShard(plugin_id=self.id)
+                shard_zlib.set_blob(blob)
                 if blob.startswith(b'PFS.HDR.'):
+                    shard_zlib.ensure_info('b26f8b5c-2209-5a60-b301-39f837883a14',
```
superm1: I guess this GUID is just something you came up with to assign?

lvfs/views_component.py
```diff
@@ -694,3 +694,25 @@ def component_checksum_add(component_id):
...
+def component_download(component_id):
+
+    # find shard
+    # get firmware component
+    md = db.session.query(Component).\
+            filter(Component.component_id == component_id).first()
+    if not md:
+        flash('No component matched!', 'danger')
+        return redirect(url_for('.firmware'))
+    if not md.fw.check_acl('@view'):
```
superm1: I would think you just want to leave this to folks matching researcher right?

plugins/blocklist/rules/bios-guard.yar
```diff
@@ -0,0 +1,17 @@
+rule IntelBiosGuard
+{
+    meta:
+        license = "GPL-2.0+"
+        description = "Uses Intel BIOS Guard"
```
superm1: I don't think you can make a claim for uses as much as a claim for "Contains"

plugins/blocklist/rules/bios-guard.yar
```diff
@@ -0,0 +1,17 @@
...
+        long_description = "Hardware-assisted authentication and protection against BIOS recovery attacks"
+        fail = false
+        claim = "success-bios-guard"
+
+    strings:
+        $signature = "_AMIPFAT"
+        $updater = "AmiFlashUpd" wide
+        $header = "__BGKH__"
```
superm1: Is this really enough to indicate that it's using Intel BIOS guard?  I would think it's just indicating that it has the Ami Flash updater executable for it's flasher.

plugins/blocklist/rules/boot-guard.yar
```diff
@@ -0,0 +1,16 @@
+rule IntelBootGuard
+{
+    meta:
+        license = "GPL-2.0+"
+        description = "Uses Intel Boot Guard"
```
superm1: Again, I think "Contains" is more accurate in this instance.

plugins/blocklist/rules/bios-guard.yar
hughsie: It's `$header or ($signature and not $updater)` -- i.e. we exclude the updater.

plugins/blocklist/rules/bios-guard.yar
hughsie: True.

lvfs/views_component.py
hughsie: I wondered if it was useful for anything else -- other @view users can just download the .cab file and extract it already.

plugins/chipsec/__init__.py
hughsie: It's `generate-guid PFS` -- if you have something more official please let me know :)

plugins/blocklist/rules/bios-guard.yar
superm1: Oh I see.  OK.

lvfs/views_component.py
superm1: Well can always expand it to more people later if there is a need though too.  I think by keeping it researcher only you'll be more likely to looped into use cases that people want it.

plugins/chipsec/__init__.py
superm1: Ah OK.

I have no idea where the name `PFS` even came from.  Whoever first reverse engineered one of the implementations maybe?

hughsie: It's the first three bytes of the decompressed Zlib section.

lvfs/views_component.py
hughsie: Fair point, will fix.

Issue #420, opened 2019-11-04
demote should go to testing rather than embargo


superm1:
I've been messaged a few times from people who aren't sure what happened to firmware because it completely disappears from LVFS when there are enough reported failures.

Principally I have no problem with the downgrades, I think they make sense to prevent more failures.  However I think they should go to `testing` instead.  Then at least people who know there are a bunch of reports coming in with failures but still want to get the firmware can get it.

hughsie:
Makes sense to me. Can you do a PR? I'm afk most of this week.

PR #421, opened 2019-11-04
cron: only demote stable->testing, not testing->embargo or stable->em…

superm1:
…bargo (Fixes: #420)


hughsie:
In principle yes, but make check needs to pass 🙂

superm1:
> In principle yes, but make check needs to pass 🙂

fixed, thanks.

PR #422, opened 2019-11-06
Allow changing the vendor group_id by the admin

hughsie:


hughsie:
> for movement between companies

Companies renaming actually, e.g. Toshiba Memory -> Kioxia

PR #424, opened 2019-11-07
Wip/hughsie/verfmt

hughsie:


migrations/versions/05217bb7d0c0_verfmt.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # create the ones we support already
+        verfmts = [
+            Verfmt(value='plain', name='Plain'),
+            Verfmt(value='pair', name='Pair of integers', specification='AA.BB', sections=2),
+            Verfmt(value='triplet', name='Microsoft', specification='AA.BB.CCDD', sections=3),
+            Verfmt(value='quad', name='Dell', specification='AA.BB.CC.DD', sections=4),
+            Verfmt(value='intel-me', name='Intel ME bitshift', specification='a+11.a.BB.CCDD', sections=4),
+            Verfmt(value='intel-me2', name='Intel ME', specification='A.A.BB.CCDD', sections=4),
+            Verfmt(value='bcd', name='Dell Style', specification='AA.BB', sections=2),
```
superm1: I think this name is wrong

lvfs/views_upload.py
```diff
@@ -231,7 +233,12 @@ def _upload_firmware():
     # fall back to a version format when unspecified and not semver
     for md in fw.mds:
         if not md.version_format and vendor.version_format and md.version.find('.') == -1:
-            md.version_format = vendor.version_format
+            md.verfmt = vendor.verfmt
+
+    # verify each component has a version format
+    for md in fw.mds:
+        if not md.version_format:
+            flash('Component {} does not have LVFS::VersionFormat set'.format(md.appstream_id), 'warning')
```
superm1: This will probably affect most firmware uploads, no?

Another way to approach this is to message that you're treating the format a certain way after uploading.

superm1:
I think it's a little confusing to have `md.verfmt` and also `md.version_format`.  future selves might scratch heads on that.

lvfs/views_upload.py
hughsie: > This will probably affect most firmware uploads, no?

Well, we fall back to the vendor version format, so in the Dell case there's no warning.

hughsie:
> I think it's a little confusing to have md.verfmt and also md.version_format

Yup, you're completely right. I'll do the extra work to convert everything over.

lvfs/views_upload.py
superm1: Oh right, I forgot about the vendor version format thing.

migrations/versions/05217bb7d0c0_verfmt.py
hughsie: Fixed, thanks. I've also simplified a few things.

lvfs/models.py
```diff
@@ -1085,6 +1089,53 @@ def success(self):
...
+                                    (v & 0x00ff0000) >> 16,
+                                    v &  0x0000ffff)
+        if self.value == 'intel-me2':
+            return '%i.%i.%i.%i' % ((v & 0xf0000000) >> 28,
+                                    (v & 0x0f000000) >> 24,
+                                    (v & 0x00ff0000) >> 16,
+                                    v &  0x0000ffff)
+        if self.value == 'bcd':
+            return '%i.%i' % ((v & 0xf0) >> 4, v & 0x0f)
+        return None
```
superm1: I guess you'll want ta second follow up commit to add those two new surface ones in the same PR while you're touching all this stuff?

lvfs/templates/verfmt-details.html
```diff
@@ -0,0 +1,53 @@
...
+    <input type="text" class="form-control" name="fwupd_version" value="{{verfmt.fwupd_version}}" />
+  </div>
+  <input type="submit" class="btn btn-primary btn-large" value="Modify">
+  <button type="button" class="btn btn-danger" data-toggle="modal" data-target="#deleteModal">Delete</button>
+</form>
+  </div>
+</div>
+
+<!-- modal dialog -->
+<div class="modal" id="deleteModal" tabindex="-1" role="dialog" aria-labelledby="deleteModalLabel" aria-hidden="true">
```
superm1: is there really actually a case for removing them even?

lvfs/templates/verfmt-list.html
```diff
@@ -0,0 +1,48 @@
...
+    <form method="post" action="{{url_for('.verfmt_add')}}" class="form">
+      <input type="hidden" name="csrf_token" value="{{csrf_token()}}"/>
+      <div class="form-group card-text">
+        <label>Value name, e.g. <code>intel-me2</code></label>
+        <input class="form-control" type="text" name="value" placeholder="Value" required />
+      </div>
+      <input class="card-link btn btn-primary" type="submit" value="Add"/>
+    </form>
+  </div>
+</div>
```
superm1: actually I would question the value of having a form to add new formats given the way to slice and transform the data actually requires a code change.

superm1: as in modifying `uint32_to_str`

lvfs/templates/verfmt-details.html
hughsie: That's a good question. From my testing point of view, I guess. If you do try and remove a version format that a firmware, vendor or protocol is using then it returns with an error.

lvfs/models.py
hughsie: Right, will do.

lvfs/templates/verfmt-list.html
hughsie: The uint32_to_str is only used in the case where the metainfo.xml is an integer, which is only typical in UEFI firmware. The idea of putting most of the data in the database was so we can add alternative formats for other devices easily (and to avoid hardcoding it differently in various places) although I concede manually having to fix the code to convert the integer isn't awesome. With the surface enums I can't see any other way to possibly convert a 32 bit number, so maybe this isn't a huge concern.

hughsie:
This actually fails in practice as some components with the dell-dock protocol use `bcd` versions and some are using `quad`. E.g. /lvfs/component/2512 is `01.21` and /lvfs/component/2510 is `01.00.00.00`...

superm1:
> This actually fails in practice as some components with the dell-dock protocol use bcd versions and some are using quad. E.g. /lvfs/component/2512 is 01.21 and /lvfs/component/2510 is 01.00.00.00...

But don't all of the uploads for `dell-dock` have version format set properly already?  Maybe for that particular one there shouldn't be an automatic mapping and instead LVFS require that the key be set per component.

hughsie:
> don't all of the uploads for dell-dock have version format set properly already

No, those two don't, the other newer ones seem to. I guess it's actually working as designed catching cases where the version format is ambiguous.

superm1:
> > don't all of the uploads for dell-dock have version format set properly already
> 
> No, those two don't, the other newer ones seem to. I guess it's actually working as designed catching cases where the version format is ambiguous.

Oh they're probably from before we started adding version format support and before the dock released (IE really old and will never be promoted).  I'm fine with those being deleted if it helps.  Nothing earlier than the first stable release (that is in stable) is needed now.

hughsie:
I've been testing this quite a bit now, and seem to have ironed out all of the regressions. Any concerns, or good to deploy on Monday? Also, does this look sane to you?

    com.hughski.colorhug = bcd
    org.altusmetrum.altos = bcd
    com.qualcomm.dfu = bcd
    com.dell.dock
    com.synaptics.mst = triplet
    org.usb.dfu = bcd
    com.st.dfuse = bcd
    com.8bitdo = bcd
    com.google.fastboot
    org.flashrom
    org.nvmexpress
    org.dmtf.redfish
    com.realtek.rts54 = bcd
    com.acme.test
    com.intel.thunderbolt = triplet
    org.uefi.capsule
    com.logitech.unifying
    com.logitech.unifyingsigned
    com.wacom.usb = pair
    com.wacom.raw = pair
    org.t13.ata
    com.qualcomm.qmi-pdc
    tw.com.ite.superio = pair
    com.synaptics.prometheus = triplet
    com.solokeys = triplet
    com.synaptics.rmi = triplet
    com.synaptics.cxaudio = quad
    org.jedec.mmc

superm1:
Thunderbolt is not triplet, it's pair.

superm1:
Otherwise looks good to me

superm1:
Looks like 8bitdo is wrong, should be pair.
https://github.com/fwupd/fwupd/blob/master/plugins/ebitdo/fu-ebitdo-device.c#L230
Also same thing for Thunderbolt: 
https://github.com/fwupd/fwupd/issues/1551

superm1:
Looks like realtek too: https://github.com/fwupd/fwupd/blob/master/plugins/rts54hid/fu-rts54hid-device.c#L249

hughsie:
> Looks like 8bitdo is wrong, should be pair.

Fixed on the LVFS

>  Also same thing for Thunderbolt:

Already set as pair...

> Looks like realtek too

Hmm. It looks like HUB is `bcd` and HID is `pair`, right? I suspect we want two protocols, not one.

superm1:
> Hmm. It looks like HUB is bcd and HID is pair, right? I suspect we want two protocols, not one.

Aren't they just two ways to talk to the same device? They should both be the same format I would think.

superm1:
> Fixed on the LVFS

I guess this will be effective after a stable metadata refresh cycle?  I just checked.

```
$ fwupdmgr refresh --force
Fetching metadata https://cdn.fwupd.org/downloads/firmware-testing.xml.gz
Downloading…             [***************************************]
Fetching signature https://cdn.fwupd.org/downloads/firmware-testing.xml.gz.asc

Fetching metadata https://cdn.fwupd.org/downloads/firmware.xml.gz
Downloading…             [***************************************]
Fetching signature https://cdn.fwupd.org/downloads/firmware.xml.gz.asc

Successfully downloaded new metadata: 2 local devices supported
$ fwupdmgr get-releases --verbose --force 
Choose a device:
0.	Cancel
1.	672c087de09848d9e7ee32aa1dea2fbeb8b81e6b (8Bitdo SF30 Pro)
2.	b6c08fb9e5384d9d101853cc1ca20cf0ce2df2e2 (System Firmware)
1
No releases found for device: Firmware version formats were different, device was 'pair' and release is 'bcd', Firmware version formats were different, device was 'pair' and release is 'bcd', Firmware version formats were different, device was 'pair' and release is 'bcd'
```

superm1:
Looks like dell-dock broke too for two of the controllers.

```
$ fwupdmgr get-releases 
Choose a device:
0.	Cancel
1.	d48d4e219eaa1c98a7d7081b3ba101676351a581 (Thunderbolt Controller)
2.	8357922f7f208e1f3b7537141ce17db04bc219d9 (Thunderbolt controller in Dell dock)
3.	1eca9eabb0c992c136e1deb1f89e3f70c465aa1c (Package level of Dell dock)
4.	ca36279eeac13463fd5d974c88794859294fbfb1 (RTS5413 in Dell dock)
5.	acb404019656654d44f80922d94735e831d9bb40 (RTS5487 in Dell dock)
6.	58931e8c9ef5eead9c007563814f96fcadb1b993 (VMM5331 in Dell dock)
7.	a455a3a21752e5d31c228f1f5b51761dad88a432 (WD19TB)
8.	6c24a747f97668873b761558e322398a91dbf394 (System Firmware)
4
No releases found for device: Firmware version formats were different, device was 'pair' and release is 'quad'
$ fwupdmgr get-releases 
Choose a device:
0.	Cancel
1.	d48d4e219eaa1c98a7d7081b3ba101676351a581 (Thunderbolt Controller)
2.	8357922f7f208e1f3b7537141ce17db04bc219d9 (Thunderbolt controller in Dell dock)
3.	1eca9eabb0c992c136e1deb1f89e3f70c465aa1c (Package level of Dell dock)
4.	ca36279eeac13463fd5d974c88794859294fbfb1 (RTS5413 in Dell dock)
5.	acb404019656654d44f80922d94735e831d9bb40 (RTS5487 in Dell dock)
6.	58931e8c9ef5eead9c007563814f96fcadb1b993 (VMM5331 in Dell dock)
7.	a455a3a21752e5d31c228f1f5b51761dad88a432 (WD19TB)
8.	6c24a747f97668873b761558e322398a91dbf394 (System Firmware)
5
No releases found for device: Firmware version formats were different, device was 'pair' and release is 'quad'
```

PR #425, opened 2019-11-08
Use PSPTool to parse the AMD PSP section

hughsie:
This is very similar to how the IntelME parser works.

### DIFF
--- a/plugins/amdpsp/__init__.py
+++ b/plugins/amdpsp/__init__.py
@@ -1,0 +1,122 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 Richard Hughes <richard@hughsie.com>
+#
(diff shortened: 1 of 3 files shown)


plugins/amdpsp/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+            # only until required
+            return True
+        if md.protocol.value != 'org.uefi.capsule':
+            return False
+
+        # match on category
+        if not md.category:
+            # only until required
+            return True
+        return md.category.matches(['X-System', 'X-ManagementEngine'])
```
superm1: Is that right?  There shouldn't be ME binaries for AMD.

superm1: Specifically I would think that it should be a new `X-PlatformSecurityProcessor`

hughsie: Fixed, thanks.

PR #428, opened 2019-11-11
Wip/hughsie/blueprint prep

hughsie:
We should have done this a long time ago...


hughsie:
> I don't think i'll be able to provide a very useful review for this one, sorry

No worries, I suspected that. I was wondering if you had some pearls of wisdom with regards to naming of the new files. I'll merge and deploy tomorrow and we can always revert if there is some horrible regression.

superm1:
> > I don't think i'll be able to provide a very useful review for this one, sorry
> 
> No worries, I suspected that. I was wondering if you had some pearls of wisdom with regards to naming of the new files. I'll merge and deploy tomorrow and we can always revert if there is some horrible regression.

Sounds good.

PR #429, opened 2019-11-11
Wip/hughsie/vendor verfmt fallback

hughsie:


hughsie:
Deployed, thanks for the super-speedy review.

PR #431, opened 2019-11-13
trivial: Don't require CSRF for login

hughsie:
Else uploading from a cron job can fail.


superm1:
Huh?  Why does uploading from a cron job cause CSRF failures?

hughsie:
This code in example.py:

    r = s.post('%s/lvfs/login' % os.environ['LVFS_SERVER'], data=data)

i.e. we don't actually load /lvfs/login before we POST to it, causing the CSRF failure.

PR #435, opened 2019-11-14
Allow syncing with other firmware databases

hughsie:
External teams can contribute metadata about firmware releases from public
sources such as a vendor homepage or the general list downloaded from
Windows Update. This functionality allows us to check both that models have
been included on the LVFS and that matching versions are present.

Vendor managers are allowed to view this data, but only if the appstream ID
matches one of their assigned namespace prefixes.

### DIFF
--- a/contrib/example.py
+++ b/contrib/example.py
@@ -35,6 +35,36 @@ def _vendor_user_add(session, vendor_id, username, display_name):
         print('failed to create user using %s: %s' % (rv.url, rv.text))
         sys.exit(1)
 
+def _mdsync_import(session, filename):
+
+    # import
+    try:
+        with open(filename, 'r') as f:
+            payload = f.read()
+    except IOError as e:
+        print('Failed to load file', str(e))
+        sys.exit(1)
+    rv = session.post('/'.join([os.environ['LVFS_SERVER'], 'lvfs', 'mdsync', 'import']), data=payload)
+    if rv.status_code != 200:
+        print('failed to import mdsync using %s: %s' % (rv.url, rv.text))
+        sys.exit(1)
+    print('imported {} successfully: {}'.format(filename, rv.text))
+
+def _mdsync_export(session, filename):
+
+    # export
+    rv = session.get('/'.join([os.environ['LVFS_SERVER'], 'lvfs', 'mdsync', 'export']))
+    if rv.status_code != 400:
+        print('failed to export mdsync: {}'.format(rv.status_code))
+        sys.exit(1)
+    try:
+        with open(filename, 'w') as f:
+            f.write(rv.text)
+    except IOError as e:
+        print('Failed to save file', str(e))
+        sys.exit(1)
+    print('exported {} successfully'.format(filename))
+
 if __name__ == '__main__':
 
     # check required env variables are present
@@ -44,7 +74,7 @@ if __name__ == '__main__':
             sys.exit(1)
 
     if len(sys.argv) < 2:
-        print('Usage: %s [upload|create-user]' % sys.argv[0])
+        print('Usage: %s [upload|create-user|mdsync-import|mdsync-export]' % sys.argv[0])
         sys.exit(1)
 
     # log in
@@ -67,6 +97,16 @@ if __name__ == '__main__':
             print('Usage: %s upload vendor_id username display_name' % sys.argv[0])
             sys.exit(1)
         _vendor_user_add(s, sys.argv[2], sys.argv[3], sys.argv[4])
+    elif sys.argv[1] == 'mdsync-import':
+        if len(sys.argv) != 3:
+            print('Usage: %s mdsync-import filename' % sys.argv[0])
+            sys.exit(1)
+        _mdsync_import(s, sys.argv[2])
+    elif sys.argv[1] == 'mdsync-export':
+        if len(sys.argv) != 3:
+            print('Usage: %s mdsync-export filename' % sys.argv[0])
+            sys.exit(1)
+        _mdsync_export(s, sys.argv[2])
     else:
         print('command not found!')
         sys.exit(1)
--- a/lvfs/__init__.py
+++ b/lvfs/__init__.py
@@ -55,6 +55,7 @@ from lvfs.analytics.routes import bp_analytics
 from lvfs.categories.routes import bp_categories
 from lvfs.components.routes import bp_components
 from lvfs.devices.routes import bp_devices
+from lvfs.mdsync.routes import bp_mdsync
 from lvfs.docs.routes import bp_docs
 from lvfs.firmware.routes import bp_firmware
 from lvfs.issues.routes import bp_issues
@@ -78,6 +79,7 @@ app.register_blueprint(bp_analytics, url_prefix='/lvfs/analytics')
 app.register_blueprint(bp_categories, url_prefix='/lvfs/categories')
 app.register_blueprint(bp_components, url_prefix='/lvfs/components')
 app.register_blueprint(bp_devices, url_prefix='/lvfs/devices')
+app.register_blueprint(bp_mdsync, url_prefix='/lvfs/mdsync')
 app.register_blueprint(bp_docs, url_prefix='/lvfs/docs')
 app.register_blueprint(bp_firmware, url_prefix='/lvfs/firmware')
 app.register_blueprint(bp_issues, url_prefix='/lvfs/issues')
--- a/lvfs/mdsync/routes.py
+++ b/lvfs/mdsync/routes.py
@@ -1,0 +1,282 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: GPL-2.0+
+
+import json
+
+from collections import defaultdict
+import dateutil.parser
+
+from flask import Blueprint, request, url_for, redirect, flash, Response, render_template, g
+from flask_login import login_required
+
+from lvfs import db, csrf
+
+from lvfs.models import Firmware, Component, Remote, ComponentRef, Protocol, Vendor
+from lvfs.util import _json_success, _json_error
+
+bp_mdsync = Blueprint('mdsync', __name__, template_folder='templates')
+
+def _md_to_mdsync_dict(md):
+    obj = {}
+    obj['component_id'] = md.component_id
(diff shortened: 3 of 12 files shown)


lvfs/devsync/routes.py
```diff
@@ -0,0 +1,227 @@
...
+    if not g.user.check_acl('@yara-query'):
+        return _json_error('Permission denied: Unable to export data')
+
+    # get a dict of all stable components, with the appstream-id as dedupe-key
+    mds_by_ids = defaultdict(list)
+    for md in db.session.query(Component).\
+                    join(Firmware).\
+                    join(Remote).\
+                    filter(Remote.name != 'private',
+                           Remote.name != 'deleted').\
```
superm1: no need to filter embargo here?

superm1: I guess you have _any_public check later

lvfs/devsync/routes.py
```diff
@@ -0,0 +1,227 @@
...
+                    mimetype="application/json")
+
+@bp_devsync.route('/import', methods=['POST'])
+@csrf.exempt
+@login_required
+def route_import():
+
+    # security check
+    if not g.user.check_acl('@yara-query'):
+        return _json_error('Permission denied: Unable to export data')
```
superm1: import

lvfs/devsync/routes.py
```diff
@@ -0,0 +1,227 @@
...
+    # parse blob
+    mdrefs = []
+    try:
+        metadata = obj['metadata']
+        if metadata['version'] != 0:
+            return _json_error('metadata schema version unsupported: ' + str(e))
+        for obj_dev in obj['devices']:
+
+            # required, but may be junk for devices not in LVFS
+            appstream_id = obj_dev['appstream_id']
```
superm1: so that means the onus to match it to the correct appstream ID is on the external syncing service before it send the data to LVFS right?

superm1: Since appstream id is actually a freeform field, I have to wonder if GUID might be a better selection for matching?

lvfs/devsync/routes.py
hughsie: An excellent question. I did originally filter this to `stable` or `testing` but the number of firmware versions that were "missing" from the LVFS was huge. Including the firmware in embargo made the delta much smaller. I'm not completely happy exposing the fact that the version *exists* in embargo, but I thought seeing as we're:

 * limiting this to trusted partners that we're trusting not to redistribute the data
 * only showing versions in embargo that have one public release (so we don't leak product names before release)
 * Only showing the version number, and **not** the update details or list of CVEs

...I thought it was pragmatic to include them. Long term I hope the firmware could go stable on the same day as published on Windows Update, but perhaps that's just too optimistic.

lvfs/devsync/routes.py
hughsie: Fixed, thanks.

lvfs/devsync/routes.py
hughsie: > the onus to match it to the correct appstream ID is on the external syncing service

The other company doesn't really have a good identifier for models -- they were using the model _name_ which is a pretty terrible way to identify a stream of firmware. We agreed they'd also use the same component IDs to disambiguate the models.

> I have to wonder if GUID might be a better selection for matching

Although you can guess the GUID from WU, you can't typically get the GUID when scraping the vendor website -- which is where most of this data is probably going to come from.

lvfs/devsync/routes.py
superm1: >  "missing" from the LVFS was huge.

As in a ton on embargo but not promoted to testing/stable?  I guess those are valuable points to call out.

> Long term I hope the firmware could go stable on the same day as published on Windows Update, but perhaps that's just too optimistic.

Way too optimistic.  There is a long lead time in promotion steps with WU.  Timing it together is going to be really hard.

lvfs/devsync/routes.py
superm1: You can do it the other way around though, if you take the GUID from LVFS you can punch it into WU catalog (https://www.catalog.update.microsoft.com/Home.aspx) to search for FW.

superm1: And yeah as for scraping a a vendor website - you're going to run into problems with those "variants" we've talked about.  It's not always a 1:1 mapping.

When you have a "-H" and a "-U" CPU variant the binary on support.dell.com will typically have both variants in the EXE.  When running it on Windows or in BIOS F12 menu it pulls out the appropriate payloads.

But the capsule on LVFS will only have the single CPU variant "-H" or "-U", meaning you have two ESRT GUIDs.

lvfs/devsync/routes.py
hughsie: > As in a ton on embargo but not promoted to testing/stable

Yes, from all vendors, including Dell :) I guess we don't do very much to make it easy to tell the vendor "hey, you're pushing 0.1.2 to stable, and 0.1.1 is still in embargo!" -- I'm not sure how we could help the vendor in that respect.

>  Timing it together is going to be really hard.

Would it help if you could set some kind of date to move the firmware from testing to stable automatically?

superm1: > Yes, from all vendors, including Dell :)

I can't represent all instances - but I know there are some where it's "We know this causes a regression on Linux systems, lets skip pushing this one and push the next one with the fix".

> Would it help if you could set some kind of date to move the firmware from testing to stable automatically?

At least for Dell I don't think there is any reason that it can't go out to LVFS before WU from these timing limitations.

hughsie: > We know this causes a regression on Linux systems

This is a really useful thing to record on the LVFS, i.e. the reason a firmware was never promoted or was promoted and then demoted manually. Do you think vendors would be able to share this kind of data with the LVFS or is too commercially sensitive?

Also, I don't think leaving the firmware in embargo is the right thing to do here. Embargo was supposed to be a place where vendors could do end-to-end tests, rather than a place where firmware just lives on half-dead forever. I'm not sure I have a very good idea on where it *should* go however.

lvfs/devsync/routes.py
hughsie: >  It's not always a 1:1 mapping

Ya, I shared this education with them last week. 

> will typically have both variants in the EXE

Are the -H and -U versions always updated in lockstep, i.e. do they always have the same firmware versions released at the same time?

lvfs/devsync/routes.py
superm1: > This is a really useful thing to record on the LVFS, i.e. the reason a firmware was never promoted or was promoted and then demoted manually. Do you think vendors would be able to share this kind of data with the LVFS or is too commercially sensitive

I don't know about non-Dell, but I don't believe that Dell would generally share this kind of information with the public.

> Also, I don't think leaving the firmware in embargo is the right thing to do here. Embargo was supposed to be a place where vendors could do end-to-end tests, rather than a place where firmware just lives on half-dead forever. I'm not sure I have a very good idea on where it should go however.

Maybe a longish policy (>60 days) that demotes from embargo to private and then another one that goes from private to deleted makes sense?

lvfs/devsync/routes.py
superm1: > Are the -H and -U versions always updated in lockstep, i.e. do they always have the same firmware versions released at the same time?

We're talking people here.  I think it's entirely possible for a person to forget to promote one or the other on LVFS.  But I think that's the goal.

hughsie:
I'm wondering if devsync is the wrong "noun" entirely. What about `mdsync` or `vendormd`?

superm1:
Agree `metadatasync` or `mdsync` do sound more descriptive to what this does

superm1:
It's not really rocket science scraping websites, just busy work.

In general I think my preference would be that the code that is scraping and syncing databases be part of LVFS rather than this import/export process.  Then at least it's a lot easier to audit where the code is failing when a vendor website changes, and also easier to turn it off if bad metadata starts to import.

hughsie:
> It's not really rocket science scraping websites, just busy work.

The scraping is really only phase 1, what I really want is for Microsoft to share their WU catalogue, Dell to share their internal metadata, Lenovo to Sync directly etc.

superm1:
> The scraping is really only phase 1, what I really want is for Microsoft to share their WU catalogue, Dell to share their internal metadata, Lenovo to Sync directly etc.

What exactly do you have in mind on this?  Like some process at Dell that builds a database and uploads it?
At least for Dell it's a pretty monumental I/T task to export stuff that isn't already being exported today to the catalog for existing tooling.  If we put the request in today _maybe_ in 3 years we might see it ;)  

I jest, but seriously those folks usually need a strong business case to make changes that would export more.

hughsie:
> What exactly do you have in mind on this? Like some process at Dell that builds a database and uploads it?

Yes, but I know it's a tall order. Maybe Lenovo will offer something more official, but the results we get from scraping are actually pretty good:

Screenshot_2019-12-05 LVFS Device Sync Status for Hughski Limited

superm1:
> Yes, but I know it's a tall order. Maybe Lenovo will offer something more official, but the results we get from scraping are actually pretty good:

Hopefully that's from the XML not HTML scraping right?  I guess that's with trying to match by the platform's name and LVFS name for those that match up?

I do thing a few valuable things that can be achieved by doing this:
1. (At least for Dell) Since you have the EXE you can download it, sha256 sum it and then strip the capsule header from the one at LVFS and sha256 sum it and make sure they match up.
2. (At least for Dell) You can automatically add a link back to the page at support.dell.com.  The URL is built like this:  https://www.dell.com/support/home/us/en/19/drivers/driversdetails?driverid=txhtk
That driver ID is in the XML as well.

superm1:
> (At least for Dell) Since you have the EXE you can download it, sha256 sum it and then strip the capsule header from the one at LVFS and sha256 sum it and make sure they match up.

I don't recall for sure, but maybe the XML even has the sha256 sum and you can avoid having to download it.

hughsie:
> then strip the capsule header from the one at LVFS

No capsule header on the windows one?

> and make sure they match up

Phase 2 I think.

superm1:
> No capsule header on the windows one?

As you can tell from https://github.com/fwupd/fwupd/blob/master/contrib/firmware_packager/install_dell_bios_exe.py the capsules for Dell are just capsule header + exe file.  So I think it's entirely reasonable to strip the header, run `file` on it and make sure it looks like a PE file and then compare checksums.

PR #438, opened 2019-11-18
Wip/hughsie/become

hughsie:
This is a powerful admin-only action that allows me to verify bugs on the production instance. Or rather to verify that the user reporting the bug can actually see the buttons on the screen :)

### DIFF
--- a/lvfs/main/routes.py
+++ b/lvfs/main/routes.py
@@ -32,7 +32,7 @@ from lvfs.models import User, Client, Event, AnalyticVendor, Remote
 from lvfs.models import _get_datestr_from_datetime
 from lvfs.hash import _addr_hash
 from lvfs.util import _get_client_address, _get_settings, _xml_from_markdown, _get_chart_labels_days
-from lvfs.util import _event_log, _error_internal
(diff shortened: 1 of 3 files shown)


superm1:
As an alternative why not just sync to testing interface and reset the user password in the testing instance database?

hughsie:
> As an alternative why not just sync to testing interface and reset the user password in the testing instance database?

Yes, this is what I was doing before, but syncing the prod->testing databases was becoming more of a pain. If you think this is a Bad Idea (tm) I can drop this, or perhaps make it conditional on DEBUG=1 being set.

superm1:
> Yes, this is what I was doing before, but syncing the prod->testing databases was becoming more of a pain. If you think this is a Bad Idea (tm) I can drop this, or perhaps make it conditional on DEBUG=1 being set.

I do think it's a bad idea, it opens up the opportunity for falsified event logs in the event admin credentials were hacked.  I know there are more dangerous things that can be done as admin, but stealthy stuff is scary.

I would think you're better off making it easier to sync DB to testing/staging.

hughsie:
> I do think it's a bad idea

Fair enough, thanks for the sanity check.

PR #442, opened 2019-11-19
Add some information to tell the consulting company

hughsie:


hughsie:
I split this out with a header, as the page was getting a bit huge will a large wall of text. I hope this is okay.

superm1:
No concern with formatting change no

superm1:
One thing i'm wondering if it's needed is something about sustaining.  You and I both know plugins do break from time to time in non-obvious ways as daemon and library move up and grow new features (The two most recent that come to mind were version format and new vfuncs).  Unit tests are ideal, but often hard to emulate all those interactions.

I think we usually catch most of them by randomly running tests with real hardware.  So perhaps when reaching out to a consulting company that's something else to ask for, continuing to ensure it works with the latest upstream version.

hughsie:
>  So perhaps when reaching out to a consulting company that's something else to ask for, continuing to ensure it works with the latest upstream version.

Right. I don't know how the consulting companies are going to deal with this, as of course everything can regress. My longer term hope is the vendors get involved with upstream through hiring Linux engineers, much like Dell has. I'll open up talks with all the companies we list and discuss if hey want other things added to the "questions to ask" page.

PR #445, opened 2019-12-03
Wip/hughsie/dell bios

hughsie:


migrations/versions/13c307a09ff1_add_verfmts_fallback.py
```diff
@@ -0,0 +1,45 @@
...
+    for md in db.session.query(Component):
+        if md.fw.vendor.group_id != 'dell':
+            continue
+        if md.verfmt and md.verfmt.value != 'quad':
+            continue
+        if md.protocol and md.protocol.value != 'org.uefi.capsule':
+            continue
+        print('fixing', md.component_id, md.names)
+        md.verfmt= verfmt
+    db.session.commit()
```
superm1: don't the fallbacks column need to be set to `quad` too?

lvfs/models.py
```diff
@@ -1098,7 +1098,7 @@ class Verfmt(db.Model):
     __table_args__ = {'mysql_character_set': 'utf8mb4'}
 
     verfmt_id = Column(Integer, primary_key=True, nullable=False, unique=True)
-    value = Column(Text, nullable=False)        # 'quad'
+    value = Column(Text, nullable=False)        # 'dell-bios'
```
superm1: I guess need to also to remember to add this to https://fwupd.org/lvfs/docs/metainfo/version

migrations/versions/13c307a09ff1_add_verfmts_fallback.py
hughsie: I just did added the dell-bios verfmt using the admin console before migrating the database, and then added the `quad` fallback after migration. I have to rebuild the dell metadata anyway manually, so it's not a huge problem.

lvfs/models.py
hughsie: That's autogenerated now :)

hughsie:
Let me sleep on this and then if I don't wake up in a panic I'll deploy it tomorrow. I assume you're okay with the fwupd side of this too?

superm1:
> Let me sleep on this and then if I don't wake up in a panic I'll deploy it tomorrow. I assume you're okay with the fwupd side of this too?

I'd like to test the fwupd side of it before merging to make sure nothing went wrong, but it _seems_ OK.

hughsie:
> I'd like to test the fwupd side of it before merging to make sure nothing went wrong

Ya, I suppose now it's decoupled there's no rush to merge the fwupd side.

PR #446, opened 2019-12-03
Add a release tag to store the vendor-specific firmware identifier

hughsie:
This could be used, for instance, by Lenovo to specify a tag of `N1NET43W` for
a specific firmware release.

### DIFF
--- a/lvfs/upload/uploadedfile.py
+++ b/lvfs/upload/uploadedfile.py
@@ -226,6 +226,13 @@ class UploadedFile:
         else:
             raise MetadataInvalid('<release> had no date or timestamp attributes')
 
+        # optional release tag
+        if 'tag' in release.attrib:
+            md.release_tag = release.attrib['tag']
+            if len(md.release_tag) < 4:
+                raise MetadataInvalid('<release> tag was too short to identify the firmware')
+            md.add_keywords_from_string(md.release_tag, priority=5)
+
         # get list of CVEs
         for issue in release.xpath('issues/issue'):
             kind = issue.get('type')
--- a/lvfs/components/routes.py
+++ b/lvfs/components/routes.py
@@ -168,6 +168,8 @@ def route_modify(component_id):
         md.name = request.form['name']
     if 'name_variant_suffix' in request.form:
         md.name_variant_suffix = request.form['name_variant_suffix']
+    if 'release_tag' in request.form:
+        md.release_tag = request.form['release_tag']
 
     # the firmware changed protocol
     if retry_all_tests:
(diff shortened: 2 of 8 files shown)


lvfs/docs/templates/docs-metainfo-intro.html
```diff
@@ -37,7 +37,7 @@
   &lt;project_license&gt;proprietary&lt;/project_license&gt;
   &lt;developer_name&gt;Hughski Limited&lt;/developer_name&gt;
   &lt;releases&gt;
-    &lt;release urgency="high" version="3.0.2" date="2017-02-09" install_duration="120"&gt;
+    &lt;release urgency="high" version="3.0.2" date="2017-02-09" install_duration="120" tag="N1CET75W"&gt;
```
superm1: Since tag is very specific to one vendor (that I know), I think it's better for the example to show an empty tag, but in the style documentation show the details of what it should contain.

superm1: The implied reason being most people will copy and paste this template for their starting point.

lvfs/upload/uploadedfile.py
```diff
@@ -226,6 +226,12 @@ def _parse_release(md, release):
         else:
             raise MetadataInvalid('<release> had no date or timestamp attributes')
 
+        # optional release tag
+        if 'tag' in release.attrib:
```
superm1: i think check and make sure > 0 too so empty tags can be sent

hughsie:
How about:

    <release tag="foo">
    The release tag is optional, and used to show a vendor-specific text identifier that is different from the version number. The tag may be unique only to the model, or be unique for the entire vendor. This attribute should not be used if the tag is not used to identify the firmware on the vendor homepage.

?

lvfs/docs/templates/docs-metainfo-intro.html
hughsie: Fair point, agreed.

lvfs/upload/uploadedfile.py
hughsie: I think if it's empty there should be no `tag=` set at all.

superm1:
> How about:
> 
> ```
> <release tag="foo">
> The release tag is optional, and used to show a vendor-specific text identifier that is different from the version number. The tag may be unique only to the model, or be unique for the entire vendor. This attribute should not be used if the tag is not used to identify the firmware on the vendor homepage.
> ```
> 
> ?

👍

hughsie:
Screenshot_2019-12-03 LVFS MetaInfo Files Style Guide

Issue #447, opened 2019-12-05
Feature Request: RSS feed for device pages


bjoern-tantau:
It would be nice to have an RSS feed available on the device pages (like https://fwupd.org/lvfs/devices/com.lenovo.ThinkPadN27ET.firmware) to be notified of new releases in case you're waiting for something specific.

hughsie:
I guess this makes sense, but I have no experience with RSS at all. Is this something you can help with?

bjoern-tantau:
I don't have much experience with Python. ;-)

But I'll try to hack something together. I'd simply output to another template when called with an `rss` query parameter.

bjoern-tantau:
I added a pull request (atom instead of RSS, as it is newer and allows for full HTML content) in #448, but the timestamps aren't formatted correctly yet. Although I suspect that most feed readers will be able to handle the current format it would be better if they got the correct format.

hughsie:
Merged, and deployed. Thanks!

PR #448, opened 2019-12-05
Added atom-feed to devices.

bjoern-tantau:
See  #447.
This is untested but should work mostly.

Only the dates should be converted from `2019-05-23 01:08:28` to `2019-05-23T01:08:28Z`. I don't know how to properly do that with python.

### DIFF
--- a/lvfs/devices/routes.py
+++ b/lvfs/devices/routes.py
@@ -81,6 +81,26 @@ def route_show(appstream_id):
                            fws=fws,
                            fw_previous=fw_previous)
 
+@bp_devices.route('/<appstream_id>/atom')
+def route_show_atom(appstream_id):
+    """
+    Show information for one device, which can be seen without a valid login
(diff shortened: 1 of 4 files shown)


hughsie:
> Only the dates should be converted

I think you want to append `.isoformat()` to do that. It's looking great already!

bjoern-tantau:
I added the `.isoformat()` but I couldn't get the website up on my Ubuntu environment. So this is completely untested. But even if the timestamps are not completely correct I suspect that most clients will be able to handle several different timestamp formats.

hughsie:
> but I couldn't get the website up on my Ubuntu environment

Where did you get stuck? I'm happy to help. We probably ought to test this in the self tests too.

hughsie:
Replaced with https://github.com/fwupd/lvfs-website/pull/460 -- thanks.

PR #449, opened 2019-12-11
Include a vendor ID by default for testing accounts

hughsie:
At the moment I create vendor accounts for OEMs and ODMs for testing. I then
create them a first user account, and over time they add more accounts, upload
firmware, test with embargo on real hardware and then ask to be able to promote
to stable.

This default state means that the new user with the @trusted ACL can push
firmware for any vendors hardware as the vendor-id was never set. It's very easy
to forget to ask the vendor what ID to use as part of the onboarding process.

Move to a model where vendors are restricted by default, so that any missing
vendor-ids are caught during testing with embargo.
For OEMs like Dell and Lenovo make the 'unrestricted' vendor ID an explicit
step (with a vendor ID of '*') rather than just a side effect of never
assigning a vendor ID.

When deploying this change we need to remember to manually give the vendor-id
of `*` to all the vendors shipping UEFI capsule or coreboot firmware so that the
vendor ID is not set, as fwupd does not set a vendor-id in all versions.

### DIFF
--- a/lvfs/main/templates/dashboard.html
+++ b/lvfs/main/templates/dashboard.html
@@ -24,6 +24,14 @@
 </div>
 {% endif %}
 
+{% if not g.user.vendor.restrictions %}
+<div class="alert alert-warning mb-3" role="alert">
+  This account has no vendor restrictions defined and firmware will not be
+  deployed onto hardware.
+  Please contact the <a href="mailto:info@fwupd.org">LVFS administrator</a>.
+</div>
+{% endif %}
+
 {% if not g.user.vendor.do_not_track and download_cnt > 0 %}
 <div class="card">
   <div class="card-body rounded-soft bg-gradient">
--- a/lvfs/search/templates/search.html
+++ b/lvfs/search/templates/search.html
@@ -61,6 +61,15 @@
       Is not uploading firmware to the LVFS
     </li>
 {% endif %}
+{% if not vendor.is_unrestricted and vendor.restrictions %}
+    <li class="list-group-item">
+      <span class="fas fa-info-circle fs-2 text-info"></span>
+      Restricted to upload firmware for any hardware with vendor ID:
+{% for restriction in vendor.restrictions %}
+        <code>{{restriction.value}}</code>
+{% endfor %}
+    </li>
+{% endif %}
       </ul>
     </td>
   </tr>
--- a/lvfs/metadata/utils.py
+++ b/lvfs/metadata/utils.py
@@ -15,7 +15,7 @@ from lxml import etree as ET
 
 from lvfs import app, db
 
-from lvfs.models import Firmware, Vendor, Remote
+from lvfs.models import Firmware, Remote
 from lvfs.util import _get_settings, _xml_from_markdown
 
 def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
@@ -61,10 +61,19 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
         for md in mds:
             if local:
                 break
-            vendor = db.session.query(Vendor).filter(Vendor.vendor_id == md.fw.vendor_id).first()
-            if not vendor:
+
+            # the vendor can upload to any hardware
+            vendor = md.fw.vendor
+            if vendor.is_unrestricted:
                 continue
+
+            # no restrictions in place!
             if not vendor.restrictions:
+                child = ET.Element('firmware')
+                child.text = 'vendor-id'
+                child.set('compare', 'eq')
+                child.set('version', 'XXX:NEVER_GOING_TO_MATCH')
+                elements['vendor-id'] = child
                 continue
 
             # allow specifying more than one ID
--- a/lvfs/metadata/utils.py
+++ b/lvfs/metadata/utils.py
@@ -63,7 +63,7 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
                 break
 
             # the vendor can upload to any hardware
-            vendor = md.fw.vendor
+            vendor = md.fw.vendor_odm
             if vendor.is_unrestricted:
                 continue
 
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -744,6 +744,10 @@ class Vendor(db.Model):
             if user.vendor_id != self.vendor_id:
                 return False
             return user.check_acl('@vendor-manager')
+        if action == '@view-restrictions':
+            if user.vendor_id != self.vendor_id:
+                return False
+            return user.check_acl('@vendor-manager')
         if action == '@modify-affiliations':
             return False
         if action == '@modify-affiliation-actions':
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -655,6 +655,13 @@ class Vendor(db.Model):
         return self.users
 
     @property
+    def is_unrestricted(self):
+        for res in self.restrictions:
+            if res.value == '*':
+                return True
+        return False
+
+    @property
     def display_name_with_team(self):
         if self.internal_team:
             return '{} ({})'.format(self.display_name, self.internal_team)
--- a/lvfs/vendors/routes.py
+++ b/lvfs/vendors/routes.py
@@ -225,13 +225,17 @@ def route_show(vendor_id):
 
 @bp_vendors.route('/<int:vendor_id>/restrictions')
 @login_required
-@admin_login_required
 def route_restrictions(vendor_id):
     """ Allows changing a vendor [ADMIN ONLY] """
+
+    # security check
     vendor = db.session.query(Vendor).filter(Vendor.vendor_id == vendor_id).first()
     if not vendor:
-        flash('Failed to get vendor details: No a vendor with that group ID', 'warning')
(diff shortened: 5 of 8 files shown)


superm1:
This itself looks fine, but I think with the increased focus on vendor ID's there's a few thoughts I have:
* It's not currently possible for a vendor manager to see what vendor ID's a vendor's LVFS account can use.
* Even if a vendor's LVFS account supports several vendor ID's, how do you assign the correct vendor ID for a firmware?  It seems like there needs to be a mapping relation between applicable vendor ID's for protocols or something.

IE Dell should probably use `DMI:Dell Inc.` for UEFI FW, but if you look at the vendor specified by Thunderbolt host controller it's actually `Dell`.  If you look at the vendor string specified by Dell's WD19 dock all the components are `Dell Inc` (no period).  The WD19 also doesn't specify vendor ID for some devices currently.

* I think the logic that is used for Synaptics touchpads reading Vendor is not good right now.  I see that on an XPS 7390 the vendor reads as `Intel Corporatation` (but `HIDRAW:0x06CB`)

hughsie:
> It's not currently possible for a vendor manager to see what vendor ID's a vendor's LVFS account can use.

It's actually on the vendor page now:

Screenshot_2019-12-12 LVFS Vendor Status

> Even if a vendor's LVFS account supports several vendor ID's, how do you assign the correct vendor ID for a firmware? It seems like there needs to be a mapping relation between applicable vendor ID's for protocols or something.

I guess we could do the protocol mapping, but maybe that's making things too complex. At the moment what happens is the component gets a requires `regex` of all the possible vendor IDs. For example, all Logitech firmware would have:

    <requires compare="regex" version="USB:0x046D|HIDRAW:0x046D">vendor-id</requires>

Although for Dell, HP and Lenovo realistically we'll have a wildcard vendor-id (i.e. no vendor-id in the metadata) for a very long time. In all honesty, it's not the big OEMs I'm worried about, it's the little ODMs that I've not really heard about. They are the ones who are going to be hacked and/or sharing _the_ LVFS username/password on whiteboards.

> but if you look at the vendor specified by Thunderbolt host controller it's actually Dell

Isn't the TBT vendor-id of the form `TBT:0x$(vid)`?

>  I see that on an XPS 7390 the vendor reads as Intel Corporatation (but HIDRAW:0x06CB)

I think lets open it as an issue and debug there.

superm1:
> It's actually on the vendor page now:

Huh, OK I guess I didn't think to look there because once you log in the left bar Home->Vendor brings you to vendor manager page and doesn't show that information.

> <requires compare="regex" version="USB:0x046D|HIDRAW:0x046D">vendor-id</requires>

Ah ok.

> Although for Dell, HP and Lenovo realistically we'll have a wildcard vendor-id (i.e. no vendor-id in the metadata) for a very long time. In all honesty, it's not the big OEMs I'm worried about, it's the little ODMs that I've not really heard about. They are the ones who are going to be hacked and/or sharing the LVFS username/password on whiteboards.

Well the blog post you made makes it sound like it will be 6 months out.  Something I wanted to throw around as an idea.  What if you created two sets of metadata?  
* A legacy one
and
* a fwupd 1.3.6+ one.

In the legacy one don't do new vendor ID stuff.  This will keep existing clients in good shape without changing.  It obviously is temporary LVFS cruft, but you can probably set a threshold or date that those old clients can't keep working.

In the 1.3.6+ one, follow the idea in https://github.com/fwupd/fwupd/issues/391#issuecomment-513218878 to solve that issue where metadata and signature get out of sync.  Make a new metadata payload for the json that a fwupd 1.3.6+ client can take that points to an atomically generated signature and metadata.  This one can be where all the vendor ID stuff gets locked down too then.

hughsie:
> brings you to vendor manager page and doesn't show that information

Would it be useful to show a read-only view of:

Screenshot_2019-12-12 LVFS Vendor Details

> Well the blog post you made makes it sound like it will be 6 months out

Otherwise nobody does anything. Realistically I'll be handing out wildcard vendor-ids like candy at Halloween to avoid breaking HP, Dell and Lenovo specifically.

>  Make a new metadata payload for the json that a fwupd 1.3.6+ client can take that points to an atomically generated signature and metadata

This also allows us to drop the various hacks like https://github.com/fwupd/lvfs-website/blob/master/lvfs/metadata/utils.py#L43

Although, then of course my worry is that we'd end up generating 10 subtly different metadata archives rather than caring about backwards compat, although perhaps that's me being dramatic.

My biggest concern is that we increase the load on the CDN (a.k.a. the bill) doing three requests in the worst case, rather than 2. I suppose best case it's no different. I'll have a think about how to do this properly in fwupd, and then we can work backwards to the LVFS.

superm1:
> Would it be useful to show a read-only view of:

Yes, I think so.

>This also allows us to drop the various hacks like https://github.com/fwupd/lvfs-website/blob/master/lvfs/metadata/utils.py#L43

Yeah, very true.

> Although, then of course my worry is that we'd end up generating 10 subtly different metadata archives rather than caring about backwards compat, although perhaps that's me being dramatic.

Doing this one time certainly doesn't set precedent, and by combining these other things that couldn't be deprecated in one go it sets up LVFS for carrying less cruft.  I think it's entirely reasonable to say when the consumption of the legacy metadata hits say 5% of LVFS overall, sunset it (or even lower if you think it makes sense).

> My biggest concern is that we increase the load on the CDN (a.k.a. the bill) doing three requests in the worst case, rather than 2. I suppose best case it's no different. I'll have a think about how to do this properly in fwupd, and then we can work backwards to the LVFS.

Oh the bill is per number of requests; not per byte?  In that case another approach is to not use a detached signature for the new one, but keep it atomic by putting the signature directly in the metadata file.  Then you could go down from 2 to 1!

superm1:
> Then you could go down from 2 to 1!

Or if it's by byte and request both; you can still have a small json file that just contains the file and expected hash and points to the latest atomically built combination.  Then client can download the json file, if it's changed download the combination, verify the hash from the json and verify the embedded signature.

lvfs/main/templates/dashboard.html
```diff
@@ -24,6 +24,14 @@
 </div>
 {% endif %}
 
+{% if not g.user.vendor.restrictions %}
+<div class="alert alert-warning mb-3" role="alert">
+  This account has no vendor restrictions defined and firmware will not be
+  deployed onto hardware.
+  Please contact the <a href="mailto:info@fwupd.org">LVFS administrator</a>.
+</div>
+{% endif %}
```
superm1: I guess vendors with '*' won't see this (`vendor.restrictions` should be `*`), but it's more for new vendors to make sure that something gets set and tell them it's wrong, right?

hughsie: Correct.

lvfs/search/templates/search.html
```diff
@@ -60,6 +60,15 @@ <h4><a href="{{url_for('vendors.route_list')}}#{{vendor.group_id}}">{{vendor.dis
...
       Is not uploading firmware to the LVFS
     </li>
+{% endif %}
+{% if vendor.restrictions %}
+    <li class="list-group-item">
+      <span class="fas fa-info-circle fs-2 text-info"></span>
+      Restricted to upload firmware for any hardware with vendor ID:
+{% for restriction in vendor.restrictions %}
+        <code>{{restriction.value}}</code>
+{% endfor %}
```
superm1: I think it might be useful to actually indicate which vendors are actually unrestricted here in words rather than expose the `*`.

hughsie: Ohh, good catch. Maybe we just don't show anything for the `[*]` case -- I don't want people to think Dell or Lenovo are doing anything wrong.

hughsie:
> Is that potentially subverting this security model

No, even embargo remotes get the vendor-id applied. The ODM or IHV would be limited to just their own vendor-id for their local testing, and when transferred to the OEM the OEMs `*` vendor Id gets used instead.

superm1:
> > Is that potentially subverting this security model
> 
> No, even embargo remotes get the vendor-id applied. The ODM or IHV would be limited to just their own vendor-id for their local testing, and when transferred to the OEM the OEMs `*` vendor Id gets used instead.

But so potentially that means a restricted `IHV` who has permissions to promote to stable could transfer to `OEM` embargo and promote to stable and the OEM `*` gets used on it.

hughsie:
> could transfer to OEM embargo and promote to stable

Yes. The `Affiliation` ACL of `promote-stable` permission basically lets the ODM to anything the OEM is able to do, i.e. the ODM is representing the OEM almost completely.

Perhaps the vendor id should always be set from the restrictions set on the uploader, i.e. the ODM.

superm1:
> Perhaps the vendor id should always be set from the restrictions set on the uploader, i.e. the ODM.

Yeah when you talk about the intent of the vendor ID, I think that makes sense.  Otherwise all these IHVs are pretty much `*` if they have the right affiliations.

lvfs/search/templates/search.html
hughsie: Fixed.

hughsie:
>  Otherwise all these IHVs are pretty much * if they have the right affiliations

Thanks for the sanity check, it's really appreciated.

PR #460, opened 2020-01-02
Added atom-feed to public device page

hughsie:


hughsie:
Screenshot from 2020-01-02 14-32-41

PR #463, opened 2020-01-10
Move the documentation to Sphinx

hughsie:
This will allow us to produce easy-to-consume formal documentation that various
vendors have requested. It also means we get searching for free too.


hughsie:
The docs are here for easy viewing: https://lvfs.readthedocs.io/

lvfs/templates/private.html
```diff
@@ -290,38 +290,8 @@
...
-            </li>
-            <li class="nav-item {{'active' if request.url_rule.endpoint == 'docs_composite'}}">
-              <a class="nav-link" href="{{url_for('docs.route_composite')}}">Composite Devices</a>
-            </li>
-            <li class="nav-item {{'active' if request.url_rule.endpoint == 'docs_telemetry'}}">
-              <a class="nav-link" href="{{url_for('docs.route_telemetry')}}">Telemetry</a>
-            </li>
-            <li class="nav-item {{'active' if request.url_rule.endpoint == 'docs_affiliates'}}">
-              <a class="nav-link" href="{{url_for('docs.route_affiliates')}}">Affiliates</a>
+              <a class="nav-link" href="{{url_for('docs.route_metainfo')}}">External Documentation</a>
```
superm1: why not link directly to the links for these pages at readthedocs?

docs/source/apply.rst
```diff
@@ -0,0 +1,74 @@
...
+You can start the process by emailing info@fwupd.org with as much information you
+have, or just with questions or for more details.
+
+Information to Supply
+=====================
+
+* The public homepage for this vendor
+* The domain used for email address assigned to this vendor, e.g. ``@realtek.com,@realtek.com.tw``
+* The update protocol are you using, and if it is already supported in fwupd
+* Some kind of proof that you have the required permission to upload to the LVFS
```
superm1: examples probably good here?

docs/source/apply.rst
```diff
@@ -0,0 +1,74 @@
...
+* The ``Vendor ID`` for all hardware uploaded by this vendor (from ``fwupdmgr get-devices`` e.g. ``USB:0x046D``)
+* The reverse DNS AppStream ID namespace prefix you plan to use for all uploaded firmware, e.g. ``com.hp``
+* An assigned "vendor manager" that can create new accounts on the LVFS in the future, and be the primary point of contact
+* If you going to be acting as an ODM to another vendor, e.g. uploading firmware on their behalf
+
+If you are acting as an ODM to another vendor:
```
superm1: ODM or IHV I think here

docs/source/apply.rst
```diff
@@ -0,0 +1,74 @@
...
+
+* Which OEM(s) will you be uploading for?
+* Do you have a contact person for the OEM?
```
superm1: .... If so, who?

docs/source/apply.rst
```diff
@@ -0,0 +1,74 @@
...
+
+New users have to match the username domain glob, so if the value for the vendor
+is ``@realtek.com,@realtek.com.tw`` then ``dave@realtek.com.tw`` could be added by
+the vendor manager -- but ``dave@gmail.com`` would be forbidden.
+
+Trusted Users
+-------------
+
+Vendor groups are created initially as ``untrusted`` which means no users can upload
+to testing and stable.
```
superm1: you mean promote to testing/stable, not upload, right?

docs/source/apply.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Once a valid firmware has been uploaded correctly and been approved by someone
+in the the LVFS admin team we will *unlock* the user account to the ``trusted``
+state which allows users to push firmware to the public remotes.
```
superm1: promote

docs/source/conf.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'LVFS'
+copyright = '2016-2020, Richard Hughes'
+author = 'Richard Hughes'
+
+# The full version, including alpha/beta/rc tags
+release = '1.2.0'
```
superm1: is this LVFS version, or docs version?  If it's LVFS version it's probably better to try to source it or detect it from a git tag.

docs/source/custom-plugin.rst
```diff
@@ -0,0 +1,163 @@
+Custom Protocol
+###############
+
+The fwupd project already supports a `huge number of flashing protocols <https://github.com/fwupd/fwupd/tree/master/plugins>`_,
+everything from standardized protocols like NMVe, ATA, DFU and also a large number
+of *vendor-specfic* protocols like ``logitech_hidpp``, ``synaptics_prometheus`` and ``wacom_raw``.
+
+Most vendors are using a protocol that fwupd already supports, and thus only
```
superm1: You tell me if this most term is true.  It seems like it might not be given how many plugins get added regularly 😜

docs/source/custom-plugin.rst
```diff
@@ -0,0 +1,163 @@
...
+=============
+
+* A PC with Linux (preferably the latest version of Fedora) installed bare metal
+  (i.e. not in VirtualBox or VMWare)
+* Working access to the internet
+* A user account (we'll use `emily` as the example here) with administrator permissions
+
+Building fwupd
+==============
+
```
superm1: I think you might want to prefix this block by say which version of Fedora you are using.  Presumably the dnf commands will simplify when a newer version is what's in the repos for example.

docs/source/custom-plugin.rst
```diff
@@ -0,0 +1,163 @@
...
+    $ cd ~
+    $ sudo dnf -y builddep fwupd
+    $ sudo dnf -y install tpm2-tss-devel
+    $ git clone https://github.com/hughsie/fwupd.git
+    $ cd fwupd
+    $ mkdir build && cd build
+    $ meson ../ --prefix=/home/emily/.root
+    $ ninja
+    $ ninja install # you can press escape and ignore the permissions dialog
```
superm1: I guess this is only true if you are using `fwupdtool`.  You'll need the various PK files otherwise.

docs/source/custom-plugin.rst
```diff
@@ -0,0 +1,163 @@
...
+    $ ./src/fwupdtool --verbose get-devices
+
+Using ``fwupdtool``
+===================
+
+The fwupd project is split into three main components:
+
+* ``fwupd``: The binary that's running all the time in the background, as root
```
superm1: All the time isn't true.  It shuts down by default after some time no?

docs/source/custom-plugin.rst
```diff
@@ -0,0 +1,163 @@
...
+* ``fwupdmgr``: The client tool that end-users use to interact with the running binary, as a normal user
```
superm1: running `fwupd` binary.

docs/source/apply.rst
hughsie: Fixed, thanks.

docs/source/apply.rst
hughsie: gahh, correct. Fixed, thanks.

docs/source/custom-plugin.rst
```diff
@@ -0,0 +1,163 @@
...
+    VID:                     0x2109
+    PID:                     0x105
+     FuFirmwareImage:
+     Data:                  0xc000
+
+Using ``fwupdmgr``
+==================
+
+You can perform the end-to-end tests using a local version of fwupd by first
+calling `ninja install` to get the new plugin installed.
```
superm1: In this instance it will have been important that the PK files got installed (or a packaged fwupd was already on the system and installed these for your).

docs/source/apply.rst
hughsie: Fixed.

docs/source/conf.py
hughsie: That's a good question. I guess LVFS version as this is what goes into the PDF too.

docs/source/custom-plugin.rst
```diff
@@ -0,0 +1,163 @@
...
+
+.. code-block:: bash
+
+    ./src/fwupd --verbose
+
+and in the second you can do:
+
+.. code-block:: bash
+
+    ./src/fwupdmgr --force install VL105.cab
```
superm1: why have `--force` by default?  My experience from sharing a command that included `--allow-older --allow-reinstall` with a team that didn't understand technical details resulted in a test case that included this and papered over an issue.  I've since learned to explain extra options but not give them as sample commands.

docs/source/custom-plugin.rst
hughsie: I think statistically the most common protocol used by vendors is UEFI UpdateCapsule, then NVMe. It's not true it's the most protocol used (that's unifying...) but I don't think that matters here. But sure, `-ETOOMANYPROTOCOLS.`

docs/source/custom-plugin.rst
hughsie: Right. It's `tpm2-tss-devel` that's the problem as it's a new dep compared to the fwupd in F31 gold. When F32 is out we can remove this line. In the meantime I've added a bash comment.

docs/source/custom-plugin.rst
hughsie: It's actually true for fwupdmgr too as long as we:

1) have fwupd installed already systemwide
2) don't start using any new polkit actions

docs/source/custom-plugin.rst
hughsie: Agreed, I've removed the middle bit.

docs/source/custom-plugin.rst
hughsie: Agreed, thanks.

docs/source/custom-plugin.rst
hughsie: Valid point. Removed.

docs/source/custom-plugin.rst
hughsie: At least for Fedora it's very difficult to have a system *without* fwupd installed :)

docs/source/apply.rst
hughsie: Thanks.

lvfs/templates/private.html
hughsie: I'm not 100% all the content is going to stay in the current places. I need to get someone "new" to read the docs and tell me if the structure makes sense.

superm1:
I don't know much about readthedocs themes, but have you looked through to see if there were any that use more of the screen real estate?  My personal opinion I feel like it's a large waste of white pixels reading this on most monitors these days:

image

docs/source/apply.rst
hughsie: Done.

docs/source/conf.py
hughsie: Fixed, although in most cases we're going to get a version like `1.1.5-55-gf46357f` which is perfectly true.

hughsie:
> I don't know much about readthedocs themes

/me also. I'll do some debugging. I think the PDF version is more suitable for wide screens.

PR #465, opened 2020-01-16
Wip/hughsie/shard claim

hughsie:
HP say that if the shard exists, the platform 100% supports SureStart.


hughsie:
Screenshot_2020-01-16 LVFS Firmware Component Details

PR #468, opened 2020-01-23
chipsec: Use UEFIExtract rather than chipset to extract shards

hughsie:
UEFIExtract supports more types of shard (e.g. Intel Microcode) and is much
more stable than using 'chipsec uefi decode' which seems to break every time we
update chipsec.

Although I guess it would be less confusing to rename the plugin, I'm not super
keen on re-running the tests just for the sake of being pedantic.


superm1:
Why not introduce it as a new plugin and keep chipsec around but disabled? Then make new stuff run with this one instead

hughsie:
If we ever do a 'repair' the new plugin will get added to all old firmware. Maybe I can just migrate the dB. We're you thinking uefi-decode or uefi-extract, or even uefi-shards as a name?

plugins/chipsec/report.py
```diff
@@ -1,69 +0,0 @@
...
-                    data['msg'] = data['msg'][:idx].strip()
-
-                if not len(fw.md_prio.shards):
-                    print('No shards: {}'.format(data['msg']))
-                else:
-                    print('Got {} shards: {}'.format(len(fw.md_prio.shards), data['msg']))
-
-            # unallocate the cached blob as it's no longer needed
-            fw.blob = None
-            writer.writerow(data)
```
globotree: #468 
Duplicate of #`

superm1:
> If we ever do a 'repair' the new plugin will get added to all old firmware. Maybe I can just migrate the dB. We're you thinking uefi-decode or uefi-extract, or even uefi-shards as a name?

Extract and shards both seem to convey more of whats going on.  Decode to me makes it sound like decompile.

hughsie:
I'll run this on all the firmware locally tonight and see if anything explodes. Thanks for the review.

hughsie:
We also probably ought to wait for https://github.com/LongSoft/UEFITool/issues/191 too

PR #470, opened 2020-01-24
Add a plugin to check for the ThinkPwn vulnerability

hughsie:
I ran this on the staging server and found nothing (not unexpected, thinkpwn is quite old now). It took 12 hours to decompile all the binaries! I'm not sure it it makes sense to have turned on for the LVFS by default, but it's certainly another tool that might be useful. Ideas welcome.

### DIFF
--- a/plugins/thinkpwn/__init__.py
+++ b/plugins/thinkpwn/__init__.py
@@ -1,0 +1,207 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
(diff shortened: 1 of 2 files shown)


superm1:
Rather than checking on all EFI files since the exploit is specific to 7C79AC8C-5E6C-4E3D-BA6F-C260EE7C172E maybe just check those binaries if they show up?

hughsie:
Ahh, I guess that makes sense. There are no firmwares on the LVFS that include 7C79AC8C-5E6C-4E3D-BA6F-C260EE7C172E, so maybe there isn't much point for this now.

PR #471, opened 2020-01-24
Allow adding component claims based on the hash of a shard

hughsie:
This allows us to add negative security claims to components that contain shards
with known security problems, for instance older versions of UsbRt.

### DIFF
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -1296,6 +1296,24 @@ def _calculate_entropy(s):
     e_x = [- p_x * math.log(p_x, 2) for p_x in probabilities]
     return sum(e_x)
 
+class ComponentShardClaim(db.Model):
+
+    # sqlalchemy metadata
+    __tablename__ = 'component_shard_claims'
+    __table_args__ = {'mysql_character_set': 'utf8mb4'}
+
+    component_shard_claim_id = Column(Integer, primary_key=True)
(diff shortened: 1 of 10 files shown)


hughsie:
Screenshot_2020-01-24 LVFS Component Shard Claims
Screenshot_2020-01-24 LVFS Component Shard Checksums
Screenshot_2020-01-24 LVFS Component Shard Components
Screenshot_2020-01-24 LVFS Component Shard Details

superm1:
Correlating what is in one vendors image versus another brings up (what I think is) a good idea.  How about uploading some "reference" Tiano Core/OVMF images from the recent git tags to LVFS and comparing to those?  You can then tell exactly what has changed from the different EDK II releases.
I guess technically you  can probably also add links to actual code when you know it hasn't changed.

hughsie:
> uploading some "reference" Tiano Core/OVMF images

That's a good idea. I'll talk to Brian and see if we can get something more automated.

superm1:
I'm curious are you seeing hashes the same across vendors?  I wonder if EDK hardcodes compiler and flags enough for reproducible builds type stuff to work.

hughsie:
> I'm curious are you seeing hashes the same across vendors?

Ya, for things like Shell.efi a lot of vendors just use the Intel reference version.

hughsie:
> nothing stands out to me

Other than the complete lack of tests :) I've added some now, nothing problematic found.

PR #472, opened 2020-01-24
Wip/hughsie/cron args

hughsie:


cron.py
```diff
@@ -288,8 +288,11 @@ def _yara_query_all():
 
 def _check_firmware():
 
-    # make a list of all the tests that need running
-    tests = db.session.query(Test).filter(Test.started_ts == None).all()
+    # make a list of the first few tests that need running
+    tests = db.session.query(Test)\
+                      .filter(Test.started_ts == None)\
+                      .order_by(Test.scheduled_ts)\
+                      .limit(50).all()
```
superm1: Is there actually more than 50 tests running on a given firmware though?  Or just preparing for future?  Or is this because of yara queries that show up as tests?

hughsie: Not 50 per firmware, 50 in the global queue. If I add a new plugin (like thinkpwn) then you get one fwcheck process stuck for 12 hours and no other tests (like uefi-capsule) are allowed to complete. Which would make the OEMs unhappy...

superm1: Oh I see, OK.

PR #474, opened 2020-01-27
Add Claim as a top level object

hughsie:
At the moment we just add a key=value attribute to each object that requires a
claim. This doesn't scale, and means we can never change claim icons from
something like 'info' to 'warning' without a database query that touches lots
of records in different tables. We also want to have a URL we can point the
user to for more information, for instance describing what BIOSGuard is,
and why it's a good thing.

To do this, create enumerated claim kinds that can be referenced by ID and
modified as required. We only really have a couple of existing claim types,
so we don't even need to worry too much about migrating data.

### DIFF
--- a/contrib/pylintrc
+++ b/contrib/pylintrc
@@ -7,7 +7,7 @@ unsafe-load-any-extension=yes
 [MESSAGES CONTROL]
 
 # Disable the message, report, category or checker with the given id(s)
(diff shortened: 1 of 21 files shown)


hughsie:
Screenshot_2020-01-27 LVFS Component Shard Claims
Screenshot_2020-01-27 LVFS Claims

hughsie:
I'll merge this now, as it's blocking another branch -- but if you do see any issues please let me know...

PR #478, opened 2020-01-29
Detect old versions of microcode

hughsie:
Add a waivable test failure if the firmware contains a known-old version of
microcode. Use the built-in database of shard attributes for firmware already
on the LVFS, and fall back to the static database from MCExtractor.

Use the CPUID and platform to match the 'same' processor.

### DIFF
--- a/plugins/microcode/__init__.py
+++ b/plugins/microcode/__init__.py
@@ -1,0 +1,128 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: GPL-2.0+
+#
+# pylint: disable=no-self-use
+
+import datetime
+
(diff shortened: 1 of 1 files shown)


hughsie:
Screenshot_2020-01-29 LVFS Runtime Tests

hughsie:
Could we maybe restrict this to firmware not in stable or testing? Or perhaps we filter out anything uploaded before the ucode date? Or both?

superm1:
Well I'm torn here; perhaps it's relevant information to tell someone that a BIOS uploaded a year ago that has an old microcode and that a new BIOS uploaded a week ago doesn't.

hughsie:
I've neutered this plugin quite a lot to reduce the false positive rate.

Screenshot_2020-01-30 LVFS Runtime Tests

What it's doing now is only failing the test if there's another component with the same appstream_id, and a _lower_ firmware version number with a *higher* microcode version. This detect about 9 cases where the microcode has been accidentally downgraded when fixing two issues "in parallel".

PR #480, opened 2020-01-30
Add a plugin to identify old microcode versions

hughsie:
This uses the database provided by MCExtractor and may produce false positives
as some microcode updates might be vendor or model specific.

Screenshot_2020-01-30 LVFS Firmware Component Details

### DIFF
--- a/plugins/microcode-mcedb/__init__.py
+++ b/plugins/microcode-mcedb/__init__.py
@@ -1,0 +1,103 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: GPL-2.0+
+#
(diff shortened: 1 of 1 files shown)


hughsie:
Note, this also tags almost all firmware on the LVFS older than a year with that warning. Whilst correct (and perhaps useful if pondering downgrading) I don't know if it would be unexpected for vendors. Maybe vendors/users don't look at old firmware versions?

superm1:
How about if you show the message on older firmware you specifically show the date of the release of the newer microcode in the message?

hughsie:
> show the date of the release of the newer microcode in the message?

What could the user do with that information?

superm1:
Well I was going to say pick a firmware released after that, but maybe that's also not useful to show

hughsie:
This is the same class of information as "Contains CompuTrace" -- the user can't do anything at all about it....

superm1:
True, okay.

PR #483, opened 2020-01-31
Make the LVFS username case insensitive

hughsie:


hughsie:
Deployed onto stable, thanks.

PR #484, opened 2020-01-31
Replace problems with non-committed claims

hughsie:
Both the Problem() and Claim() objects were doing very similar things; tracking
information with small amounts of description about the firmware.
Merge them to simplify the code base and to remove a lot of boilerplate.

Also, move to Font Awesome glyphs to make everything more consistent.


lvfs/models.py
```diff
@@ -2111,25 +2098,47 @@ def problems(self):
         # does the firmware have any warnings
         problems = []
```
superm1: perhaps this function should be renamed to claims() too?

hughsie: Good question. It actually exposes another insecurity I have with this patch; what's the difference between `autoclaims` and `problems`? They're both transient `Claim()`s generated at runtime, but one allows the move to stable, and the other doesn't. I think we could add a boolean `failable` to the `Claim()` object and just move problems to `autoclaims`.

Thinking the other way, maybe it does make sense from an efficiently point to have them as different "getters" and so this isn't a problem. Maybe `autoclaim` -> `security_claims` and `problems` -> `problem_claims`?

I'm not super fond of the autoclaims name tho. Other ideas welcome.

superm1: For another idea you could just move everything to the term "attestations" and add booleans for waivable/failable?

hughsie: I spent an hour or so going down the rabbithole unifying problems and autoclaims... but everywhere we used the shared property we were filtering out the `failable` either one way or another, i.e. it was pointless sharing a common property.

My conclusion now is that re-using the object is a good idea, but trying to unify the auto-added properties less so. I'll think more about how the update details claims are added as that's a bit yukky still.

PR #486, opened 2020-02-03
Document more of the LVFS claims and tests behaviour

hughsie:


docs/source/claims.rst
```diff
@@ -0,0 +1,194 @@
...
+Some microcode is vendor-specific, so for instance Lenovo might create an update
+on the LVFS that updates the version of microcode of CPUID 0x906ec from 0xd2 to 0xd3.
+Although Dell might be using the same processor, the motherboard hardware is not
+affected and no update will be prepared.
+
+Computrace
+==========
+
+When a computer equipped with Computrace is reported stolen, the firmware agent
+notifies the monitoring center, allowing the Absolute Theft Recovery Team to
```
superm1: Attempts to notify.  I would like to think that without web access during pre-boot and running Linux as an operating system no resident code will be able to accomplish this.

docs/source/claims.rst
```diff
@@ -0,0 +1,194 @@
...
+
+HP Sure Start
+=============
+
+Every time the PC powers on, HP Sure Start automatically validates the integrity
+of the BIOS code to help ensure that the PC is safeguarded from malicious attacks.
+
+Once the PC is operational, runtime intrusion detection constantly monitors memory.
+In the case of an attack, the PC can self-heal using an isolated "golden copy"
+of the BIOS in less than a minute.
```
superm1: Probably not worth documenting any further but doesn't that mean that if you manage to attack the *isolated* golden copy it self heals to the tampered firmware?  What protects this isolated copy from being tampered?  Presumably it's on a bus somewhere.

docs/source/claims.rst
```diff
@@ -0,0 +1,194 @@
...
+
+Typically, this is implemented by blocking SMM writes to the SPI flash chip.
+
+Intel Boot Guard
+================
+
+Intel Boot Guard is a technology introduced by Intel in the 4th Intel Core
+generation (Haswell) to verify the boot process.
+This is accomplished by flashing the public key of the BIOS signature into the
+field programmable fuses of the CPU itself, typically during the manufacturing process.
```
superm1: It's worth mentioning these are a one time action, once the fuses are blown can't be changed.

docs/source/claims.rst
```diff
@@ -0,0 +1,194 @@
...
+Devices supporting verified firmware either allow the host to read back the written
+firmware at a later time, or will return a internally-calculated checksum.
+
+This allows users to verify that devices have not been tampered with, which may
+even be a concern before first use due to supply chain attacks.
+
+For UEFI firmware, although the firmware capsule is signed by the OEM or ODM,
+we can’t reliably read the SPI EEPROM from userspace.
+We can get a hash of the firmware, or rather, a hash derived from the firmware.
+This is stored in the TPM chip as ``PCR0``.
```
superm1: These few sentences probably don't use 'we'.  And it's probably worth explaining further this is a series of hashes and PCR0 only corresponds to the final hash.

docs/source/claims.rst
```diff
@@ -0,0 +1,194 @@
...
+
+To list the various ``PCRs`` on the running system you can use
+``cat /sys/class/tpm/tpm0/pcrs`` for TPMs using protocol 1.2, or
+``tpm2_listpcrs`` for TPMs using later protocol versions.
```
superm1: No one knows if this will change in the next generation of TPMs.  Probably better to just mention 2.0 explicitly.

docs/source/claims.rst
```diff
@@ -0,0 +1,194 @@
...
+
+    <releases>
+      <release date="2019-01-08" urgency="high" version="1.2.3">
+        <checksum type="sha1" target="device">ce7dd93006be33bcce1a1965cb69634bd0a0fe35</checksum>
+        <checksum type="sha256" target="device">c479988947653b403d6a4ebe366cc60eaf7b6e147bd058fb524be418890655c9</checksum>
+      </release>
+    </releases>
+
+Multiple possible device checksums can also be set using the admin console of
+the LVFS:
```
superm1: I think it's worth mentioning that there are valid reasons for multiple PCR0 values in this sentence.

docs/source/claims.rst
hughsie: OK.

docs/source/claims.rst
hughsie: Done, I've used TXT as the example.

docs/source/claims.rst
hughsie: OK.

docs/source/claims.rst
hughsie: > if you manage to attack the isolated golden copy

There are precious little documentation on the golden copy. During a capsule update we must be able to update the golden copy somehow, otherwise we'd undo the update each boot.

docs/source/claims.rst
hughsie: > I would like to think that without web access during pre-boot

Ohh it's horrendous how it works. The NTFS driver used to modify the Windows-provided `chkdisk.exe` and inject stuff into `iexplore.exe`. Now I believe it uses the PE loading stuff provided by Windows 10, which is how Lenovo also install the OEM tools and drivers on a "clean" Win10 install.

> running Linux as an operating system

Good point will add.

hughsie:
> In each claims section maybe add a little shield next to the title for + / - / neutral

No can do, we link to the same "section" from `signed-firmware` and `no-signed-firmware` -- deliberately so to avoid writing the same thing but negated.

PR #496, opened 2020-02-11
Log any new microcode for manual uploading to mce.db

hughsie:
See https://github.com/platomav/MCExtractor/issues/30 for details.

### DIFF
--- a/plugins/microcode-mcedb/__init__.py
+++ b/plugins/microcode-mcedb/__init__.py
@@ -15,6 +15,7 @@ from lvfs import db
 from lvfs.pluginloader import PluginBase, PluginSettingBool, PluginSettingText
 from lvfs.models import Test, Claim
 from lvfs.models import _get_datestr_from_datetime
+from lvfs.util import _get_shard_path
 
 class Plugin(PluginBase):
     def __init__(self):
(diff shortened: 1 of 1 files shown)


plugins/microcode-mcedb/__init__.py
```diff
@@ -96,6 +97,12 @@ def _run_test_on_shard(self, test, shard):
                               .first()
             if claim:
                 shard.md.add_claim(claim)
+        else:
+            c.execute('SELECT version FROM Intel WHERE cpuid=? AND '
+                      'platform=? AND version==? LIMIT 1',
+                      (cpuid, platform, version,))
+            if not c.fetchone():
+                print('microcode NOT FOUND, please upload: {}'.format(_get_shard_path(shard)))
```
superm1: I think you want to actually have this action to add to the database run when the firmware goes to stable, not when the microcode test runs.  Otherwise you might inadvertently disclose the existence of new microcode before it's been in the _wild_ (so to say).

hughsie: Perhaps we should do two things:

 * email the right person rather than just printing something in the logs
 * only run the entire plugin (a new plugin?) when the firmware is in stable

I'll have a think and open a new, better PR.

PR #497, opened 2020-02-13
Wip/hughsie/mcedb log

hughsie:


lvfs/templates/email-microcode-new.txt
```diff
@@ -0,0 +1,13 @@
+Hi,
+
+We wanted to tell you an interesting CPU microcode has been uploaded to the LVFS and the firmware has now been made public.
+
+cpuid: {{shard.get_attr_value('cpuid')}}
+platform: {{shard.get_attr_value('platform')}}
+version: {{shard.get_attr_value('version')}}
+
+The microcode can be downloaded here: {{url_for('shards.route_download', component_shard_id=shard.component_shard_id, _external=True)}}
```
superm1: I thought we made shards something you need to be logged in to download?

plugins/microcode-notify/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+                  (cpuid, platform, version,))
+        if not c.fetchone():
+            test.add_pass('Not found in MCEdb')
+            email_address = self.get_setting('microcode_notify_address')
+            if email_address:
+                for addr in email_address.split(','):
+                    send_email('[LVFS] Microcode not found in MCEdb',
+                               addr,
+                               render_template('email-microcode-new.txt',
+                                               shard=shard))
```
superm1: I thought I Left a comment for this before, but maybe I forgot.  I think you need to keep a record of which microcode has been notified but isn't yet in the MCE.db so you don't notify 50 times when a new CPU microcode comes out on all the systems in LVFS.

hughsie:

PR #500, opened 2020-02-14
Support Lenovo-specific security tags

hughsie:


superm1:
If you're adding support for the lenovo ones, probably should also do the same for Dell and Intel (who also make their own pages).  The Dell ones are prefixed DSA and go here:
https://www.dell.com/support/security/en-us/

Intel ones go here: https://www.intel.com/content/www/us/en/security-center/default.html

hughsie:
> should also do the same for Dell

Have you got an example of a URL we can construct for Dell? e.g. `https://www.intel.com/content/www/us/en/security-center/advisory/INTEL-SA-00341.html` is the one for a given INTEL-SA tag.

superm1:
For Dell I think you need to actually do a query on that page I linked to get the right URL.  For example it gives out these two which are two very different URLs.

https://www.dell.com/support/article/us/en/04/sln317156/dsa-2019-089-dell-emc-server-platform-security-advisory-for-intel-sa-00233?lang=en

https://www.dell.com/support/security/en-us/details/DOC-110653/DSA-2020-034-RSA-Identity-Governance-and-Lifecycle-Security-Update-for-SUSE-Linux-Enterprise-Serv

hughsie:
> For Dell I think you need to actually do a query on that page

Ew. No hyperlink for you then :)

superm1:
> Ew. No hyperlink for you then :)

Maybe open a link in new tab for https://www.dell.com/support/security/en-us/ so they can at least search it?

I would think there is probably a way to put it in the URL too, but I'm not super savvy on that.

hughsie:
> If you're adding support for the lenovo ones, probably should also do the same for Dell and Intel

Done.

PR #501, opened 2020-02-14
Actually include LVFS::UpdateProtocol in the metadata

hughsie:
We are already checking this in fwupd when installing the downloaded file, but
for other tools that are just consuming the metadata it's very useful to have
it in the catalog too.


superm1:
Yep, LGTM

PR #504, opened 2020-02-15
Add a script to show the download count on an external LED panel

hughsie:
This is really a toy, and is only used to make myself feel warm and fuzzy.

IMG_20200215_134629

@superm1 if you want me to send you some hardware, let me know -- I made one spare just in case. :)

### DIFF
--- a/contrib/led/90-lvfs-led.rules
+++ b/contrib/led/90-lvfs-led.rules
@@ -1,0 +1,1 @@
+ACTION=="add", KERNEL=="ttyUSB*", TAG+="systemd", ENV{SYSTEMD_WANTS}+="lvfs-led@%k.service"
--- a/contrib/led/Makefile
+++ b/contrib/led/Makefile
@@ -1,0 +1,4 @@
+install:
+	install 90-lvfs-led.rules /etc/udev/rules.d
+	install lvfs-led@.service /etc/systemd/system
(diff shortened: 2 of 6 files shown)


superm1:
Ha, nice! Appreciate the thought, but I probably can skip the hardware this go around for it. You have a backup in case it breaks.

contrib/led/90-lvfs-led.rules
```diff
@@ -0,0 +1 @@
+ACTION=="add", KERNEL=="ttyUSB*", TAG+="systemd", ENV{SYSTEMD_WANTS}+="lvfs-led@%k.service"
```
superm1: Anything more informative to tag? This would set off for example a zwave controller too

hughsie:
> You have a backup in case it breaks

Well, my daughter is currently making it scroll her name, so.... :)

contrib/led/90-lvfs-led.rules
hughsie: I guess I could match on idVendor=0403, idProduct=6001 -- but it's really the generic FTDI chip in the cable. Out of interest, what's the vendor ID of the other device?

superm1: I was mixing up another dongle, this one is ttyACM0.

PR #518, opened 2020-02-24
Add category information to the metadata

hughsie:
The frontend should not have to do string searching to get the *known* category
type. We can't use ther plain 'categories' until we switch to using `.name` for
the `<name>` rather than `.name_with_category` -- which we can't realistically
do until more of the clients downloading firmware have been updated to newer
versons.

Basically, err on odd rather than breaking things.

### DIFF
--- a/lvfs/metadata/utils.py
+++ b/lvfs/metadata/utils.py
@@ -225,22 +225,22 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
                 sz.set('type', 'download')
                 sz.text = str(md.release_download_size)
 
-        # deliberately not including <category> here until 2020-01-01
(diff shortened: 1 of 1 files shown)


lvfs/metadata/utils.py
```diff
@@ -225,22 +225,22 @@ def _generate_metadata_kind(filename, fws, firmware_baseuri='', local=False):
                 sz.set('type', 'download')
                 sz.text = str(md.release_download_size)
 
-        # deliberately not including <category> here until 2020-01-01
```
superm1: Well a little late, but that seems OK :)

hughsie: Looking at the LVFS stats for this month, I was woefully optimistic about how quickly people upgrade :/

PR #519, opened 2020-02-28
Wip/hughsie/sign

hughsie:


lvfs/firmware/routes.py
```diff
@@ -179,6 +179,29 @@ def route_nuke(firmware_id):
     flash('Firmware nuked', 'info')
     return redirect(url_for('firmware.route_firmware'))
 
+@bp_firmware.route('/<int:firmware_id>/resign')
+@login_required
+@admin_login_required
+def route_resign(firmware_id):
+    """ Delete a firmware entry and also delete the file from disk """
```
superm1: need to update comment

lvfs/firmware/routes.py
```diff
@@ -179,6 +179,29 @@ def route_nuke(firmware_id):
...
+
+    # check firmware exists in database
+    fw = db.session.query(Firmware).filter(Firmware.firmware_id == firmware_id).first()
+    if not fw:
+        flash('No firmware {} exists'.format(firmware_id), 'danger')
+        return redirect(url_for('firmware.route_firmware'))
+
+    # firmware is not deleted yet
```
superm1: eh?

hughsie: Fixed, thanks.

lvfs/firmware/routes.py
hughsie: Fixed, thanks.

PR #520, opened 2020-02-28
Create Jcat files in archives and for metadata

hughsie:
A Jcat file can be used to store GPG, PKCS-7 and SHA-256 checksums for multiple
files. This allows us to sign a firmware or metadata multiple times (perhaps
by the OEM and also then the LVFS) which further decentralizes the trust model
of the LVFS.

The Jcat format was chosen as the microsoft catalog format is nonfree and not
documented. We also don't want to modify an existing .cat file created from WU
as this may make it unsuitable to use on Windows.

More information can be found here: https://github.com/hughsie/libjcat


cron.py
```diff
@@ -52,18 +53,58 @@ def _regenerate_and_sign_metadata(only_embargo=False):
...
+        invalid_fns.append(fn_xmlgz)
+
+        # create Jcat item with SHA256 checksum blob
+        jcatitem = jcatfile.get_item(r.filename)
+        jcatitem.add_blob(JcatBlobSha256(blob_xmlgz))
+
+        # write each signed file
+        for blob in ploader.metadata_sign(blob_xmlgz):
+
+            # add GPG only to archive for backwards compat with older fwupd
```
superm1: How old are we talking at this point?  The numbers still support trying to make that work?

hughsie: It's older in "any version of fwupd that doesn't know what a jcat file is" -- which is basically all of them at this point :)

superm1: 😃

jcat/jcatfile.py
```diff
@@ -0,0 +1,62 @@
...
+import json
+import gzip
+
+from . jcatitem import JcatItem
+
+class NotSupportedError(NotImplementedError):
+    pass
+
+class JcatFile():
+    """An object representing a Jcat archive """
```
superm1: Why re-invent this twice?  Why not actually use libjcat + introspection?

hughsie: That's a good question. Primarily because it's pure python and thus easier to deploy, and second I thought it would be good to have two implementations to verify each other.

hughsie:
Also, because this untangles the metadata modified path, all the cache invalidation happens in the same instant, rather than now where we invalidate the metadata, sign the file, then invalidate the signing files. This gives a couple of seconds where the mirror is out of sync.

cron.py
```diff
@@ -52,18 +53,58 @@ def _regenerate_and_sign_metadata(only_embargo=False):
     if not remotes:
         return
 
+    # create Jcat file
+    jcatfile = JcatFile()
```
hughsie: Maybe we do this after the mkdir?

cron.py
```diff
@@ -88,14 +129,47 @@ def _sign_fw(fw):
...
-            ploader.archive_sign(cabarchive, cabarchive[md.filename_contents])
+
+            # create Jcat item with SHA256 checksum blob
+            cabfile = cabarchive[md.filename_contents]
+            jcatitem = jcatfile.get_item(md.filename_contents)
+            jcatitem.add_blob(JcatBlobSha256(cabfile.buf))
+
+            # sign using plugins
+            for blob in ploader.archive_sign(cabfile.buf):
+                fn_blob = md.filename_contents + '.' + blob.filename_ext
```
hughsie: move this inside the `JcatBlobKind.GPG` conditional

cron.py
```diff
@@ -88,14 +129,47 @@ def _sign_fw(fw):
...
+                # add to Jcat file too
+                jcatitem.add_blob(blob)
+
         except KeyError as _:
             raise NotImplementedError('no {} firmware found'.format(md.filename_contents))
 
+    # sign all the metainfo.xml files
+    for fn_xml in cabarchive:
+        if not fn_xml.endswith('.metainfo.xml'):
+            continue
```
hughsie: we can use `filename_xml` now

jcat/__init__.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018-2019 Richard Hughes <richard@hughsie.com>
```
hughsie: 2020

jcat/tests/jcat_test.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
```
hughsie: 2020

jcat/jcatblob.py
```diff
@@ -0,0 +1,72 @@
...
+    def filename_ext(self):
+        if self.kind == JcatBlobKind.SHA256:
+            return 'sha256'
+        if self.kind == JcatBlobKind.GPG:
+            return 'asc'
+        if self.kind == JcatBlobKind.PKCS7:
+            return 'p7b'
+        return None
+
+class JcatBlobSha256(JcatBlob):
```
hughsie: maybe also have `JcatBlobText`?

plugins/sign-sigul/__init__.py
```diff
@@ -65,31 +65,18 @@ def settings(self):
...
 
         # create the detached signature
-        blob_asc = _sigul_detached_sign_data(cabfile.buf,
+        blob_asc = _sigul_detached_sign_data(blob,
                                              self.get_setting('sign_sigul_config_file', required=True),
                                              self.get_setting('sign_sigul_firmware_key', required=True))
-
-        # add it to the archive
-        cabarchive[detached_fn] = CabFile(blob_asc.encode('utf-8'))
+        return JcatBlob(JcatBlobKind.GPG, blob_asc.encode('utf-8'))
```
hughsie: JcatBlobText?

PR #523, opened 2020-03-02
Wip/hughsie/filename xml

hughsie:


cron.py
```diff
@@ -114,7 +116,53 @@ def _sign_fw(fw):
...
+            except ET.XMLSyntaxError as e:
+                print('FAILED to load {}:{} ({})'.format(absolute_path, fn, e))
+                continue
+            appstream_id = component.xpath('id')[0].text.replace('/', '_').strip()
+            if md.appstream_id == appstream_id:
+                md.filename_xml = fn
+                print('repairing {} filename_xml to {}'.format(md.appstream_id, md.filename_xml))
+                break
+
+        # hope the archive only has one filename
```
superm1: I think this comment for this block is a little disingenuous.  I had to look back at the code before to confirm it.  This is actually a recovery for invalid appstream ID

lvfs/models.py
```diff
@@ -1629,6 +1630,15 @@ def problems(self):
                                               'not explain what it fixes.\n'
                                               'This also can be set in the .metainfo.xml file.'))
 
+        # possible for broken legacy firmware
```
superm1: This shouldn't be possible for anything new though it seems, just worried about stuff that's already on LVFS?

cron.py
hughsie: Well, yes, I suppose it's really `# we changed the AppStream ID on the LVFS, lets hope there's only one metainfo.xml file`

lvfs/models.py
hughsie: It's the cover the few minutes between:

 * deploy commit 1
 * fix up database (~15 minutes)
 * deploy commit 2

The 2nd commit actually removes the problem as the field becomes an error if not set.

PR #524, opened 2020-03-02
Wip/hughsie/checksum contents sha256

hughsie:


lvfs/models.py
```diff
@@ -1554,7 +1555,7 @@ def autoclaims(self):
                                     icon='warning',
                                     summary='Firmware cannot be verified after flashing',
                                     url='https://lvfs.readthedocs.io/en/latest/claims.html#verified-firmware'))
-        if self.checksum_contents:
+        if self.checksum_contents_sha1:
```
superm1: Don't you need another one of these for sha256 now?

hughsie: I don't think so; we're kinda (ab?)using the fact that only stuff signed has a `checksum_contents` of either type. There are a few ways we could detect this; I'll have a play and do a follow up PR for a more obvious check.

PR #525, opened 2020-03-02
Rewrite the metainfo when signing the firmware

hughsie:
One of the common tripping points for OEMs and ODMs is that the uploaded
firmware.metainfo.xml does not get rewritten if requirements are added or other
changes are made on the LVFS UI.

Although we use the metadata to know what firmware file to download, we use
the 'as uploaded' metadata when actually installing the firmware. By rewriting
the metainfo we can ensure the two information sources match.

This does of course change the downloaded file checksum, but this is not being
served by the CDN.

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -9,9 +9,12 @@
 
 import os
 import sys
+import difflib
 import hashlib
 import datetime
 import yara
+
+from lxml import etree as ET
 
 from flask import render_template
 
@@ -25,7 +28,7 @@ from lvfs.models import Remote, Firmware, Vendor, Client, AnalyticVendor, User, 
 from lvfs.models import AnalyticFirmware, Useragent, UseragentKind, Analytic, Report, Metric
 from lvfs.models import ComponentShard, ComponentShardInfo, Test, Component, Category, Protocol, FirmwareEvent
 from lvfs.models import _get_datestr_from_datetime
-from lvfs.metadata.utils import _metadata_update_targets, _metadata_update_pulp
+from lvfs.metadata.utils import _metadata_update_targets, _metadata_update_pulp, _generate_metadata_mds
 from lvfs.util import _event_log, _get_shard_path, _get_absolute_path
 from lvfs.upload.uploadedfile import UploadedFile, MetadataInvalid
 
@@ -118,6 +121,12 @@ def _regenerate_and_sign_metadata(only_embargo=False):
     for r in remotes:
         _event_log('Signed metadata %s' % r.name)
 
+def _show_diff(blob_old, blob_new):
+    fromlines = blob_old.decode().replace('\r', '').split('\n')
+    tolines = blob_new.decode().split('\n')
+    diff = difflib.unified_diff(fromlines, tolines)
+    print('\n'.join(list(diff)[3:]))
+
 def _sign_fw(fw):
 
     # load the .cab file
@@ -156,9 +165,20 @@ def _sign_fw(fw):
         except KeyError as _:
             raise NotImplementedError('no {} firmware found'.format(md.filename_contents))
 
-    # sign all the metainfo.xml files
+    # rewrite the metainfo.xml file to reflect latest changes and sign it
     for md in fw.mds:
-        blob_xml = cabarchive[md.filename_xml].buf
+
+        # write new metainfo.xml file
+        component = _generate_metadata_mds([md], metainfo=True)
+        blob_xml = b'<?xml version="1.0" encoding="UTF-8"?>\n' + \
+                   ET.tostring(component,
+                               encoding='UTF-8',
+                               xml_declaration=False,
+                               pretty_print=True)
+        _show_diff(cabarchive[md.filename_xml].buf, blob_xml)
+        cabarchive[md.filename_xml].buf = blob_xml
+
+        # sign it
         jcatitem = jcatfile.get_item(md.filename_xml)
         jcatitem.add_blob(JcatBlobSha256(blob_xml))
         for blob in ploader.archive_sign(blob_xml):
--- a/lvfs/metadata/utils.py
+++ b/lvfs/metadata/utils.py
@@ -12,6 +12,7 @@ import gzip
 import hashlib
 
 from collections import defaultdict
+from datetime import date
 from lxml import etree as ET
 
 from lvfs import db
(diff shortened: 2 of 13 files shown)


cron.py
```diff
@@ -156,9 +166,16 @@ def _sign_fw(fw):
...
-    # sign all the metainfo.xml files
+    # rewrite the metainfo.xml file to reflect latest changes and sign it
     for md in fw.mds:
-        blob_xml = cabarchive[md.filename_xml].buf
+
+        # write new metainfo.xml file
+        component = _generate_metadata_mds([md], metainfo=True)
+        blob_xml = ET.tostring(component, encoding='UTF-8', xml_declaration=True, pretty_print=True)
+        _show_diff(cabarchive[md.filename_xml].buf, blob_xml)
+        cabarchive[md.filename_xml].buf = blob_xml
```
superm1: So something to consider (not sure if it's important) is that any comments left in the XML will probably get lost from this process.

cron.py
```diff
@@ -118,6 +121,13 @@ def _regenerate_and_sign_metadata(only_embargo=False):
     for r in remotes:
         _event_log('Signed metadata %s' % r.name)
 
+def _show_diff(blob_old, blob_new):
+    print(blob_old.decode().replace('\r', ''))
+    fromlines = blob_old.decode().replace('\r', '').split('\n')
+    tolines = blob_new.decode().split('\n')
+    diff = difflib.unified_diff(fromlines, tolines)
+    print('\n'.join(list(diff)[3:]))
```
superm1: presumably you don't want this logged _every_ time the file changes?  More of a debugging feature?

lvfs/metadata/utils.py
```diff
@@ -19,146 +19,141 @@
...
-        # until all front ends support <category> and <name_variant_suffix> append both */
+def _generate_metadata_mds(mds, firmware_baseuri='', local=False, metainfo=False):
+
+    # assume all the components have the same parent firmware information
+    md = mds[0]
+    component = ET.Element('component')
+    component.set('type', 'firmware')
+    ET.SubElement(component, 'id').text = md.appstream_id
+
+    # until all front ends support <category> and <name_variant_suffix> append both */
```
superm1: good time to revisit if this is still a big part of the fwupd population?

cron.py
superm1: And also any extra XML fields LVFS doesn't understand I guess would go away too.  But that one probably doesn't matter.

hughsie: > any comments left in the XML

This was actually deliberate; I didn't want to leak anything sensitive (e.g. how the GUIDs are being constructed by the Instance ID) -- if you think this is wrong I can certainly preserve them.

> any extra XML fields LVFS doesn't understand I guess would go away too

I think that's fine. It's **X**ML, but it's *our* version of XML :)

cron.py
hughsie: It was useful to go into the logs, if nothing else to see what was there originally as uploaded. It's less useful on the "first" rewrite but it's kinda helpful when you're just adding a requirement for example.

cron.py
superm1: > This was actually deliberate; I didn't want to leak anything sensitive (e.g. how the GUIDs are being constructed by the Instance ID) -- if you think this is wrong I can certainly preserve them.

Hmm, I guess we do hide it by default in the tool unless root so that does make it an argument it's sensitive.  OK i'm on your side, rewriting it makes sense.

lvfs/metadata/utils.py
hughsie: Want to be somewhat depressed:

Screenshot_2020-03-02 LVFS Analytics

superm1: 😭 😿 😞 
0.8.x?  Hmm I'd guess it's mostly Ubuntu 16.04 stuff still.
It was a challenge enough to get 1.2.x into Ubuntu 18.04.  If that's correct, then I think it's reasonable to kill it at the 5 year mark (that's the same as the LTS support cycle for Ubuntu 16.04).  Which sadly means next spring.

cron.py
hughsie: For instance, this is what it does to one of your firmwares:

    -<?xml version="1.0" encoding="UTF-8"?>
    +<?xml version='1.0' encoding='UTF-8'?>
     <component type="firmware">
       <id>com.dell.uefi417d4c2a.firmware</id>
    +  <name>XPS 15 7590/Precision 5540</name>
    +  <summary>Firmware for the Dell XPS 15 7590/Precision 5540</summary>
    +  <description>
    +    <p>Updating the system firmware improves performance.</p>
    +  </description>
       <provides>
         <firmware type="flashed">417d4c2a-87d1-4d7c-bcea-322041f2d5a3</firmware>
       </provides>
    -  <name>XPS 15 7590/Precision 5540 System Update</name>
    -  <summary>Firmware for the Dell XPS 15 7590/Precision 5540</summary>
    -  <description>
    -    <p>
    -      Updating the system firmware improves performance.
    -    </p>
    -  </description>
       <url type="homepage">http://support.dell.com/</url>
       <project_license>proprietary</project_license>
       <developer_name>Dell Inc.</developer_name>
    +  <categories>
    +    <category>X-System</category>
    +  </categories>
    +  <custom>
    +    <value key="LVFS::VersionFormat">quad</value>
    +    <value key="LVFS::VersionFormat">dell-bios</value>
    +    <value key="LVFS::UpdateProtocol">org.uefi.capsule</value>
    +  </custom>
       <releases>
    -    <release version="0x10500" timestamp="1577256261">
    +    <release version="66816" timestamp="1577256261" urgency="critical">
    +      <checksum type="sha1" filename="firmware.bin" target="content">39af75897b09ce599514e12c381f7f5e51c7055d</checksum>
    +      <checksum type="sha256" filename="firmware.bin" target="content">a7f8bf4c1f7e7190ac8260fd0e7b73f9ba3ac5dc04a9d74cd4480ffc29f2ee9c</checksum>
    +      <description>
    +        <p>This stable release fixes the following issues:</p>
    +        <ul>
    +          <li>Fixed the issue where the system cannot be charged using the USB Type-C port.</li>
    +          <li>Fixed the issue where the Power Delivery firmware of the dock cable fails when the Dell Dock WD15 or Dell Thunderbolt Dock TB16 is updated.</li>
    +        </ul>
    +      </description>
    +      <size type="installed">21741199</size>
         </release>
       </releases>
     </component>

Which all looks right, until you look at 0x10500 turning into 66816. This is okay I *think*, but perhaps unexpected.

hughsie: > This is okay I think

I guess it's what we've been doing for metadata for ages, but I've added a condition to preserve the hex string for the metainfo rewrite.

hughsie:
> superm1 approved these changes 

What I'll do is deploy this and **not** invalidate all the existing firmware. If we don't have any problems with the newly uploaded firmware I guess we should probably resign the others too. Maybe 3 months as a "burn in test phase" might be a good idea.

PR #534, opened 2020-03-06
add hex verfmt to models

superm1:
I recall this was moved into the UX and database, so hopefully this was the only place needed.

LVFS side of https://github.com/fwupd/fwupd/pull/1842

### DIFF
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -965,6 +965,8 @@ class Verfmt(db.Model):
             return '%i.%i.%i' % ((v & 0x00ff0000) >> 16,
                                  (v & 0x0000ff00) >> 8,
                                  v & 0x000000ff)
+        if self.value == 'hex':
+            return "%#08x" % v
         return None
 
     def __repr__(self):


lvfs/models.py
```diff
@@ -965,6 +965,8 @@ def uint32_to_str(self, v):
             return '%i.%i.%i' % ((v & 0x00ff0000) >> 16,
                                  (v & 0x0000ff00) >> 8,
                                  v & 0x000000ff)
+        if self.value == 'hex':
+            return "%x" % v
```
hughsie: Doesn't this need to be `%#x` to get the `0x` prefix?

superm1:
Can you add to LVFS database too?

hughsie:
> Can you add to LVFS database too?

Done.

superm1:
You've got a typo.  Hexidemcial -> Hexadecimal

hughsie:
It's Friday! It's close enough! Fixed :)

PR #559, opened 2020-03-30
Add some documentation for the allowed requires compare values

hughsie:


docs/source/metainfo.rst
```diff
@@ -284,6 +284,28 @@ can be used.
...
++-----------+-------------+-----------------------+
+| ``le``    | ``1.2.3``   | Less than or equal    |
++-----------+-------------+-----------------------+
+| ``gt``    | ``1.2.3``   | Greater than          |
++-----------+-------------+-----------------------+
+| ``ge``    | ``1.2.3``   | Greater than or equal |
++-----------+-------------+-----------------------+
+| ``glob``  | ``??FWA*``  | Filename glob         |
++-----------+-------------+-----------------------+
+| ``regex`` | ``FW[1-7]`` | Regular expression    |
```
superm1: for regex perhaps mention which regex format it you're using (BRE, ERE, SRE, perl)

hughsie: I'm not sure myself! It's the one built into GLib itself, so "PCRE" maybe?

superm1: Maybe link to this then: https://developer.gnome.org/glib/stable/glib-regex-syntax.html

hughsie: I've clarified, thanks.

PR #563, opened 2020-04-01
Rewrite the AppStream screenshot URL to use the server CDN

hughsie:
When the screenshot image is loaded on the client system we download the file
from the vendor specified URL. This would cause the client to ping an external
vendor on firmware install, leaking the user agent and IP address.

This probably is not what the user is expecting and certainly not mentioned in
our privacy policy.

Download the image locally when building the metadata, and rewrite the URL in
the public metadata to use the server CDN instead, fixing the privacy issue.

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -15,6 +15,7 @@ import datetime
 
 from collections import defaultdict
 
+import requests
 import yara
 
 from lxml import etree as ET
@@ -59,6 +60,10 @@ def _regenerate_and_sign_metadata(only_embargo=False):
     # nothing to do
     if not remotes:
         return
+
+    # mirror screenshots
+    for r in remotes:
+        _mirror_images_for_remote(r)
 
     # set destination path from app config
     download_dir = app.config['DOWNLOAD_DIR']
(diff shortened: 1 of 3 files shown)


cron.py
```diff
@@ -246,6 +251,31 @@ def _fsck():
...
+def _mirror_images_for_remote(remote):
+
+    # mirror all the screenshots
+    for md in db.session.query(Component)\
+                        .filter(Component.screenshot_url != None)\
+                        .join(Firmware)\
+                        .join(Remote)\
+                        .filter(Remote.remote_id == remote.remote_id)\
+                        .order_by(Component.component_id.asc()):
+        basename = 'img-{}.png'.format(hashlib.sha256(md.screenshot_url.encode()).hexdigest())
```
superm1, line 263: it's not guaranteed that the image is actually png.  DOn't you need to check that if you're going to be formatting it this way?

hughsie: I guess you're right, it doesn't have to be PNG: https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html -- maybe we grab the extension from the original link and append it?

superm1: Yeah probably fine to do that.

hughsie: maybe we just do the same as the firmware? e.g. `{hash}-{originalfilename}`

cron.py
```diff
@@ -246,6 +251,31 @@ def _fsck():
...
+        fn = os.path.join(app.config['DOWNLOAD_DIR'], basename)
+        if os.path.isfile(fn):
+            print('{} exists for component {}'.format(fn, md.component_id))
+            continue
+
+        # download it now
+        print('downloading {} for component {}'.format(md.screenshot_url, md.component_id))
+        try:
+            r = requests.get(md.screenshot_url)
+        except requests.exceptions.MissingSchema as e:
```
superm1, line 273: presumably there are other things you need to worry about too like a 404 or a timeout.  Don't want to take LVFS's cron down because a server isn't up the first time you try to run this.

cron.py
superm1: sure that keeps it out of the loop, makes sense.

hughsie: Yup, that works, good catch.

cron.py
hughsie: Good catch! Fixed by using the base error `RequestException` and using `raise_for_status()` for 404.

cron.py
exploide: > maybe we grab the extension from the original link and append it?

Would it make sense to limit this to a list of known harmless image extensions? For example SVG files can contain JavaScript that gets executed when someone opens the file URL directly in the browser. This would (even though a bit theoretical) allow a vendor to prepare a malicious file that eventually gets copied to your server and does something similar to cross-site scripting.

(PHP or server-side stuff would be even worse, but I assume you don't let the webserver interpret stuff from the download folder)

hughsie: > Would it make sense to limit this to a list of known harmless image extensions

I think we should probably check it's a .png file as well, e.g. by using PIL to load the image.

> I assume you don't let the webserver interpret stuff from the download folder

Certainly not :) The other issue this patch has is that there's no way to inform the firmware uploader that the URL is invalid or 404ing for example. I'll have a rethink; perhaps implementing this as as a firmware `Test()` is a better way to implement this.

PR #564, opened 2020-04-02
Rewrite the AppStream screenshot URL to use the server CDN

hughsie:
When the screenshot image is loaded on the client system we download the file
from the vendor specified URL. This would cause the client to ping an external
vendor on firmware install, leaking the user agent and IP address.

This probably is not what the user is expecting and certainly not mentioned in
our privacy policy.

Download the image locally when running tests, and use a CDN URL in the public
metadata, fixing the privacy issue.

### DIFF
--- a/lvfs/metadata/utils.py
+++ b/lvfs/metadata/utils.py
@@ -114,7 +114,10 @@ def _generate_metadata_mds(mds, firmware_baseuri='', local=False, metainfo=False
             if md.screenshot_caption:
                 ET.SubElement(child, 'caption').text = md.screenshot_caption
             if md.screenshot_url:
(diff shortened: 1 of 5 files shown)


hughsie:
A better version of https://github.com/fwupd/lvfs-website/pull/563

hughsie:
@exploide

exploide:
I don't get how that asserts it is a png file. Wouldn't Pillow happily load jpg and other images as well without raising exceptions?

exploide:
Oh, you convert it, I see :+1:

hughsie:
Deploying this as a test was the right thing to do. I just spent 20 minutes fixing up all the old 8bitdo https://people.freedesktop.org/~hughsient/temp/ images that no longer resolved. :)

PR #565, opened 2020-04-02
Add CPU microcode definition on uploaded files and docs

superm1:


docs/source/metainfo.rst
```diff
@@ -205,6 +205,8 @@ Allowed Category Values
 +-----------------------------+-----------------------------+
 | X-PlatformSecurityProcessor | Platform Security Processor |
 +-----------------------------+-----------------------------+
+| X-Microcode                 | CPU Microcode               |
```
hughsie, line 208: Maybe an antisocial suggestion, but would `X-CpuMicrocode` be more accurate?

superm1: I think you suggested this previously, but sure can adjust :)

hughsie: Mea culpa.

superm1:
I think you have to actually add category to database for this one when you merge though, right?

hughsie:
> I think you have to actually add category to database for this one when you merge though, right?

Just done.

PR #567, opened 2020-04-03
Store a per-remote count incremented every time we rebuild it

hughsie:


lvfs/models.py
```diff
@@ -1865,6 +1865,7 @@ class Remote(db.Model):
     name = Column(Text, nullable=False)
     is_public = Column(Boolean, default=False)
     is_dirty = Column(Boolean, default=False)
+    build_cnt = Column(Integer, default=0)
```
superm1, line 1868: Should this perhaps be `BigInteger` to prevent a difficult to debug problem some day?

hughsie: Stable is currently on it's 306th rebuild, so i think we'll be fine for some time yet with Integer :)

superm1: But what about OEM embargo remotes?  Surely those are going up quicker?

hughsie: The largest value is 1386

superm1: OK, yeah overkill.  PR LGTM.

PR #569, opened 2020-04-03
Write metadata with a build suffix to prevent CDN mismatch

hughsie:
When a fwupd client wants to refresh the metadata it firsts download the tiny
signature (typically a JCat file) and then if that has changed compared to the
existing signature on disk the much larged gzipped XML metadata catalog is then
downloaded. This saves a lot of mirror bandwidth as some remotes do not change
very much, and even stable only regenerates 4 times a day.

The drawback of this is that we have two things that need to be used to verify
a signature, and one of four scenarios could happen:

 1. The same version of sig and catalog are delivered by the CDN

 2. Between getting the sig the remote is resigned and the CDN is invalidated

 3. The CDN sends an older cached version of the sig and the current catalog

 4. The CDN sends the current sig and a cached older version of the catalog

Over the last few years the first scenario happens 99.999999% of the time, but
0.000001 * millions of daily downloads means that the bug tracking the CDN issue
gets "me too"'d almost every week.

To fix the 3 failure scenarios, do three things:

 * Write each metadata catalog as a new file with a unique URL for the CDN

 * Write the name of the unique file in the JCat file so the client knows what
   to download with an alias for the old name for our future selves.

 * Delete old metadata catalogs, leaving the last 5 builds.

In scenarios 2 and 3 we just download the n-1 metadata catalog, which is fine,
and scenario 4 is now impossible as the URLs are now different.

Fixes the server side of https://github.com/fwupd/fwupd/issues/391

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -9,6 +9,7 @@
 
 import os
 import sys
+import glob
 import difflib
 import hashlib
 import datetime
@@ -71,8 +72,14 @@ def _regenerate_and_sign_metadata(only_embargo=False):
         print('Updating: %s' % r.name)
     for r, blob_xmlgz in _metadata_update_targets(remotes):
 
-        # write metadata.xml.gz
+        # write metadata-?????.xml.gz
         fn_xmlgz = os.path.join(download_dir, r.filename)
+        with open(fn_xmlgz, 'wb') as f:
+            f.write(blob_xmlgz)
+        invalid_fns.append(fn_xmlgz)
+
+        # write metadata.xml.gz
+        fn_xmlgz = os.path.join(download_dir, r.filename_newest)
         with open(fn_xmlgz, 'wb') as f:
             f.write(blob_xmlgz)
         invalid_fns.append(fn_xmlgz)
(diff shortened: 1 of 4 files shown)


cron.py
```diff
@@ -125,7 +133,20 @@ def _regenerate_and_sign_metadata(only_embargo=False):
 
     # log what we did
     for r in remotes:
-        _event_log('Signed metadata %s' % r.name)
+        _event_log('Signed metadata {} build {}'.format(r.name, r.build_cnt))
+
+    # only keep the last 5 metadata builds
```
superm1, line 138: Is there a reason to pick 5?  Future you might want that in a comment

lvfs/models.py
```diff
@@ -1909,8 +1909,24 @@ def description(self):
...
+
     @property
     def filename(self):
+        if self.name == 'private':
+            return None
+        if self.name == 'stable':
+            return 'firmware-{}-stable.xml.gz'.format(self.build_str)
+        if self.name == 'testing':
+            return 'firmware-{}-testing.xml.gz'.format(self.build_str)
+        return 'firmware-{}-{}.xml.gz'.format(self.build_str, _qa_hash(self.name[8:]))
```
superm1, line 1926: So you don't want to do the same thing with embargo channels?

cron.py
hughsie: Good question, which I don't have a good answer for. I guess 6 (24/4) might be a better number, given we regen the stable remotes every 4 hours if there are changes (worst case scenario).

hughsie: (and because the CDN could be out of date by as much as a day)

superm1: I think that's a pretty reasonable explanation.  Something else you can do is make this number dependent upon a lookup of however many times a day it regens in case the refresh rate get changed.

PR #578, opened 2020-04-10
Add support for LVFS::UpdateMessage

hughsie:
Vendors are uploading firmware with this metadata tag set, and it's being
stripped out when we re-write the metainfo.xml during signing.

### DIFF
--- a/lvfs/components/templates/component-update.html
+++ b/lvfs/components/templates/component-update.html
@@ -95,6 +95,13 @@
         The optional URL listed here should refer specifically to this firmware release.
       </p>
   </div>
+  <div class="form-group">
+    <label for="release_message">Update Message</label>
+      <input class="form-control" type="text" name="release_message" value="{{md.release_message if md.release_message}}"/>
+      <p class="text-secondary">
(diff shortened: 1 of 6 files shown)


lvfs/components/templates/component-update.html
```diff
@@ -95,6 +95,13 @@ <h2 class="card-title">Update Details</h2>
         The optional URL listed here should refer specifically to this firmware release.
       </p>
   </div>
+  <div class="form-group">
+    <label for="release_message">Update Message</label>
+      <input class="form-control" type="text" name="release_message" value="{{md.release_message if md.release_message}}"/>
+      <p class="text-secondary">
+        The update message is shown when the firmware update has completed.
```
superm1, line 102: It might be worth changing this to something letting them know it's client dependent (since I don't think a lot of clients have implemented it yet).

hughsie: s/is shown/may be shown?

superm1: ya

PR #592, opened 2020-04-24
Use PyGnuTLS rather than using certtool

hughsie:


hughsie:

PR #600, opened 2020-04-30
Added failure percentage and test to exported data

0xfede7c8:


hughsie:
Ci needs to pass :)

    NameError: name 'json' is not defined

0xfede7c8:
I'm unussually sloppy today.

0xfede7c8:
Oh, so we have to be pylint compliant. Good to know.

hughsie:
Now, if you `git rebase master -i` and change all those other commits to fixups, then force push the branch to update this PR, you'll look perfect :)

0xfede7c8:
Doesn't Github already has an squash and merge option? I think it would be the same result you are looking without the force part.

hughsie:
> Github already has an squash and merge option

hughsie:
> Github already has an squash and merge option

It does, but we've disabled it for this repo. It's too easy to accidentally *not* squash and accidentally have a ton of "fixup", "moar fixup" commits in master that live forever. If you squash locally and force push the PR will update and we can merge this.

0xfede7c8:
I already squashed it.

PR #607, opened 2020-05-09
Disable unused user accounts for GDPR compliance

hughsie:


hughsie:
@superm1 Have you got any better words I can use for an email to send to lvfs-announce? I was thinking something like:

Title: New GDPR compliance requirements
Body:

Users with an LVFS account will now have to log in at least once per year to keep their account active. If this affects you then an email will be sent to you via email on Monday. You will need to log into the LVFS within the next 6 weeks to keep the account from being automatically disabled.

We've made this small change on the LVFS primarily for GDPR compliance reasons, but also because it makes the LVFS more secure by not having accounts open for users long moved to other companies. User names and display names can also be considered "PII" and we can't store them indefinitely without consequence. Disabled users have the display name set to `Disabled User` and are not removed, nor is firmware uploaded by them.

As a reminder, as a vendor manager on the LVFS it is your responsibility to disable users when people leave the group uploading firmware, or your company.

Any questions welcome off list.

Richard.

superm1:
That sounds great to me

PR #609, opened 2020-05-11
Redirect https://fwupd.org/downloads/ back to https://fwupd.org/

hughsie:
Fixes https://github.com/fwupd/lvfs-website/issues/608


superm1:
Maybe it's better to bring them here: https://fwupd.org/lvfs/devices/ so they can find their device?

hughsie:
> Maybe it's better to bring them here: https://fwupd.org/lvfs/devices/ so

good idea, fixed.

PR #611, opened 2020-05-14
Wip/hughsie/list id requires first

hughsie:


hughsie:
> I suppose this is going to help the cases of older fwupd not knowing about hex versionformat

I suppose so, accidentally.

PR #618, opened 2020-05-19
Allow clients to upload anonymous HSI attrs

hughsie:
At the moment we just show the last 100 HSI reports, but with more data it can
be correlated back to the Firmware() and/or Vendor() objects so we can show a
per-model overview of the submitted HSI values.

This is intended to influence customer purchasing decisions.

### DIFF
--- a/lvfs/hsireports/routes.py
+++ b/lvfs/hsireports/routes.py
@@ -1,0 +1,246 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018-2020 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: GPL-2.0+
+
+import json
+
+from flask import Blueprint, request, url_for, redirect, flash, Response, render_template
+from flask_login import login_required
+
+from lvfs import db, csrf
+
+from lvfs.hash import _is_sha256
+from lvfs.models import HsiReport, HsiReportAttr
+from lvfs.util import admin_login_required, _pkcs7_signature_info
+from lvfs.util import _json_success, _json_error
+
+bp_hsireports = Blueprint('hsireports', __name__, template_folder='templates')
+
+@bp_hsireports.route('/')
+@login_required
+@admin_login_required
+def route_list():
+    rpts = db.session.query(HsiReport)\
+                     .order_by(HsiReport.timestamp.desc())\
+                     .limit(100).all()
+    return render_template('hsireport-list.html',
+                           category='admin',
+                           rpts_filter=None,
+                           rpts=rpts)
+
+@bp_hsireports.route('/vendor/<host_vendor>')
+@login_required
+@admin_login_required
+def route_vendor_show(host_vendor):
+    rpts = db.session.query(HsiReport)\
+                     .filter(HsiReport.host_vendor == host_vendor)\
+                     .order_by(HsiReport.timestamp.desc())\
+                     .limit(100).all()
+    return render_template('hsireport-list.html',
+                           category='admin',
+                           rpts_filter=host_vendor,
+                           rpts=rpts)
+
+@bp_hsireports.route('/product/<host_product>')
+@login_required
+@admin_login_required
+def route_product_show(host_product):
+    rpts = db.session.query(HsiReport)\
+                     .filter(HsiReport.host_product == host_product)\
+                     .order_by(HsiReport.timestamp.desc())\
+                     .limit(100).all()
+    return render_template('hsireport-list.html',
+                           category='admin',
+                           rpts_filter=host_product,
+                           rpts=rpts)
+
+@bp_hsireports.route('/sku/<host_sku>')
+@login_required
+@admin_login_required
+def route_sku_show(host_sku):
+    rpts = db.session.query(HsiReport)\
+                     .filter(HsiReport.host_sku == host_sku)\
+                     .order_by(HsiReport.timestamp.desc())\
+                     .limit(100).all()
+    return render_template('hsireport-list.html',
+                           category='admin',
+                           rpts_filter=host_sku,
+                           rpts=rpts)
+
+@bp_hsireports.route('/family/<host_family>')
+@login_required
+@admin_login_required
+def route_family_show(host_family):
+    rpts = db.session.query(HsiReport)\
+                     .filter(HsiReport.host_family == host_family)\
+                     .order_by(HsiReport.timestamp.desc())\
+                     .limit(100).all()
+    return render_template('hsireport-list.html',
+                           category='admin',
+                           rpts_filter=host_family,
+                           rpts=rpts)
+
+@bp_hsireports.route('/<int:hsi_report_id>')
+@login_required
+def route_show(hsi_report_id):
+    rpt = db.session.query(HsiReport).filter(HsiReport.hsi_report_id == hsi_report_id).first()
+    if not rpt:
+        flash('HsiReport does not exist', 'danger')
+        return redirect(url_for('main.route_dashboard'))
+    # security check
+    if not rpt.check_acl('@view'):
+        flash('Permission denied: Unable to view report', 'danger')
+        return redirect(url_for('main.route_dashboard'))
+    return render_template('hsireport-details.html',
+                           category='admin',
+                           rpt=rpt)
+
+@bp_hsireports.route('/<int:hsi_report_id>/raw')
+@login_required
+def route_view(hsi_report_id):
+    rpt = db.session.query(HsiReport).filter(HsiReport.hsi_report_id == hsi_report_id).first()
+    if not rpt:
+        return _json_error('HsiReport does not exist')
+    # security check
+    if not rpt.check_acl('@view'):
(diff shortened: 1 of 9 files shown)


hughsie:
Screenshot_2020-05-19 LVFS HSI Reports(1)

superm1:
>This is intended to influence customer purchasing decisions.

So part of the problem with this design I'm realizing is that something like secure boot can be turned off by an end user, and that lowers the HSI value.  That HSI shouldn't go to LVFS reports in terms of influencing customer purchases if it's something users are doing to themselves.

superm1:
Which makes me wonder if it's only worth uploading reports that are HSI 1 or higher perhaps?

hughsie:
> if it's only worth uploading reports that are HSI 1 or higher

That way shitty laptops without even basic protections will be missing, and that won't warn people not to buy them.

>

hughsie:
> secure boot can be turned off by an end user

Sure, so can the IOMMU, TPM etc. Most of it is disableable in the BIOS setup. I think most people won't make changes, and so the "correct" value would float to the top. i.e. if there are 2 HSI:0's for an XPS 13, and 143 HS!:2s, it's pretty obvious which to choose.

hughsie:
Note, that screen is currently admin-only, for public view we'd want to sanitize things and only show when we have >5 reports for a specific product model.

superm1:
> > secure boot can be turned off by an end user
> 
> Sure, so can the IOMMU, TPM etc. Most of it is disableable in the BIOS setup. I think most people won't make changes, and so the "correct" value would float to the top. i.e. if there are 2 HSI:0's for an XPS 13, and 143 HS!:2s, it's pretty obvious which to choose.

Well IOMMU and TPM I think most people would only touch if there was a problem (which is like never).  Secure boot however is really reliant upon the kernel someone is running, the distro, etc.
The results shouldn't be penalized just because someone is running NixOS or is self-signing binaries.  I almost wonder if we should just move secure boot into `!`.

lvfs/hsireports/routes.py
```diff
@@ -0,0 +1,197 @@
...
+    if not hsireports and not sec_attrs:
+        return _json_error('no hsireports included')
+    metadata = item['Metadata']
+    if not metadata:
+        return _json_error('no metadata included')
+
+    if sec_attrs:
+
+        # required metadata for this report type
+        for key in ['HostProduct', 'HostFamily', 'HostVendor']:
```
superm1, line 155: As mentioned in the other PR, I think SKU should be here too.

lvfs/hsireports/templates/hsireport-list.html
```diff
@@ -0,0 +1,38 @@
...
+  <h2 class="card-title">
+    HSI Reports
+{% if rpts_filter %}
+    ({{rpts_filter}})
+{% endif %}
+  </h2>
+<table class="table">
+  <tr class="row table-borderless">
+    <th class="col col-sm-3">Timestamp</th>
+    <th class="col col-sm-5">Model</th>
```
superm1, line 16: As I understand for Lenovo it's Model and for Dell it's Product Name.  So I think you need to show both columns here.

hughsie: I've special-cased Lenovo now in the template.

lvfs/hsireports/routes.py
hughsie: Added.

hughsie:
> which is like never

Several people on twitter replied with custom iommu kernel command lines.

> The results shouldn't be penalized just because someone is running NixOS or is self-signing binaries

superm1:
> I don't think we'd show the results publicly until we have statistically valid responses. This is all heavily WIP, and I don't know how much real end-users will "care" enough to upload security reports.

I think that realistically this needs to get brought into the GUI somewhere to prompt to get the user to upload security reports to get enough "good" data.

I don't think an OEM will even be able to contribute these reports before something launches, because a truly accurate report requires a production platform.  Some of the most important layers of security are only enabled on production.

hughsie:
> to get the user to upload security reports to get enough "good" data

I used the same report signature code for the attrs so we can match up the LVFS user if the certificate matches.  This means we could in theory only show security reports publicly that have a HSI report submitted from a user with the vendor LVFS account.

>  Some of the most important layers of security are only enabled on production.

Right, that is something to keep in mind. I really do think there's value in comparing, for instance, a "Lenovo ThinkPad X1 7th" with a "XPS 13 7390" from a security point of view, although how to get there I'm not so sure. Like everything else I guess, baby steps shuffling in the right direction.

hughsie:
Client WIP PR here: https://github.com/fwupd/fwupd/pull/2115

superm1:
> I used the same report signature code for the attrs so we can match up the LVFS user if the certificate matches. This means we could in theory only show security reports publicly that have a HSI report submitted from a user with the vendor LVFS account.

Conceptually I think this is a good idea.  But at least for Dell I think this this will be a challenging "process" since it would require production hardware that normally no one who would have these trusted client certificates actually has.  I personally have no production laptops on my desks, only late stage pre-production which isn't adequate.

> Right, that is something to keep in mind. I really do think there's value in comparing, for instance, a "Lenovo ThinkPad X1 7th" with a "XPS 13 7390" from a security point of view, although how to get there I'm not so sure. Like everything else I guess, baby steps shuffling in the right direction.

I don't think we've gotten a lot of success reports contributed either really.  Too many people use the GUI instead of the command line tools.  I would like to think that a lot of these people would be fine opting into submitting the reports.  So my opinion is to get people submitting the data it's worth working with desktop UX and privacy folks to get this plumbed into `gnome-software`.

hughsie:
> since it would require production hardware

Right.

>  I personally have no production laptops 

With the exception of my Red Hat issued laptops, I'm in exactly the same boat.

> I don't think we've gotten a lot of success reports contributed either really.

We've had 1.1M reports sent, which I'd classify as quite a lot myself: https://fwupd.org/lvfs/metrics

If we open the flood gates from 6% to ~90% of downloads generating reports then I think we also need to make the report scanning much more efficient, perhaps scailng it out to another process or even server. The LVFS already has two separate instances acting as PostreSQL master/slave so adding another "reviews" server into the mix wouldn't actually be that hard as we can share the database trivially.

lvfs/hsireports/routes.py
```diff
@@ -0,0 +1,245 @@
...
+                     .limit(100).all()
+    return render_template('hsireport-list.html',
+                           category='admin',
+                           rpts_filter=None,
+                           rpts=rpts)
+
+@bp_hsireports.route('/vendor/<host_vendor>')
+@login_required
+@admin_login_required
+def route_vendor_show(host_vendor):
```
superm1, line 36: So I do think there would be value in a vendor being able to see their own reports here rather than requiring admin for them.  It could be indicative of say a field issue.

hughsie: Yes, the plan is to "unlock" them once we have a good story for mapping the SMBIOS vendor string to a LVFS vendor_id. I also want to make some of them public long term, but that all needs a bit more thinking about.

lvfs/hsireports/routes.py
```diff
@@ -0,0 +1,245 @@
...
+        for sec_attr in sec_attrs:
+            for key in ['AppstreamId', 'HsiResult']:
+                if not key in sec_attr:
+                    return _json_error('invalid data, expected %s' % key)
+                if sec_attr[key] is None:
+                    return _json_error('missing data, expected %s' % key)
+
+        # update any old report
+        rpt = db.session.query(HsiReport)\
+                        .filter(HsiReport.machine_id == machine_id).first()
```
superm1, line 210: you don't check that machine ID was actually sent

lvfs/hsireports/routes.py
```diff
@@ -0,0 +1,245 @@
...
+            return _json_error('invalid data, expected %s' % key)
+        if item[key] is None:
+            return _json_error('missing data, expected %s' % key)
+
+    # parse only this version
+    if item['ReportVersion'] != 2:
+        return _json_error('report version not supported')
+
+    # add each firmware report
+    machine_id = item['MachineId']
```
superm1, line 182: to prevent some level of garbage uploads, should you at least check that machine_id is the right length?

lvfs/hsireports/routes.py
hughsie: I think we do:

    # check we got enough data
    for key in ['ReportVersion', 'MachineId', 'Metadata']:

lvfs/hsireports/routes.py
hughsie: Good idea, we can check if it's a valid SHA256 hash.

lvfs/hsireports/routes.py
superm1: Ah yeah I missed that higher up in the function.

lvfs/hsireports/routes.py
superm1: OK...

lvfs/hsireports/routes.py
hughsie: Added.

PR #625, opened 2020-05-26
Changes color of search input

konaraddi:
Changes color of search input from dark gray to white. This increases the contrast which makes the search input easier to read.


hughsie:
Good catch! I think this should be `rgba(255, 255, 255, 0.7)` rather than bright white.

konaraddi:
Sounds good, I changed the search input text color from `#fff` to `rgba(255, 255, 255, 0.7)`

PR #628, opened 2020-05-27
Improve the report query speed by several orders of magnitude

hughsie:
Use stacked subqueries to do all the heavy lifting on the database server. It
turns out databases are good at doing database things.

This speeds up showing matching reports from over 60 seconds in the worst case,
(often timing out the uwsgi socket) to ~150ms which is more than acceptable.

### DIFF
--- a/lvfs/analytics/templates/analytics-reports.html
+++ b/lvfs/analytics/templates/analytics-reports.html
@@ -22,11 +22,18 @@
     <code class="float-right">{{format_humanize_naturaltime(r.timestamp)}}</code>
   </h2>
   <div class="card-body">
-    <p class="card-text">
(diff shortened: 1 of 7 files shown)


superm1:
Seems that it might have broken the reports :/
https://fwupd.org/lvfs/firmware/3829/analytics/reports brings an error page now for me

hughsie:
This is because https://fwupd.org/lvfs/components/4727 has no `name`. This is super odd, I assume it was fixed up at upload time after the metainfo file was verified. Will fix.

hughsie:
https://github.com/fwupd/lvfs-website/commit/52aa068859e3482bea5db651ae5cf7ef32242893 deployed just now.

PR #637, opened 2020-06-05
Export the success confidence to the mdsync vendor

hughsie:
The actual number of reports (which will be related to the number of firmware
downloads) is a private stat we can't share with anyone; however it would be
really useful to know the order of magnitude of reports. This way 50% failure
with just two reports would not trigger any alarms.

The high/medium/low states are granular enough to be useful, but should not
leak numbers that OEMs might consider private.

### DIFF
--- a/lvfs/models.py
+++ b/lvfs/models.py
@@ -2079,6 +2079,15 @@ class Firmware(db.Model):
         return (self.report_success_cnt * 100) / total
 
     @property
+    def success_confidence(self):
+        total = self.report_failure_cnt + self.report_success_cnt + self.report_issue_cnt
+        if total > 1000:
+            return 'high'
+        if total > 100:
+            return 'medium'
+        return 'low'
+
+    @property
     def filename_absolute(self):
         if self.is_deleted:
             return os.path.join('/deleted', self.filename)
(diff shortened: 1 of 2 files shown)


hughsie:
@0xfede7c8

lvfs/models.py
```diff
@@ -2078,6 +2078,15 @@ def success(self):
             return None
         return (self.report_success_cnt * 100) / total
 
+    @property
+    def success_confidence(self):
+        total = self.report_failure_cnt + self.report_success_cnt
```
superm1, line 2083: Don't you also need `self.report_issue_cnt`?

hughsie: > Don't you also need self.report_issue_cnt

I'm not sure. I don't think the issue_cnt should count for the success value, and so logically (statistically) it doesn't seem right to include it in the confidence for that value.

> you are calculating the same "total" value.

Right. It should be cached by the ORM so I'm not so worried.

superm1: Shouldn't it decrease the confidence though then?  Remember a known issue is just a failure that is known about.

hughsie: > a known issue is just a failure that is known about

Valid point. Fixed.

PR #667, opened 2020-06-18
Wip/hughsie/remote access token

hughsie:


superm1:
LGTM and yes makes sense to connect it to a dialog to make the action easier like that.  I was going to suggest there could be a problem with muliple calls into the database simultaneously, but since it's admin only - no concerns.

PR #668, opened 2020-06-19
Add a helper script to populate a directory from a PULP_MANIFEST

hughsie:
This is much more lightweight than installing pulp-server from a random website
just for one simple client utility.

### DIFF
--- a/contrib/sync-pulp.py
+++ b/contrib/sync-pulp.py
@@ -1,0 +1,105 @@
+#!/usr/bin/python
+# Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: GPL-2.0+
+#
+# pylint: disable=too-few-public-methods,bad-continuation
+
+import os
+import hashlib
+import sys
+import posixpath
+
+import requests
+
+
+class Pulp:
+    def __init__(self, url):
+        self.url = url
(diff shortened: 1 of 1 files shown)


contrib/sync-pulp.py
```diff
@@ -0,0 +1,101 @@
...
+            print(str(e))
+            return 1
+
+        # parse into lines
+        for line in rv.content.decode().split("\n"):
+            try:
+                fn, csum, sz = line.rsplit(',', maxsplit=2)
+            except ValueError as e:
+                pass
+            else:
```
superm1, line 88: I've never seen try/except/else.  So does that mean if the try succeeds the else runs?  At least to me it's more intuitive to just put ```self._sync_file(``` in the try block directly.  Then if parses as a valueerror you just don't even run the `_sync_file` call

hughsie: > I've never seen try/except/else

I've been trying to teach myself idiomatic python.

>  So does that mean if the try succeeds the else runs

Correct.

> At least to me it's more intuitive to just put self._sync_file( in the try block directly

The idea is the "area" that's covered by the exception is smaller (so that if `self._sync_file` threw `ValueError` you know that something doesn't smell right and needs handling higher up.

In this case you're right; we could just continue in the loop and unindent the `self._sync_file` -- which is what I did. Thanks.

PR #669, opened 2020-06-20
Add some documentation on how to use the LVFS offline

hughsie:


docs/source/offline.rst
```diff
@@ -0,0 +1,107 @@
...
+in order of simplicity:
+
+Downloading manually
+====================
+
+Download the ``.cab`` files that match your hardware and then install them
+on the target hardware via `Ansible <https://www.ansible.com/>`_ or
+`Puppet <https://puppet.com/>`_ using ``fwupdmgr install foo.cab``. You can
+use ``fwupdagent get-devices`` to get the existing firmware versions of all
+hardware in a format you can parse from scripts.
```
superm1, line 24: I don't actually think this is the simplest solution.  It's actually pretty complex to have to figure out what systems you have, what firmware you have, and write the scripts.

Furthermore, I don't want this to be the first thing people see, I think it should be a last resort and not what we guide people towards.  If the other solutions are too complex for the majority I think we need a better way to glue this together.

Lastly I'm worried this removes some of the metadata specific items, like CHID support...

My suggestion to make it better:
1. Create a `contrib/` script that will download LVFS metadata and use it to download the latest files that match a particular GUID.
2. Download files into a directory
3. Set up fwupd to scan that `directory` remote
4.  Use `fwupdmgr update` in a script to update to the latest version.

docs/source/offline.rst
```diff
@@ -0,0 +1,107 @@
...
+
+The `LVFS <https://fwupd.org/lvfs/>`_ is a public webservice designed to allow
+OEMs and ODMs to upload firmware easily, and for it to be distributed securely
+to tens of millions of end users. For some people, this simply does not work
+for various reasons:
+
+* They don't trust the LVFS team, fwupd.org, GPG, certain OEMs or the CDN we use
+* They don't want thousands of computers on an internal network downloading all
+  the files over and over again
+* The internal secure network has no internet connectivity
```
superm1, line 12: I think you want to mention commands to disable LVFS up here somewhere.

docs/source/offline.rst
```diff
@@ -0,0 +1,107 @@
...
+on the target hardware via `Ansible <https://www.ansible.com/>`_ or
+`Puppet <https://puppet.com/>`_ using ``fwupdmgr install foo.cab``. You can
+use ``fwupdagent get-devices`` to get the existing firmware versions of all
+hardware in a format you can parse from scripts.
+
+Deploy in immutable image
+=========================
+
+If the OS is shipped as an image, you can just dump the ``.cab`` files into a
+directory in the deployment along with ``/etc/fwupd/remotes.d/immutable.conf``:
```
superm1, line 30: We already distribute `vendor-directory.conf` with fwupd:
https://github.com/fwupd/fwupd/blob/master/data/remotes.d/vendor-directory.conf

I think this just needs to be "enabled" in the immuatable image.

docs/source/offline.rst
```diff
@@ -0,0 +1,107 @@
...
+    [fwupd Remote]
+    Enabled=false
+    Title=Vendor (Automatic)
+    Keyring=none
+    MetadataURI=file:///usr/share/fwupd/remotes.d/vendor/firmware
+
+Then once you disable the LVFS, running ``fwupdmgr`` or ``fwupdtool`` will use
+only the cabinet archives you deploy in your immutable image. Of course, you're
+deploying a larger image because you might have several unused firmware files
+included, but this is how Google Chrome OS is using fwupd.
```
superm1, line 43: It's probably worth mentioning they only include the firmware for their particular systems.  So it's 1-3 firmware in the immutable image. today.

docs/source/offline.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Mirror the public firmware
+==========================
+
+You can use `Pulp <https://pulpproject.org/>`_ to mirror the entire *public*
+contents of the LVFS (but never private or embargoed firmware).
+
+Create a repo pointing to `PULP_MANIFEST <https://cdn.fwupd.org/downloads/PULP_MANIFEST>`_
```
superm1, line 51: People having `pulp-server` is probably less common.  I think you should split this "Mirror the public firmware" into two subsections, one that is using `pulp-server` and one that is using hte helper script.

docs/source/offline.rst
hughsie: Fixed, thanks.

docs/source/offline.rst
hughsie: Agreed, fixed.

docs/source/offline.rst
hughsie: Gotcha, fixed.

docs/source/offline.rst
hughsie: Added

docs/source/offline.rst
hughsie: > I don't actually think this is the simplest solution.

Ya, downloading the files manually is easiest, but not parsing json files. I've move this second to last just above running your own LVFS instance.

> Lastly I'm worried this removes some of the metadata specific items, like CHID support...

PR #672, opened 2020-06-23
Set the user agent for the tools that download data

hughsie:


contrib/sync-pulp.py
```diff
@@ -17,14 +17,18 @@ class Pulp:
     def __init__(self, url):
         self.url = url
         self.manifest = 'PULP_MANIFEST'
+        self.useragent = os.path.basename(sys.argv[0])
```
superm1, line 20: It might be nice to try to source a git tag to use with this too.

hughsie: I did think about that; but realistically I think the `sync-pulp.py` is going to just be copied somewhere not in the lvfs-website source tree. What I'm trying to do is some lvfs-website server side filtering so we don't "count" the firmware downloads if the user doesn't use the cdn URL.

superm1: Gotcha, OK.

PR #680, opened 2020-06-26
Add support for LVFS::UpdateImage

hughsie:


plugins/cdn-mirror/__init__.py
```diff
@@ -74,7 +71,14 @@ def run_test_on_md(self, test, md):
...
         settings = _get_settings('firmware')
-        md.screenshot_url_safe = os.path.join(settings['firmware_baseuri_cdn'], basename)
+        return os.path.join(settings['firmware_baseuri_cdn'], basename)
+
+    def run_test_on_md(self, test, md):
+
+        if md.screenshot_url and not md.screenshot_url_safe:
+            md.screenshot_url_safe = self._cdn_mirror_file(test, md.screenshot_url)
+        if md.release_image and not md.release_image_safe:
+            md.release_image_safe = self._cdn_mirror_file(test, md.release_image)
```
superm1, line 81: this probably was a problem with screenshot url too, but how do you prevent two different vendors from using `foo.com/unplug.png` and `bar.com/unplug.png` and stomping on each other?  Since the direction we're headed is "generic" images, what do you think about instead having a pre-defined list to pick from provided by LVFS?  If you need something new you work with LVFS admin to get it added.

hughsie: >  do you prevent two different vendors from using foo.com/unplug.png and bar.com/unplug.png

The hash is added to the basename, which also makes it unguessable:

        basename = 'img-{}.png'.format(hashlib.sha256(r.content).hexdigest())

 > what do you think about instead having a pre-defined list to pick from provided by LVFS

Yes, it's on the TODO. Entering a URL is kinda sucky and hosting them on the LVFS allows us to provide an "image chooser" with some high quality **stylistically coherent** line art.

superm1: > The hash is added to the basename, which also makes it unguessable:

Oh right, I missed this.

>Yes, it's on the TODO. Entering a URL is kinda sucky and hosting them on the LVFS allows us to provide an "image chooser" with some high quality stylistically coherent line art.

But so to plumb it until the art is ready, you can add some dummy files on LVFS for the cases that make sense and at least get the UX chooser in place.

hughsie: > you can add some dummy files on LVFS for the cases that make sense and at least get the UX chooser in place

 I still think we need to support custom line art too -- for instance does the LVFS want to provide the logitech specific GREEN->RED->GREEN slider image?

I also thought the image picker would be phase 2, i.e. a different PR on a different day :)

superm1: Well I guess under the covers this will be the same content and field it fills out.  OK yeah then this PR is good to go.  I think it makes sense to me now then.

hughsie: > I think it makes sense

Cool, thanks. I had a quick play with bootstrap today but I'm EOUTOFTIME for this week. Maybe next week I'll have more of a brain for HTML and CSS :)

PR #683, opened 2020-06-27
Allow vendors to specify client requirements

hughsie:


docs/source/metainfo.rst
```diff
@@ -401,6 +401,63 @@ Depth is specified as ``1`` to match the parent device and
       <firmware depth="1">12345678-1234-1234-1234-123456789012</firmware>
     </requires>
 
+Client Features
+---------------
+
+Newer versions of fwupd can restrict updates depending on the features the
```
superm1, line 407: I think it's worth mentioning the exact version here.  The tense of "newer" will not make sense at some point.

docs/source/metainfo.rst
```diff
@@ -618,6 +675,9 @@ privacy of the remote client.
 The screenshot will only be shown by the front end client when the device has
 the ``_NEEDS_BOOTLOADER`` flag.
 
+Please also add a ``<client>`` requirement if the update cannot be performed
+without showing the image.
```
superm1, line 679: or message

docs/source/metainfo.rst
hughsie: Agree, fixed.

docs/source/metainfo.rst
hughsie: Fixed.

PR #695, opened 2020-07-06
Send a weekly email about firmware left in embargo or testing

hughsie:


lvfs/users/routes.py
```diff
@@ -34,6 +34,10 @@ def setup_periodic_tasks(sender, **_):
         crontab(hour=1, minute=0),
         _async_user_disable.s(),
     )
+    sender.add_periodic_task(
+        crontab(day=1, hour=2, minute=0),
```
superm1, line 38: Doesn't this need to be changed too though to make it monthly?  It sounds like the change made so far would be once per week, only on firmware that is at least 30 days old.

hughsie: I think day=1 restricts it to the 1st day of the month.

PR #709, opened 2020-07-19
Purge firmware that is deleted after just 30 days

hughsie:
The current 6 month value means vendors often have huge amounts of deleted
firmware, and we're using huge amounts of disk space now.


superm1:
No concerns

PR #717, opened 2020-07-24
Wip/hughsie/verfmt

hughsie:


hughsie:
@superm1 this is what the new component view looks like:

Screenshot_2020-07-28 LVFS Firmware Component Details(1)

superm1:
>This is the case for the Thelio device which reports a triplet in runtime mode
from reading a special device file, but flashes over DFU which is bcd only.

Are all 3 digits actually used in the triplet?  It feels like there needs to be an accurate conversion method for that device.

lvfs/components/routes.py
```diff
@@ -275,6 +279,7 @@ def route_show(component_id, page='overview'):
     return render_template('component-' + page + '.html',
                            category='firmware',
                            protocols=protocols,
+                           verfmts=verfmts,
```
superm1, line 282: One thought here - what do you think about in the rendering of the format you show the (XX.YY.ZZ or AA.BB).  Similar to what's on https://fwupd.org/lvfs/docs/metainfo/version

hughsie:
> Are all 3 digits actually used in the triplet?

Yup, I confirmed with System76 that there were not using the standard BC/PAIR for a PID/DID. The pre-detached plugin actually reads a different file from sysfs rather than reading the descriptor.

lvfs/components/routes.py
hughsie: I did ponder that, but I think some people might get confused why "the LVFS detects the version number as 12.34"... maybe something like `Triplet-style, e.g. 1.2.3` might make it clear?

superm1:
But can the pre-detached plugin throw away the first field for example?  Like 00.XX.YY and ignore the 00, so XX.YY is the same whether you read from descriptor or from sysfs file?

lvfs/components/routes.py
superm1: Another idea is that you can actually render in there the version format as applied to different values.  So if the IHV doesn't know what they have they can see the one that is rendering it "properly"

hughsie:
> But can the pre-detached plugin throw away the first field for example?

Nah, it's designed to be a triplet. Why a pair isn't sufficient, I have no idea!

lvfs/components/routes.py
hughsie: > they can see the one that is rendering it "properly"

An excellent idea!

hughsie: This works well:

Screenshot from 2020-07-28 15-17-44

Ignore dave. Nobody likes dave.

superm1: Yeah I like that 🛩️

PR #726, opened 2020-07-28
Vendorlist: change "recently" to "last 6 months"

mendel5:
Related to https://github.com/fwupd/lvfs-website/issues/721

The wording "uploaded recently" should be changed to "uploaded in the last 6 months" to make it more comprehensible.


mendel5:
Thanks!

PR #735, opened 2020-07-30
Add card-text for latest devices supported

mendel5:
While the section "Devices supported by the LVFS" on https://fwupd.org/lvfs/devices/ has a card-text, the section "Latest Devices Supported" does not. To increase consistency I would like to add a card-text for this section as well.

### DIFF
--- a/lvfs/devices/templates/devicelist.html
+++ b/lvfs/devices/templates/devicelist.html
@@ -22,6 +22,10 @@
 <div class="card mt-3">
   <div class="card-body">
     <h2 class="card-title">Latest Devices Supported</h2>
+    <p class="card-text">
+      This list shows the 6 latest devices supported by the LVFS.
+      This list is automatically generated and will be updated when new devices are supported.
+    </p>
     <ul class="list-group">
 {% for fw in devices %}
       <li class="list-group-item">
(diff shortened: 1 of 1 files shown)


lvfs/devices/templates/devicelist.html
```diff
@@ -22,6 +22,10 @@ <h2 class="card-title">Search for devices</h2>
 <div class="card mt-3">
   <div class="card-body">
     <h2 class="card-title">Latest Devices Supported</h2>
+    <p class="card-text">
+      This list shows the 6 latest devices supported by the LVFS.
+      This list is automatically generated and will be updated when new devices are supported.
```
hughsie, line 27: Like the concept, but the `supported` word is repeated and it seems a bit wordy. Perhaps `This list shows the 6 more recent devices supported by the LVFS and is automatically updated.`

mendel5: Thinking about this again, it seems that it's kind of obvious that the list is updated automatically and not manually. So maybe we don't need that sentence at all. How about just this:

`This list shows the 6 latest devices supported by the LVFS.`

hughsie: > it seems that it's kind of obvious that the list is updated automatically

Agree!

> This list shows the 6 latest devices supported by the LVFS.

Works for me, thanks!

mendel5: Ok, I updated my pull request.

mendel5:
Thanks!

PR #736, opened 2020-07-30
Represent concatenation in heading: hardware --> vendor and device name

mendel5:
The content of this column is the vendor name (developer_name_display) and the device name (name_with_category). These two names get concatenated. To make readers aware of this concatenation, the heading should say that it is actually two pieces of information and not just a single "hardware".

### DIFF
--- a/lvfs/firmware/templates/firmware-new.html
+++ b/lvfs/firmware/templates/firmware-new.html
@@ -17,7 +17,7 @@
 <table class="table card-text">
   <tr class="row">
     <th class="col col-sm-2">When</th>
-    <th class="col col-sm-5">Hardware</th>
+    <th class="col col-sm-5">Vendor and Device name</th>
     <th class="col col-sm-3">Version</th>
     <th class="col col-sm-2">&nbsp;</th>
   </tr>


lvfs/firmware/templates/firmware-new.html
```diff
@@ -17,7 +17,7 @@ <h2 class="card-title">New Firmware</h2>
 <table class="table card-text">
   <tr class="row">
     <th class="col col-sm-2">When</th>
-    <th class="col col-sm-5">Hardware</th>
+    <th class="col col-sm-5">Vendor and Device name</th>
```
hughsie, line 20: Hmm, this looks a bit long. Perhaps something like `Vendor & Model` instead perhaps? Maybe `Hardware` is actually correct, it's what the user would refer to the device as. @superm1 opinions?

superm1: I think hardware makes sense, but if two names desires Vendor/Model perhaps?

hughsie: I think `Hardware` is probably simplest.

mendel5: Ok

PR #737, opened 2020-07-31
Wip/hughsie/pfat

hughsie:


hughsie:
@superm1: This AFYI: this only affects Lenovo firmware as Dell uses the PFS thing.

plugins/uefi-extract/__init__.py
```diff
@@ -374,28 +464,40 @@ def _run_uefi_extract_on_md(self, test, md):
...
         db.session.commit()
 
-        # try first with the plain blob (possibly with a capsule header) and
+        # is this a AMI BIOS with PFAT sections
+        if md.blob[8:16] == b'_AMIPFAT':
+            pfat = PfatFile(md.blob)
+            shards = []
+            for shard in pfat.shards:
+                shards.append(shard)
+                if shard.name == 'com.ami.BIOS_FV_BB.bin':
```
superm1, line 473: what is BB?  burnside bridge?  why isn't this one interesting?

plugins/uefi-extract/__init__.py
```diff
@@ -374,28 +464,40 @@ def _run_uefi_extract_on_md(self, test, md):
                 db.session.delete(shard)
         db.session.commit()
 
-        # try first with the plain blob (possibly with a capsule header) and
+        # is this a AMI BIOS with PFAT sections
+        if md.blob[8:16] == b'_AMIPFAT':
```
superm1, line 468: shouldn't this also be first a check of 
`            if not shards:`

or is that saying you have shards, but they're actually this other blob you have to split up?

plugins/uefi-extract/__init__.py
```diff
@@ -135,6 +135,96 @@ def get_int(self, key):
             return int(value[:-1], 16)
         return int(value)
 
+
+class PfatFile:
```
superm1, line 139: (I guess AMI is the best suited to review if this is accurate)

cron.py
```diff
@@ -105,6 +105,26 @@ def _repair_vendor():
     # all done
     db.session.commit()
 
+def _repair_tests():
+
+    # fix all the FPAT tests
```
superm1, line 110: PFAT?

plugins/uefi-extract/__init__.py
hughsie: I don't think they'll go near this with a barge pole :)

cron.py
hughsie: Ta.

plugins/uefi-extract/__init__.py
hughsie: Not sure entirely what BB is. It's the only one that makes UEFIExtract explode :)

superm1: FV I'd suspect is "firmware volume", and so you're probably discarding a lot of valuable data to look at if you're skipping it.

plugins/uefi-extract/__init__.py
hughsie: I'm disregarding if the component already has shards, because in most cases UEFIExtract picked up a few PEIMs and missed out most of the payload as the compression was FUBAR. `Component.blob` is the thing the metainfo.xml references, i.e. the `firmware.bin`.

plugins/uefi-extract/__init__.py
superm1: If this is a system with an integrated thunderbolt controller burnside bridge might make sense.  See if it parses with fwupd's thunderbolt parser at all.

hughsie: No, that doesn’t work. It seems to be a valid FVH with just one DXE in. I'll open a bug against UEFIExtract with the backtrace and when it's fixed I'll unblock it.

hughsie: https://github.com/LongSoft/UEFITool/issues/219

PR #745, opened 2020-08-07
Add a per-day download graph

hughsie:
The LF asked when was the busiest part of the day; I had no idea. Now I do.


hughsie:
Screenshot_2020-08-07 LVFS Analytics

PR #769, opened 2020-09-04
Allow unrestricted vendor accounts only for public remotes

hughsie:
This has the effect of 'breaking' firmware updates for vendors who have not
supplied the vendor_id strings ONLY for firmware in embargo. It does not break
components already being consumed by users.

Email asking about `XXX:NEVER_GOING_TO_MATCH` incoming in 5...4...3...2...1 :)

### DIFF
--- a/lvfs/metadata/utils.py
+++ b/lvfs/metadata/utils.py
@@ -54,7 +54,11 @@ def _use_hex_release_version(md):
         return False
     return True
 
-def _generate_metadata_mds(mds, firmware_baseuri='', local=False, metainfo=False):
+def _generate_metadata_mds(mds,
+                           firmware_baseuri='',
+                           local=False,
+                           metainfo=False,
+                           allow_unrestricted=True):
 
     # assume all the components have the same parent firmware information
     md = mds[0]
@@ -298,14 +302,18 @@ def _generate_metadata_mds(mds, firmware_baseuri='', local=False, metainfo=False
 
             # the vendor can upload to any hardware
             vendor = md.fw.vendor_odm
-            if vendor.is_unrestricted:
+            if vendor.is_unrestricted and allow_unrestricted:
                 continue
 
             # no restrictions in place!
             if not vendor.restrictions:
                 vendor_ids.append('XXX:NEVER_GOING_TO_MATCH')
                 continue
+
+            # add all the actual vendor_ids
             for res in vendor.restrictions:
+                if res.value == '*':
+                    continue
                 if res.value not in vendor_ids:
                     vendor_ids.append(res.value)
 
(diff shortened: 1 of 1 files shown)


superm1 (approved):

hughsie:
@superm1 what about something like this to the lvfs-announce mailing list:

---

I've already sent some email to many of you a few days ago. Some of the original groups are currently set up with no vendor ID restrictions, probably because they was created long before the vendor_id feature ever existed.

I’d like to migrate all LVFS groups to be vendor-id restricted. This limits the vendor to only being able to deploy firmware for their own devices e.g. Dell can't deploy onto Lenovo etc. With the recent growth of the LVFS, this is something that I think we should enforce. Limiting the firmware that can be uploaded by a vendor means that an accidental leak of a user password (or an attack on the LVFS itself) limits the damage an attacker could do. Just to clarify, there has been no attack or leak and this is all proactive.

Obviously I also don't want to break anything. I've asked many of you for a list of vendor_id’s for your hardware and many thanks for those that have replied already. As a reminder, you can get this using "fwupdmgr get-devices" – for instance:

├─THNSN5512GPU7:
│     Vendor:              Toshiba Corporation (NVME:0x1179)

In this case the vendor ID is the value inside the brackets. You might need to use a non-obsolete version of fwupd if the vendor ID values are not visible; using a Fedora 32 LiveCD is a really easy way to use a new version of fwupd. We can include as many vendor IDs as you need, of any NVME/ATA/DMI etc prefix.

To speed up adoption a bit, I've changed the vendor embargo remotes to require vendor_ids. If you have an unrestricted account then testing and stable will continue to be generated with no changes (so nothing will change for end users). If you're using a embargo remote for testing you might now get an error like:

Requirement failed: XXX:NEVER_GOING_TO_MATCH expected ATA:0x14A4

...or really, any string that mentions XXX:NEVER_GOING_TO_MATCH. If you see this please send me an email with the error string and I'll add the vendor_id straight away and regenerate the metadata; I can add a vendor ID on the admin panel with one click which applies instantly.

Many thanks,
---

Comments welcome.

superm1:
>Requirement failed: XXX:NEVER_GOING_TO_MATCH expected ATA:0x14A4

Hmm, I think this is going to be pretty confusing.  How about you special case this error message to something like:

"Vendor ID for device not in %s's list of approved vendors, contact $email to get it added"

hughsie:
>  How about you special case this error message to something like

It's client side, so in lots of old fwupd versions. The only text we control from the LVFS is the `XXX:NEVER_GOING_TO_MATCH` bit.

superm1:
> It's client side, so in lots of old fwupd versions. The only text we control from the LVFS is the XXX:NEVER_GOING_TO_MATCH bit.

Ah dang.  Well at least lets clean it up for newer versions!  We can spin out an updated 1.4.6 snap and flatpak.  I know a lot of Dell's partners use the snap at least.

hughsie:
> "Vendor ID for device not in %s's list of approved vendors, contact $email to get it added"

I'm not sure where we'd get the `%s` from client side (as it's all just "lvfs" at that point) and `$email` is going to be remote specific. I'm not sure it's worth a huge amount of developer time; this is really just for the ~10 remaining OEMs that have not provided data yet. All the "new" LVFS vendors get the IDs as part of the onboarding of the group creation.

PR #778, opened 2020-09-09
Wip/hughsie/race

hughsie:


superm1 (approved):
`regenerate_ts` is a poor man's mutex 👍

joakim-tjernlund:
Now I just got:
```
fwupdtool refresh
Loading…                 [***************************************]
15:06:33:0072 Fwupd                fwupd_remote_get_metadata_uri_sig: assertion 'FWUPD_IS_REMOTE (self)' failed
no metadata signature URI available
```
Do I need to do something on my side?

hughsie:
> no metadata signature URI available

Can you attach your files in /etc/fwupd/remotes.d please?

hughsie:
> fwupdtool refresh

Gahh, that was supposed to be `fwupdmgr refresh` sorry. `fwupdtool` is supposed to work too, but that's not what we're fixing here.

superm1:
> > fwupdtool refresh
> 
> Gahh, that was supposed to be `fwupdmgr refresh` sorry. `fwupdtool` is supposed to work too, but that's not what we're fixing here.

I think it's some memory corruption happening when calling `fu_util_download_out_of_process` where the first argument is actually modified somewhere down that chain.  It seems to happen consistently on the 3rd time `fu_common_spawn_sync` is called when the main loop runs (line 733).  You can see that `argv[1]` is destroyed in this case.

It only runs 3 times if 2 remotes are enabled, hence why it happens when lvfs-testing gets turned on.

joakim-tjernlund:
> > > fwupdtool refresh
> > 
> > 
> > Gahh, that was supposed to be `fwupdmgr refresh` sorry. `fwupdtool` is supposed to work too, but that's not what we're fixing here.
> 
> I think it's some memory corruption happening when calling `fu_util_download_out_of_process` where the first argument is actually modified somewhere down that chain. It seems to happen consistently on the 3rd time `fu_common_spawn_sync` is called when the main loop runs (line 733). You can see that `argv[1]` is destroyed in this case.
> 
> It only runs 3 times if 2 remotes are enabled, hence why it happens when lvfs-testing gets turned on.

I see, if you are about to fix that too I will leave my fwupd alone until it is fixed and try again

hughsie:
> some memory corruption happening

I have a fix give me 30 mins

hughsie:
> some memory corruption happening

https://github.com/fwupd/fwupd/pull/2368

joakim-tjernlund:
> > some memory corruption happening
> 
> [fwupd/fwupd#2368](https://github.com/fwupd/fwupd/pull/2368)

```
fwupdtool refresh
Loading…                 [***************************************]
```
Looks better :) Thanks

PR #782, opened 2020-09-10
Linux Foundation support files

lfdwt:


hughsie:
@lfdwt I see you closed this, but I was about to approve it; if you want to re-open it LGTM.

PR #785, opened 2020-09-10
Force the uploader after autoimporting issues

hughsie:
Autoimporting issues is hugely useful for uploaders, but several have not been
checking the 'fixed up' update descriptions for sanity, leading to bugs like
https://github.com/fwupd/lvfs-website/issues/758 that keep popping up.

There's no way we can possibly try to make the description sane in all cases,
so just force the uploader to remove the REMOVE_ME keywords and they'll
hopefully make the description a bit easier to read at the same time.

### DIFF
--- a/lvfs/components/routes.py
+++ b/lvfs/components/routes.py
@@ -595,13 +595,11 @@ def _autoimport_issues(md, prefix, kind):
         # advance string to end of CVE number
         start = idx + len(prefix) + issue_len
 
+        # add something the user has to manually remove
+        description_new += 'REMOVE_ME'
+
     # success
     if issues:
-        for empty in ['()', '( )', '(  )', '(   )', '(    )',
-                      '(,)', '(, , )', '(, , , )', '(, , , , )',
-                      '( & )']:
-            description_new = description_new.replace(empty, '')
-        description_new = description_new.replace(' \n', '\n')
         md.release_description = description_new
         md.issues.extend(issues)
 
--- a/lvfs/components/routes_test.py
+++ b/lvfs/components/routes_test.py
@@ -193,8 +193,9 @@ class LocalTestCase(LvfsTestCase):
         rv = self.app.get('/lvfs/firmware/1/problems')
         assert b'CVEs in update description' not in rv.data, rv.data.decode()
         rv = self.app.get('/lvfs/components/1/update')
-        assert b'- Address security advisories\n' +\
-               b'- Firmware updates to address security advisory' in rv.data, rv.data.decode()
+        assert b'- Address security advisories' in rv.data, rv.data.decode()
+        assert b'REMOVE_ME' in rv.data, rv.data.decode()
+        assert b'- Firmware updates to address security advisory' in rv.data, rv.data.decode()
 
     def test_device_checksums(self):
 
--- a/lvfs/util.py
+++ b/lvfs/util.py
@@ -164,6 +164,8 @@ def _check_both(problems, txt):
         _add_problem(problems, 'Intel-specific security advisory tag in update description')
     if txt.find('INTEL-TA-') != -1:
         _add_problem(problems, 'Intel-specific technical advisory tag in update description')
+    if txt.find('REMOVE_ME') != -1:
+        _add_problem(problems, 'Update description should be checked after auto-importing issues')
 
 def _check_is_fake_li(txt):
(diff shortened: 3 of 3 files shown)


superm1:
I'm still of the opinion it's better to let it stay in the format the uploader put it (leave INTEL-SA-#### or various tags) but pull out relevant things as _well_.  I realize the metadata will be bigger in doing this with some duplication, but I think this provides a better experience for everyone balancing things like when the uploader might not be the author of the changelog, or speak the language well and older clients that don't know how to see the tags will still see it as OEM intended originally.

hughsie:
> I realize the metadata will be bigger in doing this with some duplication

superm1:
>Not only that, but visually the user will see it twice, as for compatibility with old clients we actually write the issues into the update description as part of the remote generation: https://github.com/fwupd/lvfs-website/blob/master/lvfs/metadata/utils.py#L255

But so for those instead of writing the security fixes section, could just "not" fixup.  Older clients will show it like it was supposed to.  You'll have the same information then.

>This is the problem; there needs to be some kind of QA of the update description. We can't show junk written by some BIOS technician to potentially hundreds of thousands of potentially anxious users.

Right - so the description being uploaded has already been approved/sanitized/etc.  What this is asking is to have essentially a _second_ run of someone signing off on the description because the LVFS one and the "regular" one need to be worded differently.

superm1:
> By who? Some of the Dell ones I see pasted in are terrible. They certainly don't look like a human has read them, and also don't actually tell the user why they should update to a new firmware version. Look at https://user-images.githubusercontent.com/24682367/90750952-cf2d8600-e2d5-11ea-97d7-1cac897826de.png and tell me the user is supposed to know whether to update or not...

I guess I don't agree that's not user consumable.  I look at that and see giant pile of vulnerabilities are fixed, as an end user I'd say that's in "I should install this update".

Non user consumable stuff is like "Switched GOP driver default resolution to 1920x1080" or "Modified PchHsio.c to fix potential buffer overflow issue".

But even if we were on the same page here on the quality of update text, I don't feel it's appropriate for LVFS to say "no your changelog needs to be organized this standard way". When it becomes too much work to maintain multiple changelogs for an OS independent firmware binary it's asking for companies to start putting out updates like this to LVFS:
"Improve performance and stability of the system".
or
"Fix recently disclosed security vulnerabilities"

hughsie:
> I don't feel it's appropriate for LVFS to say "no your changelog needs to be organized this standard way

I think it's exactly appropriate, otherwise it's chaos and we might as well not have any update details text at all.

> Improve performance and stability of the system".
> "Fix recently disclosed security vulnerabilities

I think that is much better than just showing the user a list of random uppercase letters and digits. With the <issues> we're including in the metadata now we have clickable links if the user wants to know more about the security issues fixed.

superm1:
>I think it's exactly appropriate, otherwise it's chaos and we might as well not have any update details text at all.

Well actually if you go and look at how some other operating systems handle firmware updates that's exactly what they do:
* Windows doesn't tell you anything about firmware updates from WU, they just happen as part of OS upgrades
* ChromeOS does upgrades in background, you don't even know they upgraded many components

But the point is we're trying to give a user more control to make a conscious decision, which is why we're doing changelogs in the first place.  Every OEM has a different approach to changelog content, and the box is entirely a freeform box.  If you want to back an OEM into a corner on what is supposed to be going into the freeform box, I think it should actually be a form with multiple sections to fill out.  IE:
* "Security fixes": Filled out with CVE-XYZ, INTEL-SA-ABC, etc
* Enhancements
* Fixes
* Additional notes

Then use those sections to form it however you see fit.
I'm not saying it will actually get the content filled out properly, but it will at least organize it the way you want without magic complaining about your style when you press the submit button.

>I think that is much better than just showing the user a list of random uppercase letters and digits. With the we're including in the metadata now we have clickable links if the user wants to know more about the security issues fixed.

But the same thing can be accomplished by just making hyperlinks inline to where the text was previously.
`Firmware updates to address CVE-XYZXYZ` could easily make the `CVE-XYZCXZY` a hyperlink.  It doesn't need to be organized at the bottom.

hughsie:
> and the box is entirely a freeform box

hughsie:
tl;dr: I'll merge this, and work on something better -- as what we're ending up with now pleases nobody.

PR #786, opened 2020-09-11
Add support files for move to The Linux Foundation

lfdwt:
Add supporting files for containerization of lvfs-website to support the move to The Linux Foundation's stewardship

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,54 @@
+FROM centos:8
+ARG commit
+EXPOSE 5000
+
+RUN dnf -y upgrade
+RUN dnf install -y epel-release dnf-plugins-core
+RUN dnf config-manager --set-enabled PowerTools epel
+RUN dnf group install -y "Development Tools"
+RUN dnf -y install \
+	GeoIP-devel \
+	bsdtar \
+	cairo-gobject-devel \
+	clamav \
+	clamav-update \
+	fdupes \
+	git \
+	gnutls-utils \
+	gobject-introspection-devel \
+	libgcab1 \
+	logrotate \
+	python3-celery \
+	python3-devel \
+	python3-pip \
+	python3-psutil \
+        s3cmd
+
+# create all our dirs
+RUN bash -c 'mkdir -p /app/{scripts,conf,logs/uwsgi}'
+RUN bash -c 'mkdir -p /data/{downloads,shards,uploads,deleted,firmware}'
+RUN bash -c 'mkdir /backups'
+WORKDIR /app
+
+# create and activate a venv
+ENV VIRTUAL_ENV=/app/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# install all the things
+COPY ../../requirements.txt /app/conf
+RUN pip3 install -r conf/requirements.txt
+RUN pip3 install uwsgi
+
+# copy the app; various configs and scripts
+COPY scripts/*.sh /app/scripts/
+COPY lvfs-website/ /app/www/
+COPY files/uwsgi.ini /app/conf/uwsgi.ini
+COPY files/flaskapp.cfg /app/www/lvfs/flaskapp.cfg
+COPY files/lvfs-entrypoint.sh /app/lvfs-entrypoint.sh
+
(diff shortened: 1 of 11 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+	cairo-gobject-devel \
+	clamav \
+	clamav-update \
+	fdupes \
+	git \
+	gnutls-utils \
+	gobject-introspection-devel \
+	libgcab1 \
+	logrotate \
+	python3-celery \
```
hughsie, line 21: Out of interest, why use the system installed celery rather than the one sourced in `requirements.txt`?

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
+FROM centos:8
+ARG commit
+EXPOSE 5000
+
+RUN dnf -y upgrade
+RUN dnf install -y epel-release dnf-plugins-core
+RUN dnf config-manager --set-enabled PowerTools epel
+RUN dnf group install -y "Development Tools"
```
hughsie, line 8: That seems a big hammer; do we need more than gcc and make?

docker/files/application/flaskapp.cfg
```diff
@@ -0,0 +1,54 @@
...
+SQLALCHEMY_DATABASE_URI = secrets['db_server']
+CELERY_BROKER_URL = secrets['celery_broker']
+CELERY_RESULT_BACKEND = secrets['celery_results']
+MAIL_SERVER = secrets['smtp_server']
+MAIL_PASSWORD = secrets['smtp_password']
+ADMIN_EMAIL = secrets['admin_email']
+
+
+######## regular config
+DEBUG = True
```
hughsie, line 28: I'm a bit confused; won't this overwrite the secret config above?

docker/files/application/uwsgi.ini
```diff
@@ -0,0 +1,14 @@
+[uwsgi]
+chdir = /app/www
+module = lvfs:app
+uid = nobody
+gid = nobody
+processes = 4
+buffer-size = 65536
+enable-threads = true
+harakiri = 180
+lazy-apps = true
```
hughsie, line 10: do you want lazy-apps? I removed it from lvfs-puppet as it was causing threading problems with sqlachemy.

hughsie:
@superm1 related point; you have no problem if we move the repo from here to git**lab** do you? Having it there means we can set up the CI pipeline so that we can spin up deployment images (docker) automatically.

superm1:
No concern, but I think we should set up an automatic GitHub mirror and some messaging that it's a mirror

superm1:
Here's the other option - using Gitlab CI/CD with Github:
https://about.gitlab.com/solutions/github/

docker/Dockerfile
lfdwt: I need to revisit if this is still necessary. IIRC it was added because we had to build psycopg. If you look further down, it's removed from the system after the build.

docker/Dockerfile
lfdwt: Vestigial...I'll remove

docker/files/application/flaskapp.cfg
lfdwt: I'm not sure I follow, which bit? (FWIW this bit is only roughed-in and still needs testing against AWS)

docker/files/application/uwsgi.ini
lfdwt: Good to know, I'll remove it.

docker/Dockerfile
lfdwt: > Here's the other option - using Gitlab CI/CD with Github:
> https://about.gitlab.com/solutions/github/

If we want to keep the source on Github, then I think it's probably advantageous to use Github Actions/Packages instead of Gitlab's services.

lfdwt:
> Here's the other option - using Gitlab CI/CD with Github:
> https://about.gitlab.com/solutions/github/

If we want to keep the source on Github, then I think it's probably advantageous to use Github Actions/Packages instead of Gitlab's services.

hughsie:
> If we want to keep the source on Github

I'm not hugely bothered either way TBH, I don't want to create more work for you.

hughsie:

